PR #1, opened 2024-07-19
feat: init repo

rbiseck3:
### Description
Migration of all ingest related code from the original unstructured repo with what changes are needed to get all CI tests to pass again. This introduces only the bare minimum changes from the original code base to get it to work in it's own repo.


potter-potter (approved):
Working well from my testing.

PR #4, opened 2024-07-30
feat: add precheck to all connectors

rbiseck3:
### Description
All steps in the pipeline now support a precheck, which is a more general approach to the previous check connection functionality and applies to any step in the pipeline. Checks are run before pipeline starts.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,8 +4,6 @@ on:
   release:
     types:
       - published
-      - created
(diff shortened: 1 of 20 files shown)


potter-potter (approved):
Looks good for right now.

We'll probably have to enhance some of them to check more things later. For instance the google drive id can be bad and it still passes.

But I think this is on par with what we had in legacy.

Thank you!

PR #7, opened 2024-07-31
feat: Add filter step

rbiseck3:
### Description
Support filtering params, such as the current file glob, that can be run independently of the connector. As part of this expansion, a new filter was added to support filtering on file size, which also added this as a metadata field being persisted. 

Migration of existing PR: https://github.com/Unstructured-IO/unstructured/pull/3421

Had to update how docker compose is installed in CI since the current action doesn't support the latest version of docker compose and the older one is going to be deprecated soon.


potter-potter:
Can you add an additional e2e test that uses the 2 filtering flags. Kind of like how there are tests to demonstrate the different embedders.

So maybe add an s3 source test that filters on filename and filesize

potter-potter (approved):
works well

PR #12, opened 2024-08-06
Add Couchbase v2 Destination Connector

lokesh-couchbase:
couchbase.com is a NoSQL Key-Value Store Database. Couchbase would like to integrate with unstructured.io.

This PR Focuses on implementing v2 Destination Connector for Couchbase Database.


test_e2e/dest/couchbase.sh
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env bash
```
rbiseck3, line 1: This file needs to be updated to be executable (`chmod +x test_e2e/dest/couchbase.sh`)

unstructured_ingest/v2/processes/connectors/couchbase.py
```diff
@@ -0,0 +1,157 @@
...
+if TYPE_CHECKING:
+    from couchbase.cluster import Cluster
+
+CONNECTOR_TYPE = "couchbase"
+SERVER_API_VERSION = "1"
+
+
+@dataclass
+class CouchbaseAccessConfig(AccessConfig):
+    connection_string: str = enhanced_field(sensitive=True)
```
rbiseck3, line 33: Don't need to mark all the fields within the AccessConfig class as sensitive since the access config field in the ConnectionConfig is sensitive.

unstructured_ingest/v2/processes/connectors/couchbase.py
```diff
@@ -0,0 +1,157 @@
...
+        cluster = Cluster(connection_string, options)
+        cluster.wait_until_ready(timedelta(seconds=5))
+        return cluster
+
+    def run(self, contents: list[UploadContent], **kwargs: Any) -> None:
+        elements_dict = []
+        for content in contents:
+            with open(content.path) as elements_file:
+                elements = json.load(elements_file)
+                # Modify the elements to match the expected couchbase format
```
rbiseck3, line 125: This sort of manipulation should live in the stager. The uploader itself shouldn't be responsible for modifying the data by the time it gets it.

unstructured_ingest/v2/processes/connectors/couchbase.py
```diff
@@ -0,0 +1,157 @@
...
+                for element in elements:
+                    new_doc = {
+                        element.pop("element_id", None): {
+                            "embedding": element.pop("embeddings", None),
+                            "text": element.pop("text", None),
+                            "metadata": element.pop("metadata", None),
+                            "type": element.pop("type", None),
+                        }
```
rbiseck3, line 133: Since you're creating a new dictionary, no need to use `pop` and modify the original dict. Also we want to make sure the key exists so use the concrete `[]` here rather than `pop` or `get`:
```suggestion
                        element["element_id"]: {
                            "embedding": element.get("embeddings", None),
                            "text": element.get("text", None),
                            "metadata": element.get("metadata", None),
                            "type": element.get("type", None),
                        }
```

unstructured_ingest/v2/processes/connectors/couchbase.py
```diff
@@ -0,0 +1,157 @@
...
+        from datetime import timedelta
+
+        from couchbase.auth import PasswordAuthenticator
+        from couchbase.cluster import Cluster
+        from couchbase.options import ClusterOptions
+
+        connection_string = username = password = None
+        access_conf = self.connection_config.access_config
+        try:
+            if access_conf.connection_string is not None:
```
rbiseck3, line 104: When would this break and lead to the Exception? Also would make sense to have this as a method on the connection config class.

unstructured_ingest/v2/processes/connectors/couchbase.py
rbiseck3, line 33: Are these all required? If not, we should mark them all as optional with a default of `None` and provide some checks in a `__post_init__` method, checking that either the connection string was provided or the username/password was.

unstructured_ingest/v2/processes/connectors/couchbase.py
```diff
@@ -0,0 +1,157 @@
...
+    connection_config: CouchbaseConnectionConfig
+    cluster: Optional["Cluster"] = field(init=False, default=None)
+    connector_type: str = CONNECTOR_TYPE
+
+    def __post_init__(self):
+        self.cluster = self.connect_to_couchbase()
+
+    @requires_dependencies(["couchbase"], extras="couchbase")
+    def connect_to_couchbase(self) -> "Cluster":
+        from datetime import timedelta
```
rbiseck3, line 95: This import can live at the top of the file. Only connector specific ones need to live within a method wrapped by `requires_dependencies` decorator.

test_e2e/dest/couchbase.sh
```diff
@@ -0,0 +1,100 @@
...
+source "$SCRIPT_DIR"/cleanup.sh
+
+function cleanup() {
+
+   # Remove docker container
+  echo "Stopping Couchbase Docker container"
+  docker-compose -f "$SCRIPT_DIR"/env_setup/couchbase/common/docker-compose.yaml down --remove-orphans
+
+  # Kill couchbase background process
+  pgrep -f couchbase-dest | xargs kill
```
rbiseck3, line 35: Why is this needed if couchbase is being run within docker?

test_e2e/python/test-ingest-couchbase-output.py
```diff
@@ -0,0 +1,136 @@
...
+@click.option("--collection", type=str)
+@click.pass_context
+def cli(
+    ctx,
+    username: str,
+    password: str,
+    connection_string: str,
+    bucket: str,
+    scope: str,
+    collection: str,
```
rbiseck3, line 40: Doesn't look like these variables are being used.

test_e2e/dest/couchbase.sh
lokesh-couchbase: added

unstructured_ingest/v2/processes/connectors/couchbase.py
lokesh-couchbase: removed sensitive

unstructured_ingest/v2/processes/connectors/couchbase.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def run(
+        self,
+        elements_filepath: Path,
+        file_data: FileData,
+        output_dir: Path,
+        output_filename: str,
+        **kwargs: Any,
+    ) -> Path:
+        with open(elements_filepath) as elements_file:
```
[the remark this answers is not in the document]
lokesh-couchbase: Added the element manipulation logic here

unstructured_ingest/v2/processes/connectors/couchbase.py
lokesh-couchbase: Added the element manipulation logic in stager

unstructured_ingest/v2/processes/connectors/couchbase.py
lokesh-couchbase: Updated

unstructured_ingest/v2/processes/connectors/couchbase.py
lokesh-couchbase: Removed unused exception. Instead added a __post_init__

unstructured_ingest/v2/processes/connectors/couchbase.py
lokesh-couchbase: marked as optional with a checker

unstructured_ingest/v2/processes/connectors/couchbase.py
lokesh-couchbase: updated

test_e2e/dest/couchbase.sh
```diff
@@ -0,0 +1,100 @@
...
+fi
+
+# shellcheck disable=SC1091
+source "$SCRIPT_DIR"/cleanup.sh
+
+function cleanup() {
+
+   # Remove docker container
+  echo "Stopping Couchbase Docker container"
+  docker-compose -f "$SCRIPT_DIR"/env_setup/couchbase/common/docker-compose.yaml down --remove-orphans
```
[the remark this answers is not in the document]
lokesh-couchbase: updated

test_e2e/dest/couchbase.sh
lokesh-couchbase: I actually thought this was some unstructured process. Totally my bad her. Removed this

unstructured_ingest/v2/processes/connectors/couchbase.py
```diff
@@ -0,0 +1,159 @@
...
+    connector_type: str = CONNECTOR_TYPE
+
+    def __post_init__(self):
+        # verify if required fields are present
+        required_fields = ["connection_string", "username", "password"]
+        for cb_field in required_fields:
+            if (not getattr(self.access_config, cb_field)) or getattr(
+                self.access_config, cb_field
+            ) == "":
+                raise ValueError(f"Missing field in couchbase access config: {cb_field}")
```
rbiseck3, line 56: Move this up as a check on the access configs
```suggestion
```

unstructured_ingest/v2/processes/connectors/couchbase.py
```diff
@@ -0,0 +1,159 @@
...
+
+CONNECTOR_TYPE = "couchbase"
+SERVER_API_VERSION = "1"
+
+
+@dataclass
+class CouchbaseAccessConfig(AccessConfig):
+    connection_string: Optional[str] = None
+    username: Optional[str] = None
+    password: Optional[str] = None
```
rbiseck3, line 35: I was thinking of something like this. Correct me if I'm wrong in what's required from this. 
```suggestion
    password: Optional[str] = None
    
    def __post_init__(self):
        if not self.connection_string and not (self.username and self.password):
            raise ValueError(f"Either connection string or username and password must be set")
```

rbiseck3: Actually what does the connection string look like? Would this contain the username/password info in it? I see in your code to get the connection, all three are required.

test_e2e/dest/couchbase.sh
```diff
@@ -0,0 +1,99 @@
...
+OUTPUT_ROOT=${OUTPUT_ROOT:-$SCRIPT_DIR}
+OUTPUT_DIR=$OUTPUT_ROOT/structured-output/$OUTPUT_FOLDER_NAME
+WORK_DIR=$OUTPUT_ROOT/workdir/$OUTPUT_FOLDER_NAME
+DOWNLOAD_DIR=$OUTPUT_ROOT/download/$OUTPUT_FOLDER_NAME
+max_processes=${MAX_PROCESSES:=$(python3 -c "import os; print(os.cpu_count())")}
+DESTINATION_CB_SCOPE="_default"
+DESTINATION_CB_COLLECTION="_default"
+CI=${CI:-"false"}
+
+source "$SCRIPT_DIR"/env_setup/couchbase/common/constants.env
```
rbiseck3, line 17: Since we're using docker to simulate the connection, can we simply set these env vars in the test file directly? Usually we use this approach when the env vars contain secret values we don't want to expose.

unstructured_ingest/v2/processes/connectors/couchbase.py
rbiseck3: Looking at this again, the only value that needs to be in the `access_config` is the password. Since neither the username nor connection string have sensitive data, these can live in the connection config.

rbiseck3: Also can a user connect to couchbase without a username or password? This will impact if these are required on the config and we might not need the checks in the `__post_init__` method.

unstructured_ingest/v2/processes/connectors/couchbase.py
lokesh-couchbase: Updated

unstructured_ingest/v2/processes/connectors/couchbase.py
lokesh-couchbase: I have updated to only have password be sensitive.

> Also can a user connect to couchbase without a username or password?

No, username and password are separate from connection string and are required fields to connect to couchbase cluster

unstructured_ingest/v2/processes/connectors/couchbase.py
```diff
@@ -0,0 +1,152 @@
...
+
+@dataclass
+class CouchbaseAccessConfig(AccessConfig):
+    connection_string: Optional[str] = None
+    username: Optional[str] = None
+    password: str = enhanced_field(sensitive=True, default=None)
+
+    def __post_init__(self):
+        if not self.connection_string and not (self.username and self.password):
+            raise ValueError("Either connection string or username and password must be set")
```
rbiseck3, line 39: Now that I understand the use of these fields, since password is the only sensitive value, it's the only thing that should live in `CouchbaseAccessConfig`, with no `__post_init__` require since it's not marked `Optional`. Let's move `username` and `connection_string` to `CouchbaseConnectionConfig`.
```suggestion
@dataclass
class CouchbaseAccessConfig(AccessConfig):
    password: str
```

unstructured_ingest/v2/processes/connectors/couchbase.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+@dataclass
+class CouchbaseConnectionConfig(ConnectionConfig):
+    access_config: CouchbaseAccessConfig = enhanced_field(default_factory=CouchbaseAccessConfig)
+    bucket: Optional[str] = None
+    scope: Optional[str] = None
+    collection: Optional[str] = None
+    batch_size: int = 50
+    connector_type: str = CONNECTOR_TYPE
```
rbiseck3, line 49: Also are `bucket`, `scope` or `collection` required? If so, let's not mark then as `Optional` with a default value set. 
```suggestion
@dataclass
class CouchbaseConnectionConfig(ConnectionConfig):
    connection_string: str
    username: str
    access_config: CouchbaseAccessConfig = enhanced_field(default_factory=CouchbaseAccessConfig)
    bucket: Optional[str] = None
    scope: Optional[str] = None
    collection: Optional[str] = None
    batch_size: int = 50
    connector_type: str = CONNECTOR_TYPE
```

unstructured_ingest/v2/processes/connectors/couchbase.py
lokesh-couchbase: Makes sense. So I have made all the fields as required (Because they are required to tell exactly where to store data)

unstructured_ingest/v2/processes/connectors/couchbase.py
lokesh-couchbase: I have added few defaults so that it becomes a bit easier for a local couchbase cluster. However, in most cases, these values are needed to be overriden

unstructured_ingest/v2/processes/connectors/couchbase.py
```diff
@@ -0,0 +1,150 @@
...
+        logger.info(
+            f"writing {len(elements)} objects to destination "
+            f"bucket, {self.connection_config.bucket} "
+            f"at {self.connection_config.access_config.connection_string}",
+        )
+        bucket = self.cluster.bucket(self.connection_config.bucket)
+        scope = bucket.scope(self.connection_config.scope)
+        collection = scope.collection(self.connection_config.collection)
+
+        for chunk in batch_generator(elements, self.upload_config.batch_size):
```
rbiseck3, line 140: Also to improve performance, this can be fanned out using multiprocessing. Take a look at how that's done [here](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/processes/connectors/pinecone.py#L162-L172).

lokesh-couchbase: I don't know if it will be the best way. Because other rate-limiting on couchbase side might happen a bit earlier, but I will have to check more on that. Maybe let's proceed with approach and multi-processing can be added later-on if required?

rbiseck3 (approved):

rbiseck3:
Need to add an entry to the CHANGELOG file and bump the version to `0.0.4-dev0` in that file and the `__version__.py` file.

lokesh-couchbase:
Added necessary version updates and changelog

lokesh-couchbase:
Hey @rbiseck3 , there was some issue with how docker was creating couchbase and test case was failing because of it. I have fixed it (by not using interactive mode). Can you please re-run the CI. Thanks

PR #14, opened 2024-08-07
feat/explicit opensearch classes

rbiseck3:
### Description
To avoid confusion, define explicit config classes for the opensearch connectors rather than reusing the elasticsearch ones when creating the registry entries.

Additional update:
* Add missing fsspec destination precheck


potter-potter (approved):
looks good. tested opensearch works well.

PR #16, opened 2024-08-12
feat/Leverage pydantic base model for all configs

rbiseck3:
### Description
This PR updates all user-provided configs to be based off of pydantic base models. This gives us a lot of functionality out of the box that was being manually handled, such as:
* Handling sensitive data. Originally the enhanded data class enrichment was added for this purpose. Pydantic has a `Secret` type that can be used for the exact same purpose.
* json serialization is already handled when exporting a base model for complex types such as `Path` and `Datetime` that needed to be manually handled. 
* Better schema handling when data isn't passed in correctly.

A lot of the files changed in this PR are the dependencies because `pydantic` was added as a base dependency that trickled down into all other dependencies.


requirements/common/constraints.txt
```diff
@@ -61,3 +61,5 @@ langchain-community>=0.2.5
 # NOTE(robinson): choma was pinned to importlib-metadata>=7.1.0 but 7.1.0 was installed
 # instead of 7.2.0. Need to investigate
 importlib-metadata==7.1.0
+
+nltk==3.8.1
```
potter-potter, line 65: this seems unrelated. Is it limited by the pydantic version?

rbiseck3: This was due to an update in unstructured that I think it going ton be fixed in the next release.

unstructured_ingest/v2/utils.py
```diff
@@ -0,0 +1,45 @@
...
+    # Case SecretStr
+    return isclass(value) and issubclass(value, _SecretBase)
+
+
+def serialize_base_model(model: BaseModel) -> dict:
+    # To get the full serialized dict regardless of if values are marked as Secret
+    model_dict = model.dict()
+    for k, v in model_dict.items():
+        if isinstance(v, _SecretBase):
+            secret_value = v.get_secret_value()
```
potter-potter, line 23: Can you write a test for this. I would sleep better knowing that we don't ever mess this up.

unstructured_ingest/v2/interfaces/processor.py
```diff
@@ -1,24 +1,24 @@
...
-from unstructured_ingest.enhanced_dataclass import EnhancedDataClassJsonMixin
+from pydantic import BaseModel, ConfigDict, Field
 
 DEFAULT_WORK_DIR = str((Path.home() / ".cache" / "unstructured" / "ingest" / "pipeline").resolve())
 
 
-@dataclass
-class ProcessorConfig(EnhancedDataClassJsonMixin):
+class ProcessorConfig(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
potter-potter, line 12: what is the purpose of this? I saw it in extract_config but it doesn't seem to do anything.

unstructured_ingest/v2/pipeline/steps/chunk.py
```diff
@@ -77,8 +73,8 @@ async def _run_async(
         return ChunkStepResponse(file_data_path=file_data_path, path=str(output_filepath))
 
     def get_hash(self, extras: Optional[list[str]]) -> str:
-        hashable_string = json.dumps(
-            self.process.config.to_dict(), sort_keys=True, ensure_ascii=True
+        hashable_string = serialize_base_model_json(
```
potter-potter, line 76: This will hash with plaintext secrets. I can't remember is someone unhashes, will they see the secrets?

unstructured_ingest/v2/utils.py
rbiseck3: Added

unstructured_ingest/v2/interfaces/processor.py
rbiseck3: This allows the base config to contain content that isn't supported by pydantic, such as the Semaphore in this one.

unstructured_ingest/v2/pipeline/steps/chunk.py
rbiseck3: This should be safe.

potter-potter (approved):
Looks good.

PR #20, opened 2024-08-14
feat/autogenerate click options

rbiseck3:
### Description
To have a single source of truth for all configs and defaults and help texts, this PR introduces logic to autogenerate the click options needed for each command from the pydantic base models now being used. 

All files that were explicitly generating a cmd for each connector were now removed and these are now autogenerated, with the help text moved to the description parameter of the `Field` used to mark every variables exposed by the base models.


potter-potter:
This is wonderful. Removed an entire section of the code by leveraging pydantic's `default` and `description`.

potter-potter (approved):
lgtm

PR #23, opened 2024-08-15
feat/remove unstructured as a required dependency to run ingest

rbiseck3:
### Description
Given that partitioning/chunking can be accomplished via the hosted api and embedding is an optional step, unstructured doesn't need to be a required dependency anymore. This refactors the code base to have it be an optional dependency.

To avoid the requirement on Unstructured, the `assign_and_map_hash_ids` method was rewritten to work on dict content rather than `Element` but a unit test was added to make sure the logic never drifts. 

New CI test that only uses the API and makes sure everything works without ever installing the unstructured dependency.


test/test_chunking_utils.py
```diff
@@ -0,0 +1,31 @@
...
+project_root = test_file_path.parent.parent
+docs_path = project_root / "example-docs"
+
+
+def test_assign_and_map_hash_ids():
+    # Make sure the new logic working on dict content matches the
+    # results if using the unstructured version
+    file_path = docs_path / "book-war-and-peace-1p.txt"
+    elements = partition(filename=str(file_path.resolve()), strategy="fast")
+    chunked_elements = dispatch.chunk(elements=elements, chunking_strategy="basic")
```
potter-potter, line 21: Can we do one more test that uses the "title" chunking_strategy as well?

potter-potter (approved):
lgtm

PR #29, opened 2024-08-20
bugfix/Fix local indexer to not include directories

rbiseck3:
### Description
The local indexer was returning pointers to directories along with the files in them when the recursive flag was being passed in. This added a filter to that list to make sure to only return file content.


potter-potter (approved):
works well

PR #33, opened 2024-08-22
fix: Update calls to unstructured-client for forward compatibility

awalker4:
Calls to the unstructured-client before 0.22.0 looked like this:
```
req = shared.PartitionParameters(
    files=...,
    strategy=...,
)

res = client.general.partition(request=req)
```

Since then, the expected usage has changed with a new object:
```
req = operations.PartitionRequest(
    partition_parameters=shared.PartitionParameters(
        files=...,
        strategy=...,
    ),
)

res = client.general.partition(request=req)
```

The old call that does not use PartitionRequest is no longer supported in the upcoming 0.26.0.

# Testing
I don't have a env set up to run the ingest code so I need someone to verify this for me. All your existing ingest code should still be able to call the API.

### DIFF
--- a/unstructured_ingest/v2/processes/partitioner.py
+++ b/unstructured_ingest/v2/processes/partitioner.py
@@ -13,8 +13,8 @@ from unstructured_ingest.v2.logger import logger
 
 if TYPE_CHECKING:
     from unstructured_client import UnstructuredClient
-    from unstructured_client.models.shared import PartitionParameters
     from unstructured_client.models.operations import PartitionRequest
+    from unstructured_client.models.shared import PartitionParameters
 
 
 class PartitionerConfig(BaseModel):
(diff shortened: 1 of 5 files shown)


unstructured_ingest/v2/processes/partitioner.py
```diff
@@ -189,14 +190,16 @@ async def partition_via_api(
...
+        from unstructured_client.models.operations import PartitionRequest
 
         logger.debug(f"partitioning file {filename} with metadata: {metadata}")
         client = UnstructuredClient(
             server_url=self.config.partition_endpoint,
             api_key_auth=self.config.api_key.get_secret_value(),
         )
         partition_params = self.create_partition_parameters(filename=filename)
-        resp = await self.call_api(client=client, request=partition_params)
+        partition_request = PartitionRequest(partition_params)
```
amadeusz-ds, line 201: You could be more consistent and name this object either `partition_request` or `partition_req_obj` in both files.

amadeusz-ds (approved):
LGTM, just one suggestion

unstructured_ingest/v2/processes/partitioner.py
awalker4: I had auto merge on, but agreed! I'll keep this in mind when we're in here next.

PR #36, opened 2024-08-23
feat/support parent mutliprocessing fan out

rbiseck3:
### Description
Many uploaders at the moment still use a basic iterator or have to manually manipulate multiprocessing fan outs. This adds a new option to the uploader to run in batch or per file which allows the parent process to handle the multiprocessing fan out.


ahmetmeleq (approved):
LGTM. We do let go of the parallelization capability within a document, but that's a reasonable choice given the problems with the previous approach.

PR #37, opened 2024-08-26
[DS-21] specify fields to remain in milvus stager

jakub-sandomierz-deepsense-ai:
Added support for filtering which fields will remain in the document so user can align document structure to collection schema.
Also added option to control metadata flattening to enable better flexibility (user does not have to delete entire metadata).

See unit tests in `test/connector/test_milvus_stager.py` for examples.


test/connector/test_milvus_stager.py
```diff
@@ -0,0 +1,164 @@
...
+        ),
+    ],
+)
+def test_milvus_stager_processes_elements_correctly(
+    given_element, given_field_include_list, then_element, then_error
+):
+    config = MilvusUploadStagerConfig(fields_to_include=given_field_include_list)
+    stager = MilvusUploadStager(upload_stager_config=config)
+
+    try:
```
mateuszkuprowski, line 55: If I understand correctly you are expecting KeyError to happen here when given then_error data. 
You could consider this approach:
` if then_error:
        with pytest.raises(KeyError):
            stager.conform_dict(data=given_element)`
  it simplifies the logic plus would generate better infor for the user if somethine else goes wrong here.

mateuszkuprowski (approved):
LGTM, however please consider a small suggestion from me

test/connector/test_milvus_stager.py
jakub-sandomierz-deepsense-ai: I have added two test branches for true and false cases of then_error argument.

rbiseck3:
Why does the include/exclude metadata field in the partition step not accomplish this same thing?

jakub-sandomierz-deepsense-ai:
> Why does the include/exclude metadata field in the partition step not accomplish this same thing?

We did some analysis on this with @mateuszkuprowski . It looked like it does the same initially. The main difference is that partition step is at the beginning of processing chain and there is no guarantee that no processing block in between will keep the format uniform. The only guarantee for this comes just right before uploading. 

There are two basic scenarios for uploading the data:
- we upload to empty collection
- we upload to already populated collection

Especially second use case enforces more strict filtering of fields which goes beyond filtering metadata. For other destinations it might not make sense to drop any additional fields at all.

This is just a first quick version which enables minimal level compatibility, however to be fully compatible we should also introduce mapping step which takes input dictionary and maps its keys to the new dictionary. This step could be separate processing block in order to enable flexibility in integration of any third-party service which requires specific input schema

cragwolfe:
> For other destinations it might not make sense to drop any additional fields at all.

also, it ultimately should be possible to write to N destinations after the same partition step for the same workflow, rather than just 1 destination that we always test on now. in that case, partition really should not be making any assumptions about downstream datastores, but let datastore-specific transformation happen later or "just in time".

mateuszkuprowski:
I agree with everything aformentioned. I don't think it's partitioner's job to know what kind of destination it will write to. I strongly believe output from pipeline element should be highly standardized and ajusted when necessery only by the entities that care about data ordering etc. I believe processing steps like partitioner shouldn't have to know what's on the other side.

rbiseck3:
Given that dropping metadata isn't inherently specific to the connectors and this can be applied to any destination connector, I would advocate to move this logic into a dedicated step (call it transform maybe) similar to how the filter or uncompess logic was moved to it's own step. That way we can move the logic out of the partition step and have it applied right before the destination connector. Otherwise the same fields could be reproduces across a lot of other connectors.

PR #88, opened 2024-09-05
fix: Generalize tar decompression

ds-filipknefel:
Function for extracting tar archives used `"r:gz"` mode to open them, assuming their gzip compressed which is not true for `.tar` files which are supposed to be supported.

This PR modifies the mode to more general `"r:*"` which will open both compressed and uncompressed tar files.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 
 ### Fixes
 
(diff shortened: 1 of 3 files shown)


hubert-rutkowski85 (approved):
Maybe just add comment that r:* will open both compressed and uncompressed tar files

vangheem (approved):

rbiseck3 (approved):

PR #89, opened 2024-09-05
fix opensearch connector

vangheem:
It seems to be a mistake that this code was added here when we moved to pydantic: https://github.com/Unstructured-IO/unstructured-ingest/commit/61c5cee2759d163fa6bf02f038ff603b6c150dc8#diff-935dab3cef0f36ec1e94ddb813d4a975aad7773f545f8ba49b2cdd7a1617e290R90


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 0.0.11
+
+### Enhancements
+
+* Fix OpenSearch connector
```
rbiseck3, line 5: Nit: the usual pattern we have in the changelog is: 
```
* **Short description** Longer description
```

I think there were a couple that snuck in without that pattern.

rbiseck3 (approved):

PR #91, opened 2024-09-06
Improve pinecone payload size handling

vangheem:
Pinecone has a limit of only 2mb payload size for docs. This makes it extremely easy to fail when uploading to pinecone.

Changes to the pinecone destination connector:
1. No longer blindly send all metadata. Use a selection of known metadata like we do for sql connector.
2. Do not send `element_serialized` -- this just dramatically increases the size with unknown benefits
3. Smarter batching to handle 2mb limit.


unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -150,9 +190,21 @@ def run(self, path: Path, file_data: FileData, **kwargs: Any) -> None:
...
 
-        pinecone_batch_size = self.upload_config.batch_size
-        for pinecone_batch in batch_generator(elements_dict, pinecone_batch_size):
-            self.upsert_batch(batch=pinecone_batch)
+        max_batch_size = self.upload_config.batch_size
+        batch = []
+        batch_size = 0
+        for element in elements_dict:
+            element_size = len(json.dumps(element))
+            if (len(batch) > 0 and batch_size + element_size > (MAX_PAYLOAD_SIZE - 256)) or len(
```
rbiseck3, line 195: There's already a util to help with this that you can use: [generator_batching_wbytes](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/utils/data_prep.py#L18)

vangheem: nice! I'll use it

rbiseck3 (approved):

PR #92, opened 2024-09-09
feat/migrate embedders over

rbiseck3:
### Description
Move all embed code over to the new repo from the unstructured repo. As part of the migration, the implementation was also updated to remove any reference to the original repo and use generic dict type rather than `Element` when processing the content. 

Should be deployed in sync with this PR in the other repo to mark it deprecated there: https://github.com/Unstructured-IO/unstructured/pull/3614


unstructured_ingest/embed/voyageai.py
```diff
@@ -0,0 +1,69 @@
...
+    @requires_dependencies(
+        ["langchain", "langchain_voyageai"],
+        extras="embed-voyageai",
+    )
+    def get_client(self) -> "VoyageAIEmbeddings":
+        """Creates a Langchain VoyageAI python client to embed elements."""
+        from langchain_voyageai import VoyageAIEmbeddings
+
+        return VoyageAIEmbeddings(
+            voyage_api_key=self.api_key,
```
potter-potter, line 29: does this need get_secret_key?

test_e2e/dest/chroma.sh
```diff
@@ -22,8 +21,7 @@ source "$SCRIPT_DIR"/cleanup.sh
 
 function cleanup() {
   # Kill chroma background process
-  pgrep -f chroma-dest | xargs kill
-  cleanup_dir "$DESTINATION_PATH"
+  docker compose -f "$SCRIPT_DIR"/env_setup/chroma/docker-compose.yml down --remove-orphans -v
```
potter-potter, line 24: Thanks for switching this to docker...

potter-potter:
If you could bring this test over please.
https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured_ingest/src/local-embed-mixedbreadai.sh

unstructured_ingest/embed/voyageai.py
rbiseck3: So the voyager embeddings is also using a secret string in a base model, so passing this straight through is just fine .

potter-potter (approved):
looks good

PR #94, opened 2024-09-11
feat/refactor pinecone to upload async batches

rbiseck3:
### Description
To leverage the [parallel support](https://docs.pinecone.io/guides/data/upsert-data#sending-upserts-in-parallel) of the pinecone SDK, the destination connector was slightly refactored to use this approach from their docs.

### DIFF
--- a/unstructured_ingest/v2/processes/connectors/pinecone.py
+++ b/unstructured_ingest/v2/processes/connectors/pinecone.py
@@ -149,15 +149,23 @@ class PineconeUploader(Uploader):
             raise DestinationConnectionError(f"failed to validate connection: {e}")
 
     @requires_dependencies(["pinecone"], extras="pinecone")
-    def upsert_batch(self, batch):
+    def upsert_batches_async(self, elements_dict: list[dict]):
         from pinecone.exceptions import PineconeApiException
 
-        try:
-            index = self.connection_config.get_index()
-            response = index.upsert(batch)
-        except PineconeApiException as api_error:
-            raise DestinationConnectionError(f"http error: {api_error}") from api_error
-        logger.debug(f"results: {response}")
+        index = self.connection_config.get_index()
+        with index:
+            async_results = [
+                index.upsert(vectors=chunk, async_req=True)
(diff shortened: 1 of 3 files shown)


unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -149,15 +155,28 @@ def precheck(self):
...
+        chunks = list(
+            generator_batching_wbytes(
+                iterable=elements_dict,
+                batch_size_limit_bytes=MAX_PAYLOAD_SIZE - 100,
+                max_batch_size=self.upload_config.batch_size,
+            )
+        )
+        logger.info(f"Split doc with {len(elements_dict)} elements into {len(chunks)} batches")
+
+        pool_threads = self.upload_config.pool_threads or len(chunks)
```
vangheem, line 170: Couldn't `len(chunks)` be a big number? I would cap this at something like 10.

I'm really concerned about rate limiting here though and would like to default `pool_threads = 1`

We've had so many issues with pinecone failing and I would prefer we add error handling and retries before we optimize too much.

rbiseck3: Updated to use 1 thread by default.

vangheem (approved):

unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -149,15 +156,32 @@ def precheck(self):
...
+                batch_size_limit_bytes=MAX_PAYLOAD_SIZE - 100,
+                max_batch_size=self.upload_config.batch_size,
+            )
+        )
+        logger.info(f"Split doc with {len(elements_dict)} elements into {len(chunks)} batches")
+
+        max_pool_threads = min(len(chunks), MAX_POOL_THREADS)
+        if self.upload_config.pool_threads:
+            pool_threads = min(self.upload_config.pool_threads, max_pool_threads)
+        else:
```
vangheem, line 174: This else is probably not necessary anymore

PR #115, opened 2024-09-11
Fix onedrive bugs and some type hints

Coniferish:
Fixes some bugs in the Onedrive connector and some type hints in other files. 
This PR stems from https://github.com/Unstructured-IO/unstructured/pull/3579 where the bugs were preventing the `Office365-REST-Python-Client` dependency from being updated.

### DIFF
--- a/unstructured_ingest/v2/pipeline/interfaces.py
+++ b/unstructured_ingest/v2/pipeline/interfaces.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import asyncio
 import logging
 import multiprocessing as mp
@@ -132,7 +134,7 @@ class PipelineStep(ABC):
     async def _run_async(self, fn: Callable, **kwargs: Any) -> Optional[Any]:
         raise NotImplementedError
 
-    def run(self, _fn: Optional[Callable] = None, **kwargs: Any) -> Optional[Any]:
+    def run(self, _fn: Callable[..., Any] | None = None, **kwargs: Any) -> Optional[Any]:
         kwargs = kwargs.copy()
         otel_handler = OtelHandler(otel_endpoint=self.context.otel_endpoint, log_out=logger.debug)
         tracer = otel_handler.get_tracer()
--- a/unstructured_ingest/v2/pipeline/pipeline.py
+++ b/unstructured_ingest/v2/pipeline/pipeline.py
@@ -1,7 +1,9 @@
+from __future__ import annotations
+
 import logging
 import multiprocessing as mp
 from dataclasses import InitVar, dataclass, field
-from typing import Any, Optional, Union
+from typing import Any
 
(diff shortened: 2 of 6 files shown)


unstructured_ingest/v2/pipeline/interfaces.py
```diff
@@ -132,7 +134,7 @@ def _run(self, fn: Callable, **kwargs: Any) -> Optional[Any]:
     async def _run_async(self, fn: Callable, **kwargs: Any) -> Optional[Any]:
         raise NotImplementedError
 
-    def run(self, _fn: Optional[Callable] = None, **kwargs: Any) -> Optional[Any]:
+    def run(self, _fn: Callable[..., Any] | None = None, **kwargs: Any) -> Optional[Any]:
```
rbiseck3, line 137: We can't support the `type | None` syntax because we need to support python3.9 when that didn't exist.

unstructured_ingest/v2/pipeline/pipeline.py
```diff
@@ -48,33 +50,33 @@ class Pipeline:
...
         self,
         indexer: IndexerT,
         downloader: DownloaderT,
         partitioner: Partitioner,
-        chunker: Chunker = None,
-        embedder: Embedder = None,
-        stager: UploadStager = None,
-        uploader: Uploader = None,
-        filterer: Filterer = None,
+        chunker: Chunker | None = None,
```
rbiseck3, line 75: These all need to use `Optional[]`

unstructured_ingest/v2/pipeline/interfaces.py
Coniferish: I believe the `from __future__ import annotations` line at the top of the file resolves that. I've been using that syntax in `unstructured` where we also support 3.9. Will look it up in the docs, though.

unstructured_ingest/v2/pipeline/pipeline.py
Coniferish: Same as above comment. Will double-check

unstructured_ingest/v2/pipeline/interfaces.py
Coniferish: This [article](https://blog.derlin.ch/python-type-hints-and-future-annotations) has a decent explanation. As of py 3.7 and [PEP 563](https://peps.python.org/pep-0563/), the evaluation of annotations is deferred and treated as strings. It seems the only thing we have to be careful of is that "future.annotations lets you use anything as annotations, including features that are not yet available (or even "garbage"). This is very dangerous if you plan on reading types at runtime." But I assume our tests in 3.9 would cover that.

rbiseck3: Great! I wasn't aware of that.

rbiseck3 (approved):

PR #117, opened 2024-09-12
feat/migrate airtable source connector

rbiseck3:
### Description
Handles issue https://github.com/Unstructured-IO/unstructured-ingest/issues/97


unstructured_ingest/v2/examples/airtable.py
```diff
@@ -0,0 +1,44 @@
...
+    logger.info(f"Writing all content in: {work_dir.resolve()}")
+    Pipeline.from_configs(
+        context=ProcessorConfig(work_dir=str(work_dir.resolve()), verbose=True),
+        indexer_config=AirtableIndexerConfig(
+            list_of_paths=["app5YQxSfp220fWtm", "appJ43QmP8I17zu88"]
+        ),
+        downloader_config=AirtableDownloaderConfig(download_dir=download_path),
+        source_connection_config=AirtableConnectionConfig(
+            access_config=AirtableAccessConfig(
+                personal_access_token=os.getenv("AIRTABLE_PERSONAL_ACCESS_TOKEN")
```
hubert-rutkowski85, line 37: I understand for running this test locally, developer needs to register to Airtable, put there files from `test_e2e/expected-structured-output/airtable-diff/app5YQxSfp220fWtm` and set his PAT as this env. Is that right? 

Btw, I also see Airtable mentioned in e2e.yaml in GH action. So for that scenario, there is the code in `test_e2e/env_setup/airtable/create_scale_test_components.py` , and it's going to be run also for this v2 Airtable?

hubert-rutkowski85 (approved):
LGTM, just two questions

unstructured_ingest/v2/examples/airtable.py
rbiseck3: Now that v2 is exposed, it's what will be used by the ingest tests.

unstructured_ingest/v2/processes/connectors/airtable.py
```diff
@@ -0,0 +1,212 @@
...
+            elif len(components) == 2:
+                airtable_meta.append(
+                    AirtableTableMeta(base_id=components[0], table_id=components[1])
+                )
+            else:
+                airtable_meta.append(
+                    AirtableTableMeta(
+                        base_id=components[0], table_id=components[1], view_id=components[2]
+                    )
+                )
```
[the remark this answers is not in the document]
rbiseck3: Added better validation.

PR #119, opened 2024-09-12
feat/improve ingest tests

rbiseck3:
### Description
Because the text content coming from partition can change based on underlying OS changes and not relevant to this repo, the ingest tests were updated to detect changes in all content **other than** text and the derived element_id. This also removed the metrics evaluation as this is based on the text content itself. 

Also updates all ingest tests to use paid api for partitioning.


.github/workflows/e2e.yml
```diff
@@ -95,17 +96,9 @@ jobs:
...
-        make install-pandoc
-        sudo add-apt-repository -y ppa:alex-p/tesseract-ocr5
-        sudo apt-get update
-        sudo apt-get install -y tesseract-ocr
-        sudo apt-get install -y tesseract-ocr-kor
-        sudo apt-get install diffstat
-        tesseract --version
         sudo make install-docker-compose
         docker compose version
+        pip freeze
```
potter-potter, line 101: Is this just for development?

potter-potter (approved):

potter-potter (approved):
Very nice.

potter-potter:
Looks like you just have to address some of the CI tests. Then we are gtg.

.github/workflows/e2e.yml
rbiseck3: This is intentional to know what the final version of dependencies are being used in case things get installed in a particular order when setting up the environment.

PR #120, opened 2024-09-12
bugfix/fix model serialization

rbiseck3:
### Description
Model serialization with secret values was only being handled at the root level. Logic was updated to serialize any nested configs as well.


potter-potter (approved):
works great.

PR #121, opened 2024-09-12
bugfix/fix sharepoint permission config requirement

rbiseck3:
### Description
The logic in `process_permissions` wasn't first checking if the permission config even exists before looking for the fields in it. This add a null check first.


potter-potter (approved):
Works great

PR #125, opened 2024-09-12
test: setup test

badGarnet:
- only run s3 src test
- print out elements
- run s3 test twice in a row


potter-potter:
Is this PR still needed?

PR #128, opened 2024-09-16
chore: add type hints to chroma connector

Coniferish:
Type hints were updated in `unstructured` [here](https://github.com/Unstructured-IO/unstructured/pull/3625), so replicating those changes here.


unstructured_ingest/v2/processes/connectors/chroma.py
```diff
@@ -27,16 +29,17 @@
...
-    from chromadb import Client
+    from chromadb.api import ClientAPI
+    from chromadb.config import Settings
 
 CONNECTOR_TYPE = "chroma"
 
 
 class ChromaAccessConfig(AccessConfig):
-    settings: Optional[Annotated[dict, BeforeValidator(conform_string_to_dict)]] = Field(
+    settings: Optional[Annotated[Settings, BeforeValidator(conform_string_to_dict)]] = Field(
```
rbiseck3, line 39: These fields are now used to derive the click options, and I think this will break the mapping to click types. Might need a function to get the raw dict into the expected `Settings` type.

Coniferish: I'm going to just close this PR for now since they were errors I found while updating dependencies and I thought they would be simple changes to fix the type hints, but it looks like the click options you mention use more strict typing/pydantic and I am not as familiar with this code.

PR #129, opened 2024-09-16
bugfix/make sharepoint permission configs optional

rbiseck3:
### Description
Without setting the fields on the base config as optional, when this gets converted into CLI params, the CLI expects these fields to be required.


potter-potter (approved):
Works well

PR #133, opened 2024-09-18
feat/validate read/write permissions in fsspec prechecks

rbiseck3:
### Description
Check write access for fsspec destination connectors by writing a small blank file to the desired location. 

This was checked with a properly configured access run and without and got the expected error when access was permitted: 
```
unstructured_ingest.error.DestinationConnectionError: failed to validate connection: Access Denied
```

The indexer was also updated to make a head request on the first file it listed to check read permissions.

### DIFF
--- a/.github/actions/base-cache/action.yml
+++ b/.github/actions/base-cache/action.yml
@@ -29,7 +29,9 @@ runs:
       shell: bash
       run: |
         python${{ inputs.python-version }} -m pip install --upgrade virtualenv
-        python${{ inputs.python-version }} -m venv .venv
+        if [ ! -d ".venv" ]; then
+          python${{ inputs.python-version }} -m venv .venv
+        fi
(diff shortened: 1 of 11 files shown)


ryannikolaidis:
add a test that fails with no write access and succeeds with access?

six5532one (commented):
I see that `precheck` is defined for `FsspecUploader` and `FsspecIndexer`. Would it make sense to also define `precheck` for `FsspecDownloader`, since `get` is a distinct functionality?

rbiseck3:
@six5532one the issue with having a precheck for the downloader is we would need a sample file to read to validate the connection. But there's no guarantee that's there's a really file in the destination we can use for the downloader and the downloader doesn't know what files exist there until the indexer runs.

ryannikolaidis (approved):
lgtm!

PR #134, opened 2024-09-19
fix: Use latest astrapy interface for backend Astra ops

erichare:
This pull request updates the Astra DB Ingest v2 interface to use the, well, AstraPy v2 interface for managing connections, etc. From a user perspective, the functionality is the same, but the underlying code is more maintainable and easier to interact with.


unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -108,34 +105,38 @@ def precheck(self) -> None:
             raise DestinationConnectionError(f"failed to validate connection: {e}")
 
     @requires_dependencies(["astrapy"], extras="astradb")
-    def get_collection(self) -> "AstraDBCollection":
-        from astrapy.db import AstraDB
+    def get_collection(self) -> Any:
```
rbiseck3, line 108: Why did the typing become less specific?

erichare: So, to be honest i updated that because thats the typing hint for the return value of the function in AstraPy - i think the issue is it doesn't necessarily return a collection if that collection doesn't exist. Now with that said - if the only choices are Collection or None i agree with not using Any.... but i have to check why in the AstraPy client library its set to Any right now...

erichare: Ahh, sorry @rbiseck3 - turns out this was an issue in my local development environment. Indeed the type hint for the return is astrapy.Collection. I've pushed a commit to update this accordingly, thanks for calling it out.

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -25,7 +25,7 @@
 )
 
 if TYPE_CHECKING:
-    from astrapy.db import AstraDBCollection
+    import astrapy
```
rbiseck3, line 28: To help with typing, can we be more explicit in what's being imported? 
```suggestion
    from astrapy import Collection as AstraDBCollection
```

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -108,34 +108,38 @@ def precheck(self) -> None:
             raise DestinationConnectionError(f"failed to validate connection: {e}")
 
     @requires_dependencies(["astrapy"], extras="astradb")
-    def get_collection(self) -> "AstraDBCollection":
-        from astrapy.db import AstraDB
+    def get_collection(self) -> "astrapy.Collection":
```
rbiseck3, line 111:
```suggestion
    def get_collection(self) -> "AstraDBCollection":
```

erichare: Done!

unstructured_ingest/v2/processes/connectors/astradb.py
erichare: Updated!

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -109,33 +110,42 @@ def precheck(self) -> None:
...
         )
 
         # Create and connect to the newly created collection
-        astra_db_collection = astra_db.create_collection(
-            collection_name=collection_name,
-            dimension=embedding_dimension,
-            options=options,
-        )
+        try:
+            astra_db_collection = astra_db.create_collection(
```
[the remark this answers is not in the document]
erichare: @rbiseck3 Good point! Sorry about that! I just updated it. The only thing is that the existing test (which incidentally was out of date) was assuming that the collection got created - so i updated the test itself to create the collection, but that logic does not exist in the ingest code, just in the test script. Let me know if this looks okay!

erichare:
Hi @rbiseck3 , just wondering if there are any other changes you'd like here? Thanks so much with all the help on this.

rbiseck3:
One small comment, but otherwise everything looks good. Just need to get the tests to pass in the CI. Looks like most linting.

erichare:
Awesome, thanks so much. I just updated namespace to use the deprecated message in the Field() class.

So for the unit tests, i'm a little confused because the error that it's returning is:

```
scripts/version-sync.sh -c \
	-f "unstructured_ingest/__version__.py" semver
From https://github.com/Unstructured-IO/unstructured-ingest
 * branch            main       -> FETCH_HEAD
 * [new branch]      main       -> origin/main
Error: there is already a commit associated with version 0.0.19.
make: *** [Makefile:114: check-version] Error 1
```

But, this PR doesn't make any changes to the version. Do you have any idea why it would be complaining about this?

erichare:
@rbiseck3 I made a feeble attempt at merging in the main branch into this PR, in order to try to get rid of that version number warning... but still it doesnt like it haha. Let me know if you have any ideas when you can. Appreciate all the help.

rbiseck3:
@erichare something that we enforce in CI is that each PR bumps the version in the [CHANGELOG](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/CHANGELOG.md) as well as in [__version__.py](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/__version__.py) file to match which is why you're seeing those errors.

erichare:
Aha @rbiseck3 thank you! I was thinking that i should keep it on the unreleased version - I just pushed a changelog and version update `0.0.20` ... i think the linting is good now!

rbiseck3:
CI tests won't pass because you won't have access to the github secrets beign a fork. I'll manually merge it in assuming everything else looks appropriate.

erichare:
> CI tests won't pass because you won't have access to the github secrets being a fork. I'll manually merge it in assuming everything else looks appropriate.

Got it, makes sense. Thanks so much!

rbiseck3 (approved):

erichare:
Hi @rbiseck3 , i saw the build in main failed. I think its because of:

astrapy==1.4.2

This should be astrapy>=1.5.0 . Since `astradb.in` only specifies `astrapy`, i thought it would auto generate with the latest version? Let me know if there's any fix i can push, but that's the reason for the build failure

awalker4:
Hi @erichare - I tried updating to 1.50 in [my branch](https://github.com/Unstructured-IO/unstructured-ingest/pull/136) and it looks like there's a [different error](https://github.com/Unstructured-IO/unstructured-ingest/actions/runs/11060480883/job/30731461178?pr=136) now. Any thoughts?
```
  File "/home/runner/work/unstructured-ingest/unstructured-ingest/unstructured_ingest/pipeline/doc_factory.py", line 10, in run
    docs = self.source_doc_connector.get_ingest_docs()
  File "/home/runner/work/unstructured-ingest/unstructured-ingest/unstructured_ingest/utils/dep_check.py", line 45, in wrapper
    return func(*args, **kwargs)
  File "/home/runner/work/unstructured-ingest/unstructured-ingest/unstructured_ingest/connector/astradb.py", line 145, in get_ingest_docs
    astra_db_docs = list(self.astra_db_collection.paginated_find())
AttributeError: 'Collection' object has no attribute 'paginated_find'
Error: 'Collection' object has no attribute 'paginated_find'
```

erichare:
Austin, very sorry about that! I missed that change, the source connector
workflow was failing for me in my branch because of a missing API key. I
just opened: https://github.com/Unstructured-IO/unstructured-ingest/pull/140 to
fix it. I didn’t bump the version however, i wasn’t sure if you wanted to
just incorporate the fix manually rather than releasing a new version from
it - let me know. Thanks for all the help!

PR #135, opened 2024-09-23
Update Databricks CI test

potter-potter:
Databricks wants client_id and client_secret now instead of username password.


unstructured_ingest/v2/processes/connectors/databricks_volumes.py
```diff
@@ -140,11 +142,12 @@ def precheck(self) -> None:
...
         output_path = os.path.join(self.upload_config.path, path.name)
-        self.get_client().files.upload(
-            file_path=output_path,
-            contents=path,
-            overwrite=self.upload_config.overwrite,
-        )
+        with open(path, "rb") as elements_file:
+            self.get_client().files.upload(
+                file_path=output_path,
+                contents=elements_file,
```
rbiseck3, line 148: Do they no longer have a method that supports upload of a file directly?

potter-potter: @rbiseck3 I was not able to find one.

https://github.com/databricks/databricks-sdk-py/blob/e23b4acd0cc967ee2a3cba3c3d26666698600e35/databricks/sdk/service/files.py#L957C26-L957C34

Seems to want BinaryIO

When the code was switched from this (below) was it ever successful in CI?
https://github.com/Unstructured-IO/unstructured/commit/6e4d9ccd5b61ae10bebfbd8024c689e218fc4073#diff-f0aaac0be2c80341a0360aa912bef07c03e2e0f738a9a0de0cfb85fcd2be5275

potter-potter: When I used the old code, it either wanted to iterate over the posixpath or if I converted it into a string it just wrote the string (path) as a line of text in the file.

rbiseck3: Looks like they may have renamed it to `put()`: [files.py#L695](https://github.com/databricks/databricks-sdk-py/blob/main/databricks/sdk/service/files.py#L695)

rbiseck3 (approved):

PR #136, opened 2024-09-23
fix: Forward compatibility with unstructured-client

awalker4:
Address a few more compatibility issues in the upcoming 0.26.0 `unstructured-client`. All calls to `s.general.partition` will now require keyword args - see [this PR](https://github.com/Unstructured-IO/unstructured-python-client/pull/181) for context. 

While I'm in here, I extracted the SDK code to its own file, to be reused by the Partitioner and Chunker.

# Testing
You can install the preview of the new SDK, and then run a pipeline with partitioning/chunking by API to hit both paths. If you try this on main, you'll see errors related to the positional args of the SDK, e.g. `partition() takes 1 positional argument but 2 were given`

```
pip install --pre -U unstructured-client
unstructured-ingest local --input-path $SAMPLE_DIR --output-dir local-ingest-output --num-processes 2 --partition-by-api --api-key $SERVERLESS_KEY --chunking-strategy=by_title --chunk-by-api --chunk-api-key $SERVERLESS_KEY
```


rbiseck3:
Probably want to ergnernate teh requirements file associated with this to pull in the latest version of the client when we test in CI: https://github.com/Unstructured-IO/unstructured-ingest/blob/main/requirements/remote/client.in

rbiseck3:
Also if this breaks how it uses the client with previous versions, you can also bump the constraint on that library in the `client.in` file.

rbiseck3:
Also if the client now supports async, we could also update the ingest library to use that rather than wrapping it in an event loop. I'll leave that up to you if it's something you want to tackle in this PR.

awalker4:
> Also if the client now supports async, we could also update the ingest library to use that rather than wrapping it in an event loop. I'll leave that up to you if it's something you want to tackle in this PR.

We've got

rbiseck3 (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.0.19-dev0
+
+### Fixes
+
+* **Fix forward compatibility issues with `unstructured-client==0.26.0`. Update syntax and create a new SDK util file for resuse in the Partitioner and Chunker
```
rbiseck3, line 5:
```suggestion
* **Fix forward compatibility issues with `unstructured-client==0.26.0`. Update syntax and create a new SDK util file for reuse in the Partitioner and Chunker
```

awalker4:
@rbiseck3 I'm seeing ingest differences that shouldn't be related to this pr. Should I update the fixtures on this branch?

rbiseck3:
@awalker4 if the changes persist after rerunning the CI, then yes. That means we're getting different results from the api.

awalker4:
Looks like these final failures may be due to Astra changes. I'll bump the dependencies per [here](https://github.com/Unstructured-IO/unstructured-ingest/pull/134#issuecomment-2377839460) so we can see hopefully get everything green.

potter-potter:
I think if you add the astrapy fix here: we will be in good shape.
https://github.com/Unstructured-IO/unstructured-ingest/pull/140

potter-potter:
> I think if you add the astrapy fix here: we will be in good shape. #140

Actually looks like dest tests passed...without it. ?

PR #137, opened 2024-09-25
bugfix/Use `validate_default` to set pydantic secret defaults

rbiseck3:
### Description
Originally the default values for secrets was being set by explictly mapping it to the secret type, but this breaks json schema creation on the pedantic base models. Based on feedback from this [issue](https://github.com/pydantic/pydantic/issues/10471), approach updated to the appropriate initialization based on pydantic dev's feedback.


potter-potter (approved):
looks good

PR #140, opened 2024-09-26
FIX: use find instead of paginated_find in source connector

erichare:
This PR fixes the Astra DB source connector to be compatible with AstraPy 1.5.0+

### DIFF
--- a/unstructured_ingest/connector/astradb.py
+++ b/unstructured_ingest/connector/astradb.py
@@ -142,8 +142,14 @@ class AstraDBSourceConnector(SourceConnectorCleanupMixin, BaseSourceConnector):
     @requires_dependencies(["astrapy"], extras="astradb")
     def get_ingest_docs(self):  # type: ignore
         # Perform the find operation
-        astra_db_docs = list(self.astra_db_collection.paginated_find())
+        astra_db_docs_cursor = self.astra_db_collection.find({})
 
+        # Iterate over the cursor
+        astra_db_docs = []
+        for result in astra_db_docs_cursor:
+            astra_db_docs.append(result)
+
+        # Create a list of AstraDBIngestDoc objects
         doc_list = []
         for record in astra_db_docs:
             doc = AstraDBIngestDoc(


erichare:
@awalker4 @potter-potter sorry about the issue here! This should fix the problem with the paginated_find() method.

erichare:
@awalker4 Just checking, would you like me to bump the version to 0.0.21 and add a changelog entry to fix the linting issue here?

potter-potter:
@erichare looks like this fix is in the old astradb.py I'll see if it fixes CI and will also try it in v2.

I know its confusing but only v2 versions run unless we haven't converted them over yet. Soon all connectors will be converted. Astradb was converted to v2 a while ago.
https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/processes/connectors/astradb.py

potter-potter:
Though this does seem to fix the astrapy test. Thanks!

erichare:
> Though this does seem to fix the astrapy test. Thanks!

Glad to hear it, and sorry about the build issue. I should have caught that one!

potter-potter:
@erichare my mistake. I think we are using astradb.py in v1 for a source connector still.

erichare:
> @erichare my mistake. I think we are using astradb.py in v1 for a source connector still.

I think you are correct. I was looking this morning at migrating it to a v2 source connector - i dont know if there's an "easy" way to do that, or a guide, but i'd be happy to if you'd like.

potter-potter:
@erichare The V2 docs are here:
https://github.com/Unstructured-IO/unstructured/blob/129abd0a14c51194824c3b7f7fad9623b7cbfa54/unstructured/ingest/v2/README2.md
But they might need an overhaul.

awalker4:
Some of the ingest test output has changed - I ran the update ingest tests action on my branch the other day. So, let me go ahead and merge that as 0.0.21-dev0 and that should get the tests sorted over here.

awalker4:
> @awalker4 Just checking, would you like me to bump the version to 0.0.21 and add a changelog entry to fix the linting issue here?

I think that makes sense!

erichare:
@potter-potter you've incorporated this fix into your branch, right? So my question from above may not apply. Let me know if i should just close this.

potter-potter:
@erichare you can close this. The same thing is in my branch.

PR #143, opened 2024-09-30
feat/astradb source connector

potter-potter:
Thanks to @erichare at DataStax / Astra DB we have an AstraDB V2 connector!


unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -45,6 +89,154 @@ class AstraDBUploadStagerConfig(UploadStagerConfig):
...
+            config=self.index_config,
+        )
+
+    @requires_dependencies(["astrapy"], extras="astradb")
+    def astra_record_to_file_data(self, astra_record: Dict[str, Any]) -> FileData:
+        return FileData(
+            identifier=astra_record["_id"],
+            connector_type=CONNECTOR_TYPE,
+            source_identifiers=SourceIdentifiers(
+                fullpath=astra_record["_id"] + ".txt",
```
potter-potter, line 168: @roman Should we add the .txt here or down in the Downloader?

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -31,6 +43,38 @@
 CONNECTOR_TYPE = "astradb"
 
 
+def get_astra_collection(connection_config, config) -> "AstraDBCollection":
```
rbiseck3, line 46: Can the connection config class have a method that returns just the client: 
```python
    def get_client(self) -> "AstraDBClient":
        from astrapy import DataAPIClient as AstraDBClient

        # Create a client object to interact with the Astra DB
        # caller_name/version for Astra DB tracking
        return AstraDBClient(
            caller_name=integration_name,
            caller_version=integration_version,
        )
```

This matches the pattern elsewhere. 

Or if we move some of the fields, it seems that the database connection is required as part of the connection, regardless of download or upload, so some of those can be moved to the connection config and have a method that return the database object:

```python
    def get_database(self) -> "AstraDBDatabase":
        from astrapy import DataAPIClient as AstraDBClient

        # Create a client object to interact with the Astra DB
        # caller_name/version for Astra DB tracking
        client =  AstraDBClient(
            caller_name=integration_name,
            caller_version=integration_version,
        )

        return client.get_database(
            api_endpoint=self.api_endpoint,
            token=self.token,
            keyspace=self.keyspace,
        )
```

unstructured_ingest/v2/processes/connectors/astradb.py
rbiseck3: Given that astradb is a database, we probably want to be using the batch support here, meaning we index by pulling in the ids and expanding that in the downloader. Take a look at the elasticsearch source implementation [here](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/processes/connectors/elasticsearch.py#L173)

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -45,6 +89,154 @@ class AstraDBUploadStagerConfig(UploadStagerConfig):
...
+        # Get the download path
+        download_path = self.get_download_path(file_data=file_data)
+        download_path.parent.mkdir(parents=True, exist_ok=True)
+
+        # Write "record" to a json file at download path
+        with open(download_path, "w") as file:
+            # List comp. to prevent newlines at the end of the file which affects partitioning
+            file.write("\n".join([str(x) for x in record.values()]))
+
+        return DownloadResponse(file_data=file_data, path=download_path)
```
rbiseck3, line 237: Will want to wrap this like [this](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/processes/connectors/google_drive.py#L303):
```python
return self.generate_download_response(file_data=file_data, download_path=download_path)

```

unstructured_ingest/v2/processes/connectors/astradb.py
shreyanid: I added a get_client method in the connection config class. I wasn't able to add a database/collection getter in the connection config as well, the info needed is in (and unique to?) the Indexer/Downloader/Uploader configs
Since the existing get_astra_collection is used for both the indexer and downloader, I left that method outside the classes.

rbiseck3: can this method at least have typing info on the input params.

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -1,50 +1,303 @@
...
+    )
+    batch_size: int = Field(default=20, description="Number of records per batch")
+
+
+@dataclass
+class AstraDBIndexer(Indexer):
+    connection_config: AstraDBConnectionConfig
+    index_config: AstraDBIndexerConfig
+
+    @requires_dependencies(["astrapy"], extras="astradb")
```
rbiseck3, line 159: Don't need to decorate this with `requires_dependencies` since that dependency isn't being directly imported in this method. That decorator should be used to wrap methods that explicitly have an `import ...` statement being called inside them.

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -1,50 +1,303 @@
...
+        """Fetches all document ids in an index"""
+        # Initialize set of ids
+        ids = set()
+
+        # Get the collection
+        collection = self.get_collection()
+
+        # Perform the find operation to get all items
+        # TODO find just ids?
+        astra_db_docs_cursor = collection.find({}, projection={"_id": True, "metadata": True})
```
rbiseck3, line 183: What metadata is coming back with this? This is content we could use to enrich the `FileData` associated with this record.

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -1,50 +1,303 @@
...
+    def is_async(self) -> bool:
+        return False
+
+    def map_astra_results(self, record: dict) -> str:
+        # List comp. to prevent newlines at the end of the file which affects partitioning
+        return "\n".join([str(x) for x in record.values()])
+
+    def process_doc_id(self, collection, record_id: str) -> DownloadResponse:
+        # Fetch the record from the collection
+        record = collection.find_one({"_id": record_id})
```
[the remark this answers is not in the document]
rbiseck3: I think the `collection.find()` method is what we'd want here. Take a look at some of the other batch style connectors on how they pass a list of IDs from the indexer to the downloader.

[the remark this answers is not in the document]
rbiseck3: There's also an `AsyncCollection` in the SDK we should use to support `async` here.

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -1,50 +1,303 @@
...
+            for i in range(
+                (len(ids) + self.index_config.batch_size - 1) // self.index_config.batch_size
+            )
+        ]
+        for batch in id_batches:
+            # Make sure the hash is always a positive number to create identified
+            identified = str(hash(batch) + sys.maxsize + 1)
+            fd = FileData(
+                identifier=identified,
+                connector_type=CONNECTOR_TYPE,
```
rbiseck3, line 216:
```suggestion
                connector_type=CONNECTOR_TYPE,
                doc_type="batch",
```

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -1,50 +1,303 @@
...
+                metadata=FileDataSourceMetadata(
+                    url=f"{self.index_config.collection_name}",  # TODO figure out what goes here
```
rbiseck3, line 218: Nothing has to go here if it doesn't apply. It's an optional field.

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -1,50 +1,303 @@
...
+    namespace: Optional[str] = Field(
+        default=None,
+        description="The Astra DB connection namespace.",
+        deprecated="Please use 'keyspace' instead.",
+    )
+    batch_size: int = Field(default=20, description="Number of records per batch")
+
+
+class AstraDBDownloaderConfig(DownloaderConfig):
+    collection_name: str = Field(
```
rbiseck3, line 118: These fields should come from the metadata set on the FileData rather than being redefined by the DownloaderConfig. Since the indexer needs this info when pulling in the IDs, it can set these values in the `additional_metadata` on the `FileData` and this config no longer needs to pass it in again.

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -1,50 +1,303 @@
...
+class AstraDBDownloader(Downloader):
+    connection_config: AstraDBConnectionConfig
+    download_config: AstraDBDownloaderConfig
+    connector_type: str = CONNECTOR_TYPE
+
+    def is_async(self) -> bool:
+        return False
+
+    def map_astra_results(self, record: dict) -> str:
+        # List comp. to prevent newlines at the end of the file which affects partitioning
```
rbiseck3, line 238: To avoid losing the column information, can we write this as a csv file? That should be a valid file format for the partitioner.

unstructured_ingest/v2/processes/connectors/astradb.py
shreyanid: that makes more sense, ty

unstructured_ingest/v2/processes/connectors/astradb.py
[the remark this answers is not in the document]
shreyanid: updated to use collection.find() instead of find_one. working on async now

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -37,14 +55,288 @@ class AstraDBAccessConfig(AccessConfig):
...
+) -> "AstraDBCollection":
+    # Build the Astra DB object.
+    access_configs = connection_config.access_config.get_secret_value()
+
+    # Create a client object to interact with the Astra DB
+    # caller_name/version for Astra DB tracking
+    client = connection_config.get_client()
+
+    # Get the database object
+    astra_db = client.get_database(
```
rbiseck3, line 86: You can call this instead of having to call `astra_db.to_async()` right after. 
``suggestion
    astra_db = client.get_async_database(
```

File: unstructured_ingest/v2/processes/connectors/astradb.py [hunk 12]
```diff
@@ -37,14 +55,288 @@ class AstraDBAccessConfig(AccessConfig):
...
+        ids = list(all_ids)  # TODO check for empty/none
+        id_batches: list[frozenset[str]] = [
+            frozenset(
+                ids[
+                    i
+                    * self.index_config.batch_size : (i + 1)  # noqa
+                    * self.index_config.batch_size
+                ]
+            )
+            for i in range(
```
rbiseck3, line 219: We actually have a util method that generates batches for you given a size (`batch_generator`): [data_prep.py#L12](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/utils/data_prep.py#L12)

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -37,14 +55,288 @@ class AstraDBAccessConfig(AccessConfig):
...
+    def get_identifier(self, record_id: str) -> str:
+        f = f"{record_id}"
+        if self.download_config.fields:  # is this used??
+            f = "{}-{}".format(
+                f,
+                hashlib.sha256(",".join(self.download_config.fields).encode()).hexdigest()[:8],
+            )
+        return f
+
+    def write_astra_result_to_csv(self, astra_result: dict, download_path: str) -> str:
```
rbiseck3, line 260:
```suggestion
    def write_astra_result_to_csv(self, astra_result: dict, download_path: str) -> None:
```

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -37,14 +55,288 @@ class AstraDBAccessConfig(AccessConfig):
...
+        try:
+            self.write_astra_result_to_csv(astra_result=result, download_path=download_path)
+        except Exception as e:
+            logger.error(
+                f"failed to download from record {record_id} to {download_path}: {e}",
+                exc_info=True,
+            )
+            raise SourceConnectionNetworkError(f"failed to download file {file_data.identifier}")
+        return DownloadResponse(
+            file_data=FileData(
```
rbiseck3, line 282: Rather than redefine a brand new file data, since one is being passed in, can we modify that one and use it to retain any data the indexer might set on it?

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -37,14 +55,288 @@ class AstraDBAccessConfig(AccessConfig):
...
+                    record_locator={
+                        # add keyspace / collection name? available from file_data
+                        "document_id": record_id,
+                    },
+                ),
+            ),
+            path=download_path,
+        )
+
+    def run(self, file_data: FileData, **kwargs: Any) -> download_responses:
```
rbiseck3, line 297: We can have this method simply `raise NotImplementedError()` since only one implementation should exist in a connector.  having both might cause confusion.

unstructured_ingest/v2/processes/connectors/astradb.py
shreyanid: done. some of the info from the indexer I feel should not be maintained though, like doc_type = batch or additional metadata ids (from the batcher). remove those?

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -37,14 +55,288 @@ class AstraDBAccessConfig(AccessConfig):
...
+        download_path.parent.mkdir(parents=True, exist_ok=True)
+        try:
+            self.write_astra_result_to_csv(astra_result=result, download_path=download_path)
+        except Exception as e:
+            logger.error(
+                f"failed to download from record {record_id} to {download_path}: {e}",
+                exc_info=True,
+            )
+            raise SourceConnectionNetworkError(f"failed to download file {file_data.identifier}")
+        return DownloadResponse(
```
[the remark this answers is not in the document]
shreyanid: could you clarify this? does returning the parent's download response also return the existing download response somehow?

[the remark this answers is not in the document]
shreyanid: and what is the parent class in this case? base DownloaderConfig? just trying to understand how it fits in here

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -37,14 +55,253 @@ class AstraDBAccessConfig(AccessConfig):
...
+            logger.error(
+                f"failed to download from record {record_id} to {download_path}: {e}",
+                exc_info=True,
+            )
+            raise SourceConnectionNetworkError(f"failed to download file {file_data.identifier}")
+
+        # modify input file_data for download_response
+        copied_file_data = replace(file_data)
+        copied_file_data.identifier = filename
+        # todo set doc_type? currently "batch"
```
rbiseck3, line 274: Yep, anything that needs to be updated should be updated on the copied version, such as updating the dock_type back to `file` rather than `batch`

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -37,14 +55,253 @@ class AstraDBAccessConfig(AccessConfig):
...
+        # Retrieve results from async collection
+        download_responses = []
+        async_astra_collection = await get_async_astra_collection(
+            connection_config=self.connection_config,
+            collection_name=collection_name,
+            keyspace=keyspace,
+        )
+        async for result in async_astra_collection.find({"_id": {"$in": ids}}):
+            download_responses.append(
+                self.generate_download_response(result=result, file_data=file_data)
```
rbiseck3, line 299:
```suggestion
                self.generate_download_response(result=await result, file_data=file_data)
```

shreyanid: hm, seems to be working without this await. the async collection [find](https://docs.datastax.com/en/astra-api-docs/_attachments/python-client/astrapy/collection.html#astrapy.collection.AsyncCollection.find) is an async generator/cursor, so i believe the structure of the async for loop waits for `result` to be created before getting into the loop body

unstructured_ingest/v2/processes/connectors/astradb.py
shreyanid: ok, updated doc_type to "file" instead of "csv"

unstructured_ingest/v2/processes/connectors/astradb.py
rbiseck3: My IDE was warning me that `result` was a coroutine rather than a concrete value. wasn't sure if this needed that `await`, but if everything is working maybe I misunderstood.

rbiseck3 (approved):

PR #144, opened 2024-10-01
Update Connector Docs 

potter-potter:
How to develop V2 connectors.


rbiseck3:
FYI, docs on how to create a connector should live here: https://github.com/Unstructured-IO/unstructured-ingest/blob/46940528d031e6cb125a027b27d5d04052644577/docs/v2.md. That file was a simple copy-paste from the legacy repo. The README in this location should ideally include things like the diagrams and how this works.

ahmetmeleq:
Looking great! 

Would it be reasonable to keep representative PR url at the beginning of each  "Building a Connector" section?

PR #155, opened 2024-10-07
Outlook source connector migration

ds-filipknefel:
Migrates outlook source connector to V2


unstructured_ingest/v2/processes/connectors/outlook.py
```diff
@@ -0,0 +1,204 @@
...
+            messages += list(mail_folder.messages.get().top(MAX_EMAILS_PER_FOLDER).execute_query())
+
+            if recursive:
+                mail_folders += list(mail_folder.child_folders.get().execute_query())
+
+        return messages
+
+    def _get_selected_root_folders(self) -> list["MailFolder"]:
+        client_user = self.connection_config.get_client().users[self.index_config.user_email]
+        root_mail_folders = client_user.mail_folders.get().execute_query()
```
ds-filipknefel, line 119: @rbiseck3 

There's no native async functions provided by the package, however I think it would be possible to make it async by taking the desination URLs from the client and the sending the calls to the API directly. Something like

```py
client = connection_config.get_client()
messages = client.users[indexer.index_config.user_email].messages

url = messages.resource_url

async def get(url):
    async with aiohttp.ClientSession() as session:
        async with session.get(url, headers = {'Authorization': f'Bearer {connection_config._acquire_token()['access_token']}'}) as response:
            json_data = await response.json()
            print(json_data)

await get(url)
```

The question on this example is whether going into that is worth it? Or if there's no native support we should stick with sync unless there's a demand for speeding it up.

unstructured_ingest/v2/processes/connectors/outlook.py
```diff
@@ -0,0 +1,204 @@
...
+                date_modified=message.last_modified_datetime.strftime("%s"),
+                date_created=message.created_datetime.strftime("%s"),
+                date_processed=str(time.time()),
+                record_locator={
+                    "message_id": message.id,
+                    "user_email": self.index_config.user_email,
+                },
+            ),
+            # TODO(Filip Knefel): What about additional metadata in this case?
+            # additional_metadata={}
```
ds-filipknefel, line 160: @rbiseck3 If you have any inputs on the metadata wrt to this

unstructured_ingest/v2/processes/connectors/outlook.py
```diff
@@ -0,0 +1,204 @@
...
+        from office365.graph_client import GraphClient
+
+        return GraphClient(self._acquire_token)
+
+
+class OutlookIndexerConfig(IndexerConfig):
+    outlook_folders: list[str] = Field(
+        description="Folders to download email messages from. Do not specify subfolders. "
+        "Use quotes if there are spaces in folder names."
+    )
```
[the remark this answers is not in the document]
rbiseck3: For now lets leave it as required to match the current behavior. I agree on removing the default value.

rbiseck3:
> TODO(Filip Knefel): Based on V1 I used web_link but there's also 
message.resource_url, which one better suits this field?

I saw this comment you left, do you by chance have an example of what those two look like for a given doc?

unstructured_ingest/v2/processes/connectors/outlook.py
rbiseck3: Not worth writing custom logic that forks the underlying SDK. If we really wanted to wrap it in async, we could just through the sync code into an explicit asyncio event loop like how we call the unstructured api in the partitioner. But for now, let's leave it as a multiprocessing approach until we get feedback that we need to speed this connector up.

ds-filipknefel:
> > TODO(Filip Knefel): Based on V1 I used web_link but there's also
> > message.resource_url, which one better suits this field?
> 
> I saw this comment you left, do you by chance have an example of what those two look like for a given doc?

So I've looked into it and:
- `resource_url` is a graph API url we'd send the request to, in order to access the resource e.g.
`https://graph.microsoft.com/v1.0/users/<email>onmicrosoft.com/messages/<message-id>`
- `web_link` is a link to view the resource content via outlook in the browser

unstructured_ingest/v2/processes/connectors/outlook.py
```diff
@@ -0,0 +1,215 @@
...
+        fullpath = self._generate_fullpath(message)
+
+        return FileData(
+            identifier=message.id,
+            connector_type=CONNECTOR_TYPE,
+            source_identifiers=SourceIdentifiers(filename=fullpath.name, fullpath=str(fullpath)),
+            metadata=FileDataSourceMetadata(
+                url=message.web_link,
+                # TODO(Filip Knefel): Based on V1 I used web_link but there's also
+                # message.resource_url, which one better suits this field?
```
ds-filipknefel, line 151: Copying over my answer for easier read, which is more akin to what this url in metadata is supposed to provide?
> - `resource_url` is a graph API url we'd send the request to, in order to access the resource e.g.
>   `https://graph.microsoft.com/v1.0/users/<email>onmicrosoft.com/messages/<message-id>`
> - `web_link` is a link to view the resource content via outlook in the browser

unstructured_ingest/v2/processes/connectors/outlook.py
```diff
@@ -0,0 +1,215 @@
...
+
+@dataclass
+class OutlookDownloader(Downloader):
+    connector_type: str = CONNECTOR_TYPE
+    connection_config: OutlookConnectionConfig
+    download_config: OutlookDownloaderConfig = field(default_factory=OutlookDownloaderConfig)
+
+    def run(self, file_data: FileData, **kwargs: Any) -> download_responses:
+        download_path = self.get_download_path(file_data)
+        # NOTE: Indexer should provide source identifiers required to generate the download path
```
ds-filipknefel, line 190: I was considering adding adding an `assert` alongside this note, so that if there's a change in the Indexer incompatible with this assumption it has a chance to break in an informative way during testing, WDYT?
Applies to the note in `_download_message` as well.

unstructured_ingest/v2/processes/connectors/outlook.py
```diff
@@ -0,0 +1,218 @@
...
+    client_id: str = Field(description="Microsoft App client ID")
+    tenant: str = Field(
+        default="common", description="ID or domain name associated with your Azure AD instance"
+    )
+    authority_url: str = Field(
+        default="https://login.microsoftonline.com",
+        description="Authentication token provider for Microsoft apps",
+    )
+
+    @requires_dependencies(["msal"])
```
rbiseck3, line 52: Is there an extra associated with this dependency that can be added here?

unstructured_ingest/v2/processes/connectors/outlook.py
rbiseck3: `assert` should usually be limited to unit tests. If we want to add a check, you could have it raise a `ValueError` or something. But that should also be caught in the ingest tests in the CI.

unstructured_ingest/v2/processes/connectors/outlook.py
ds-filipknefel: No there isn't. `pip install msal` is sufficient.

However based on your question I've looked into the `requires_dependencies` and it seems that I it works in slightly different way then I though. In the other usage of `@requires_dependecies` in this file, which I've taken directly from V1, we have this snippet
```py
    @requires_dependencies(["office365"], extras="graph")
    @SourceConnectionError.wrap
    def get_client(self) -> "GraphClient":
        from office365.graph_client import GraphClient
```
if office365 is missing from dependencies it will suggest
```
ImportError: Following dependencies are missing: office365. Please install them using `pip install "unstructured-ingest[graph]"`.
Error: Following dependencies are missing: office365. Please install them using `pip install "unstructured-ingest[graph]"`.
```
which doesn't make sense. Even if interpreted as I initially did, I expected it to suggest installing `office365[graph]` that also wouldn't be accurate because we need to install `Office365-REST-Python-Client` actually.

Looks like the `requires_dependencies` decorator in its current form doesn't fit this case at all.

unstructured_ingest/v2/processes/connectors/outlook.py
hubert-rutkowski85: Hmm it depends on what's the intent of url field in FileDataSourceMetadata. There's no docs or comment about it so hard to tell. But I would guess that it's for storing url so that the resource can be accessed/fetched by machine. So a `resource_url` matches that better. But if `web_link` was used already, then maybe it was just for having a identifier... uri let's say.

potter-potter: I'm getting an error here on web_link. But resource_url works.

potter-potter: Lets use `resource_url`

unstructured_ingest/v2/processes/connectors/outlook.py
```diff
@@ -0,0 +1,218 @@
...
+                ),
+                date_created=str(message.created_datetime.replace(tzinfo=timezone.utc).timestamp()),
+                date_processed=str(time.time()),
+                record_locator={
+                    "message_id": message.id,
+                    "user_email": self.index_config.user_email,
+                },
+            ),
+            additional_metadata={
+                "sent_from": str(message.sent_from),
```
potter-potter, line 165: I'm getting errors with `message.sent_from`

Are we using the same sdk as V1?

ds-filipknefel: We are but the `sent_from` specifically was not used in V1, because there wasn't additional_metadata before.
Where do you get the errors Potter?

unstructured_ingest/v2/processes/connectors/outlook.py
ds-filipknefel: I've added the checks that will throw `ValueError`.

unstructured_ingest/v2/processes/connectors/outlook.py
```diff
@@ -0,0 +1,239 @@
...
+        authority_url = f"{self.authority_url.rstrip('/')}/{self.tenant}"
+        app = ConfidentialClientApplication(
+            authority=authority_url,
+            client_id=self.client_id,
+            client_credential=self.access_config.get_secret_value().client_credential,
+        )
+        token = app.acquire_token_for_client(scopes=["https://graph.microsoft.com/.default"])
+        return token
+
+    @requires_dependencies(["office365"], extras="graph")
```
potter-potter, line 68: I think this should be:
`@requires_dependencies(["office365"], extras="outlook")`
which results in prompting the user to 
`pip install "unstructured-ingest[outlook]"`
if they need the dependencies.

If you change it to `graph` you would have to make a bunch of changes in the requirements files etc... because when I was prompted to `pip install "unstructured-ingest[graph]"` nothing happened.

unstructured_ingest/v2/processes/connectors/outlook.py
potter-potter: I am no longer getting errors here.

unstructured_ingest/v2/processes/connectors/outlook.py
ds-filipknefel: You're right, I've confused it when writing from V1 and was actually sure it's been that way there for some reason.
I've updated that here and in the second usage of this function for `msal` import.
I've also added a doc-string to the decorator.

rbiseck3 (approved):

PR #156, opened 2024-10-07
feat/support passing in metadata to write to pinecone payload

rbiseck3:
### Description
Rather than hard code the metadata mapping, use current list as default but support passing in custom list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.0.24
+
+### Enhancements
+
+* **Support dynamic metadata mapping in Pinecone uploader**
+
 ## 0.0.23
 
 ### Fixes
--- a/test/processes/connectors/test_pinecone.py
+++ b/test/processes/connectors/test_pinecone.py
@@ -1,11 +1,12 @@
 import pytest
+
 from unstructured_ingest.v2.processes.connectors.pinecone import (
     PineconeUploadStager,
     PineconeUploadStagerConfig,
 )
 
 
-@pytest.fixture()
+@pytest.fixture
 def test_element_dict():
     return {
         "embeddings": [0, 1],
@@ -13,7 +14,7 @@ def test_element_dict():
         "metadata": {
             "text_as_html": "text as html",
             "foo": "foo",
-        }
+        },
     }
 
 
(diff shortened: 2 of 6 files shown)


vangheem (approved):

cragwolfe (commented):
I'm basically ok with this but before rushing to approve, noting that:

1/ we do not chunk Tables with other elements to form CompositeElements. this is likely a mistake, and adding this is a bit of a "double down" on that mistake
2/ we are missing a ton of metadata that for sure customers will ask about in CompositeElements. imo, this might be mostly solved with the already gzip'ed base64 encoded data of `.orig_elements`
3/ given 1 and 2, what is the right thing to do with Tables right now? i would argue, treat it as a CompositeElement with 1 element where text_as_html is available in `.orig_elements`. then the access pattern for additional metadata is similar to other chunks.

badGarnet:
> I'm basically ok with this but before rushing to approve, noting that:
> 
> 1/ we do not chunk Tables with other elements to form CompositeElements. this is likely a mistake, and adding this is a bit of a "double down" on that mistake 2/ we are missing a ton of metadata that for sure customers will ask about in CompositeElements. imo, this might be mostly solved with the already gzip'ed base64 encoded data of `.orig_elements` 3/ given 1 and 2, what is the right thing to do with Tables right now? i would argue, treat it as a CompositeElement with 1 element where text_as_html is available in `.orig_elements`. then the access pattern for additional metadata is similar to other chunks.

for table we are not chunking yet because it makes less sense to have contents of a table being joint or broken into pieces. At least with the simplistic algorithm we have for now counting characters. To chunk tables we want to approach with more ML research to understand what is a proper way to chunk table for downstream application.

cragwolfe:
[sidebar from initial "stop the bleeding" PR]

> for table we are not chunking yet because it makes less sense to have contents of a table being joint or broken into pieces. At least with the simplistic algorithm we have for now counting characters. To chunk tables we want to approach with more ML research to understand what is a proper way to chunk table for downstream application.

i mean, one way would be to use the table description concatenated with before and after chunks. but to your point, more research could be undertaken to find a optimal solution for a certain dataset. in the spirit of a flexible Platform, it should also be configurable by the user.

of course, how to chunk matters small tables. very large tables may be broken up into smaller tables in the first place.

the main point is, if we have any belief that someday tables will be chunked, customers will need to get html metadata at runtime from `.orig_elements` in the same way the would get other constituent element info.

PR #159, opened 2024-10-08
feat/support pinecone namespace

rbiseck3:
### Descriptionm
To support multitenancy, add optional namespace parameter that gets passed through during upsert call in Pinecone.

### DIFF
--- a/unstructured_ingest/v2/processes/connectors/pinecone.py
+++ b/unstructured_ingest/v2/processes/connectors/pinecone.py
@@ -94,6 +94,10 @@ class PineconeUploaderConfig(UploaderConfig):
     pool_threads: Optional[int] = Field(
         default=1, description="Optional limit on number of threads to use for upload"
     )
+    namespace: Optional[str] = Field(
+        default=None,
+        description="The namespace to write to. If not specified, the default namespace is used",
+    )
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -94,6 +94,10 @@ class PineconeUploaderConfig(UploaderConfig):
     pool_threads: Optional[int] = Field(
         default=1, description="Optional limit on number of threads to use for upload"
     )
+    namespace: Optional[str] = Field(
```
potter-potter, line 97: I would expect this to show up in the --help but it doesn't. Why is that?

potter-potter:
This works great! Just see note on --help

unstructured_ingest/v2/processes/connectors/pinecone.py
rbiseck3: I can see it when I print out the `--help` option:
```python
PYTHONPATH=. python unstructured_ingest/main.py s3 pinecone --help
...
Options:
  --batch-size FLOAT              Optional number of records per batch. Will
                                  otherwise limit by size.
  --pool-threads FLOAT            Optional limit on number of threads to use
                                  for upload  [default: 1]
  --namespace TEXT                The namespace to write to. If not specified,
                                  the default namespace is used
...
```

potter-potter: can you try `unstructured-ingest local pinecone --help` which is what I think was the traditional way.

potter-potter: looks like maybe the cli style help only hits V1?

rbiseck3: You need to install the cli fresh from the PR to see that:
```shell
pip install .
unstructured-ingest local pinecone --help
Usage: unstructured-ingest local pinecone [OPTIONS]

Options:
  --batch-size FLOAT              Optional number of records per batch. Will
                                  otherwise limit by size.
  --pool-threads FLOAT            Optional limit on number of threads to use
                                  for upload  [default: 1]
  --namespace TEXT                The namespace to write to. If not specified,
                                  the default namespace is used
...
```

potter-potter: damn. theres always something. I made a fresh venv to test something and forgot that part. Thanks!

PR #161, opened 2024-10-10
feat/Kafka V2 Connector Source

Beppeth:
Perfoming V2 migration for Kafka Source


unstructured_ingest/v2/processes/connectors/kafka.py
```diff
@@ -0,0 +1,324 @@
+import base64
+import json
+import socket
+import typing as t
```
rbiseck3, line 4: Can we use the explicit import rather than an alias? 
```suggestion
from typing import Optional, ...
```

unstructured_ingest/v2/processes/connectors/kafka.py
```diff
@@ -0,0 +1,324 @@
...
+    timeout: t.Optional[float] = 1.0
+
+
+
+class KafkaIndexerConfig(IndexerConfig):
+    pass
+
+class KafkaIndexer(Indexer):
+    connection_config = KafkaConnectionConfig
+    index_config = KafkaIndexerConfig
```
rbiseck3, line 62: This shouldn't be setting a default value, but instead setting the type. 
```suggestion
    connection_config: KafkaConnectionConfig
    index_config: KafkaIndexerConfig
```

unstructured_ingest/v2/processes/connectors/kafka.py
```diff
@@ -0,0 +1,324 @@
...
+        self.kafka_consumer.subscribe([topic])
+
+    @property
+    def kafka_consumer(self):
+        if self._consumer is None:
+            self._consumer = self.create_consumer()
+        return self._consumer
+
+    @requires_dependencies(["confluent_kafka"], extras="kafka")
+    def create_consumer(self) -> "Consumer":
```
rbiseck3, line 76: This method should live in the connection config class. Similar to many of the other `create_client` methods on other connection configs.

unstructured_ingest/v2/processes/connectors/kafka.py
```diff
@@ -0,0 +1,324 @@
...
+        pdf_data = base64.b64decode(self.raw_content)
+
+        with open(self.filename, "wb") as file:
+            file.write(pdf_data)
+
+    @SourceConnectionError.wrap
+    def run(self, file_data: FileData, **kwargs: t.Any) -> download_responses:
+        self._get_file()
+
+        return DownloadResponse(file_data=file_data, path=Path(self.filename))
```
rbiseck3, line 159: need to register this as a source entry to be part of the full list of connectors. Take a look at the bottom of any of the other connector files.

rbiseck3, line 159: Should wrap this in a `generate_download_response` call. 
```suggestion
        return self.generate_download_response(file_data=file_data, download_path=download_path)
```

unstructured_ingest/v2/processes/connectors/kafka.py
```diff
@@ -0,0 +1,324 @@
...
+    raw_content: str
+    raw_filename: str
+    registry_name: str = "kafka"
+
+    def _tmp_download_file(self):
+        topic_file = self.connection_config.topic + "-" + self.raw_filename
+        return Path(self.download_dir) / topic_file
+
+    @property
+    def version(self) -> t.Optional[str]:
```
rbiseck3, line 131: Don't need any of these `@property` methods anymore.

rbiseck3:
Might need to take a look at some of the other v2 connectors to have an idea of what the migration should look like. There's currently a lot of stuff being migrated that doens't apply anymore.

unstructured_ingest/v2/processes/connectors/kafka.py
```diff
@@ -0,0 +1,306 @@
...
+
+class KafkaAccessConfig(AccessConfig):
+    kafka_api_key: Optional[SecretStr] = Field(
+        description="Kafka API key to connect at the server", default=None
+    )
+    secret: Optional[SecretStr] = Field(description="", default=None)
+
+
+class KafkaConnectionConfig(ConnectionConfig):
+    access_config: Secret[KafkaAccessConfig]
```
rbiseck3, line 54: Let's not make the access config itself optional but instead default it to an instantiated `KafkaAccessConfig` model. Example: [pinecone](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/processes/connectors/pinecone.py#L41)

unstructured_ingest/v2/processes/connectors/kafka.py
```diff
@@ -0,0 +1,197 @@
...
+            "client.id": socket.gethostname(),
+            "group.id": "your_group_id",
+            "enable.auto.commit": "false",
+            "auto.offset.reset": "earliest",
+            "message.max.bytes": 10485760,
+        }
+
+        if is_confluent:
+            api_key = self.access_config.kafka_api_key
+            secret = self.access_config.secret
```
rbiseck3, line 73: This would cause an error. 
```suggestion
            api_key = self.access_config.get_secret_value().api_key
            secret = self.access_config.get_secret_value().secret
```

unstructured_ingest/v2/processes/connectors/kafka.py
```diff
@@ -0,0 +1,197 @@
...
+    connector_type: str = CONNECTOR_TYPE
+
+    def _get_messages(self):
+        from confluent_kafka import KafkaError
+
+        consumer = self.connection_config.get_client()
+        running = True
+
+        collected = {}
+        num_messages_to_consume = self.connection_config.num_messages_to_consume
```
rbiseck3, line 104: I don't see this value being used anywhere

unstructured_ingest/v2/processes/connectors/kafka.py
```diff
@@ -0,0 +1,197 @@
...
+        logger.info(f"config set for blocking on {num_messages_to_consume} messages")
+        # Consume specified number of messages
+        while running:
+            msg = consumer.poll(timeout=self.connection_config.timeout)
+            if msg is None:
+                logger.debug("No Kafka messages found")
+                continue
```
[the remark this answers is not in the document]
rbiseck3: Should we instead `break` here?

unstructured_ingest/v2/processes/connectors/kafka.py
```diff
@@ -0,0 +1,197 @@
...
+    pass
+
+
+@dataclass
+class KafkaIndexer(Indexer):
+    connection_config: KafkaConnectionConfig
+    index_config: KafkaIndexerConfig
+    connector_type: str = CONNECTOR_TYPE
+
+    def _get_messages(self):
```
rbiseck3, line 97: Can this have a return type specified?

unstructured_ingest/v2/processes/connectors/kafka.py
```diff
@@ -0,0 +1,197 @@
...
+
+@dataclass
+class KafkaDownloader(Downloader):
+    connection_config: KafkaConnectionConfig
+    download_config: KafkaDownloaderConfig = field(default_factory=lambda: KafkaDownloaderConfig)
+    connector_type: str = CONNECTOR_TYPE
+    version: Optional[str] = None
+    source_url: Optional[str] = None
+
+    def _tmp_download_file(self, filename: str):
```
rbiseck3, line 168: A new approach taken with `v2` is specifying the filepath and filename using the `FileData` source identifiers: [example](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/processes/connectors/fsspec/fsspec.py#L173). Can we populate that with the filepath and use it here when downloading.

unstructured_ingest/v2/processes/connectors/kafka.py
mateuszkuprowski: Implemented a bit improved message read loop where I now properly use the value

potter-potter (approved):
lg

PR #169, opened 2024-10-15
feat/reintroduce fixtures int tests

rbiseck3:
### Description
For source connector tests, save the file data and a record of the directory structure after download to compare across tests. 

This can be regenerated locally since there shouldn't be any OS discrepencies by setting the `OVERWRITE_FIXTURES=true` env var and running `make integration-test-connectors-src`


potter-potter (approved):
looks good

PR #170, opened 2024-10-15
Chore: move default API url parameter value to serverless

yuming-long:
### Summary

Point all default values of API URL to serverless API


rbiseck3 (approved):

rbiseck3:
Just need to fix the changelog issue, otherwise looks good!

PR #173, opened 2024-10-15
feat/add validation that access config always wrapped in secret

rbiseck3:
### Description
Given that the comon approach to creating new connectors is implementing concrete classes for the access config and connection config, a check was added to make sure that the type on the access config is always wrapped by a pydantic `Secret`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
-## 0.0.26-dev2
+## 0.0.26-dev3
 
 ### Enhancements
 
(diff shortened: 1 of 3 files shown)


test/unit/test_intefaces_v2.py
```diff
@@ -0,0 +1,26 @@
+import pytest
```
potter-potter, line 1: please rename this .py file. Its missing an r

potter-potter (approved):
works well. See filename correction note

PR #176, opened 2024-10-16
feat/togetherai embedder

rbiseck3:
### Description
Add support for [together.ai](https://www.together.ai/) embeddings.

As part of this PR, also refactored the embedders to move common functionality to the base class and add integration tests to make sure none of this functionality broke due to the refactor.


potter-potter (approved):
works well

PR #178, opened 2024-10-16
feat: Migrate qdrant destination to V2

guilherme-uns:
Migrates qdrant destination connector to V2 and introduces an integration test for it.

Notable changes to V1:
- utilizes asynchronous SDK
- deprecates `--num-processes`


unstructured_ingest/v2/processes/connectors/qdrant.py
```diff
@@ -0,0 +1,223 @@
...
+
+if TYPE_CHECKING:
+    from qdrant_client import QdrantClient
+
+
+CONNECTOR_TYPE = "qdrant"
+
+
+class QdrantAccessConfig(AccessConfig):
+    api_key: Optional[str] = Field(default=None, sensitive=True, description="API Key")
```
rbiseck3, line 38: remove `sensitive`, it doesn't mean anything in a pydantic `Field`

unstructured_ingest/v2/processes/connectors/qdrant.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+
+class QdrantConnectionConfig(ConnectionConfig):
+    collection_name: str
```
rbiseck3, line 43: These should all have a `Field(description=...` since the CLI text help is based on that. Copy it from the previous CLI config.

unstructured_ingest/v2/processes/connectors/qdrant.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+class QdrantConnectionConfig(ConnectionConfig):
+    collection_name: str = Field(description="Collection Name")
+    location: Optional[str] = Field(default=None, description="Location")
+    url: Optional[str] = Field(default=None, description="URL")
+    port: Optional[int] = Field(default=6333, description="Port")
+    grpc_port: Optional[int] = Field(default=6334, description="GRPC Port")
+    prefer_grpc: Optional[bool] = Field(default=False, description="Prefer GRCP")
+    https: Optional[bool] = Field(default=None, description="HTTPS")
```
rbiseck3, line 43: For these help descriptions, can we leverage better text pulled from the client sdk: [async_qdrant_client.py](https://github.com/qdrant/qdrant-client/blob/master/qdrant_client/async_qdrant_client.py#L59-L84)

unstructured_ingest/v2/processes/connectors/qdrant.py
```diff
@@ -0,0 +1,197 @@
...
+
+        logger.debug("Loaded %i elements from %s", len(elements), path)
+
+        batches = list(batch_generator(elements, batch_size=self.upload_config.batch_size))
+        logger.debug(
+            "Elements split into %i batches of size %i.",
+            len(batches),
+            self.upload_config.batch_size,
+        )
+        await asyncio.gather(*[self._upsert_batch(batch) for batch in batches])
```
rbiseck3, line 164: This never returns anything.

unstructured_ingest/v2/processes/connectors/qdrant.py
```diff
@@ -0,0 +1,197 @@
...
+            logger.debug(
+                "Upserting %i points to the '%s' collection.",
+                len(points),
+                self.connection_config.collection_name,
+            )
+            response = await client.upsert(
+                self.connection_config.collection_name, points=points, wait=True
+            )
+        except Exception as api_error:
+            logger.debug("Upsert response: %s", response)
```
rbiseck3, line 181: If this errors, `response` might not exist.

test/integration/connectors/test_qdrant.py
```diff
@@ -0,0 +1,74 @@
...
+                "size": VECTOR_SIZE,
+                "distance": VECTOR_DISTANCE,
+            },
+        )
+
+        await uploader.run_async(path=staged_upload_file, file_data=file_data)
+
+        collection = await async_client.get_collection(COLLECTION_NAME)
+
+    assert collection.points_count == EXPECTED_POINT_COUNT
```
rbiseck3, line 74: Can we add an embedder check against this vector db?

test_e2e/dest/qdrant.sh
```diff
@@ -14,7 +14,7 @@ CONTAINER_NAME="qdrant_test"
 QDRANT_PORT=6333
```
rbiseck3, line 14: We can drop the e2e test since you created the integration test.

unstructured_ingest/v2/processes/connectors/qdrant.py
ds-filipknefel: True, I'll drop that from here

test/integration/connectors/test_qdrant.py
ds-filipknefel: Can you elaborate?

ds-filipknefel: Added the embedding test which searches for a Point based on an embedding from the file and compares received text

test/integration/connectors/env_setup/qdrant/docker-compose.yaml
```diff
@@ -0,0 +1,5 @@
+services:
```
rbiseck3, line 1: If we're just running one container, you don't need to use the docker compose approach for testing but can instead use the docker library.  Take a look at how that's done in the snowflake test: [test_snowflake.py#L84](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/test/integration/connectors/sql/test_snowflake.py#L84)

test/integration/connectors/test_qdrant.py
```diff
@@ -0,0 +1,82 @@
...
+        else {"path": str(tmp_path / LOCAL_DIRECTORY_NAME)}
+    )
+
+    with open(upload_file) as file:
+        elements = json.load(file)
+        EXPECTED_POINT_COUNT = len(elements)
+        EXPECTED_TEXT = elements[0]["text"]
+        EMBEDDINGS = elements[0]["embeddings"]
+
+    with docker_compose_context(DOCKER_COMPOSE_DIR):
```
rbiseck3, line 46: Do you need docker compose running when this is using local mode? If the test approach is so different, might make sense to have this as two separate tests rather than a single one that's parameterized.

unstructured_ingest/v2/processes/connectors/qdrant.py
```diff
@@ -0,0 +1,225 @@
...
+CONNECTOR_TYPE = "qdrant"
+
+
+class QdrantAccessConfig(AccessConfig):
+    api_key: Optional[str] = Field(
+        default=None, description="API key for authentication in Qdrant Cloud."
+    )
+
+
+class QdrantConnectionConfig(ConnectionConfig):
```
rbiseck3, line 38: Can we actually break this up into a `qdrant-local` and `qdrant-hosted` to better isolate the auth needed for both, since I assume a lot of these parameters are only needed when connecting to the hosted version?

potter-potter (approved):
lgtm

PR #179, opened 2024-10-16
feat/Migration - Discord Source to Connector V2 Structure

ghost:
[Discord_Archive.zip](https://github.com/user-attachments/files/17401683/Discord_Archive.zip)

I have attached the evidence containing `download/discord/` and `workdir/discord/index & partition`

Ps: To avoid issues with the version conflict, I have set this to version `0.0.30`


unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,169 @@
...
+CONNECTOR_TYPE = "discord"
+
+
+class DiscordAccessConfig(AccessConfig):
+    token: str = Field(description="Discord API token")
+
+
+class DiscordConnectionConfig(ConnectionConfig):
+    access_config: Secret[DiscordAccessConfig]
+    channels: Optional[List[str]] = Field(
```
rbiseck3, line 32: `list` is now a native type in python
```suggestion
    channels: Optional[list[str]] = Field(
```

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,169 @@
...
+    def get_client(self):
+        import discord
+
+        intents = discord.Intents.default()
+        intents.message_content = True
+        return discord.Client(intents=intents)
+
+    def run(self, **kwargs: Any) -> Generator[FileData, None, None]:
+        client = self.get_client()
+        processed_channels: Set[str] = set()
```
rbiseck3, line 58: `set` is also a native type in python now. 
```suggestion
        processed_channels: set[str] = set()
```

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,169 @@
...
+        channels_to_process: Set[str] = set(self.connection_config.channels or [])
+
+        while channels_to_process:
```
rbiseck3, line 61: Why does this need to be a `while` loop? I don't see it adding channels to be processed to warrant the while loop since that would dynamically change the list. As far as I can tell, this is simply iterating through the list of channels and getting the file data for it.

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,169 @@
...
+
+class DiscordConnectionConfig(ConnectionConfig):
+    access_config: Secret[DiscordAccessConfig]
+    channels: Optional[List[str]] = Field(
+        default=None, description="List of Discord channel IDs to process"
+    )
+
+
+class DiscordIndexerConfig(IndexerConfig):
+    recursive: bool = Field(
```
rbiseck3, line 38: I don't see this field being used anywhere in the connector.

unstructured_ingest/v2/processes/connectors/discord.py
rbiseck3: This something maybe get list in migration that previously took the recursive flag into consideration?

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,169 @@
...
+            return FileData(
+                identifier=identifier,
+                connector_type=CONNECTOR_TYPE,
+                source_identifiers=source_identifiers,
+                metadata=metadata,
+                additional_metadata=additional_metadata,
+            )
+        except Exception as e:
+            logger.error(f"Error retrieving channel {channel_id}: {e}")
+            return None
```
rbiseck3, line 100: Rather than catch any error quietly, lets not catch it and have the downloader raise the error. I don't want records to be dropped in case something else breaks in this chunk of code.

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,169 @@
...
+    pass
+
+
+@dataclass
+class DiscordDownloader(Downloader):
+    connection_config: DiscordConnectionConfig
+    download_config: DiscordDownloaderConfig
+
+    @requires_dependencies(["discord"], extras="discord")
+    def get_client(self):
```
rbiseck3, line 113: Usually the `get_client()` method lives on the connection config class. To maintain that pattern.

unstructured_ingest/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "0.0.26-dev4"  # pragma: no cover
+__version__ = "0.0.30"  # pragma: no cover
```
potter-potter, line 1: how did it go from 26-dev4 to 30? Also please use dev versions unless it is a release.

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,156 @@
...
+        intents.message_content = True
+        return discord.Client(intents=intents)
+
+    def run(self, **kwargs: Any) -> Generator[FileData, None, None]:
+        client = self.get_client()
+        processed_channels: set[str] = set()
+        channels_to_process: set[str] = set(self.connection_config.channels or [])
+
+        for channel_id in list(channels_to_process):
+            if channel_id in processed_channels:
```
rbiseck3, line 59: I'm confused why we're maintaining a `processed_channels` list here? It doesn't seem like this is doing anything that would need it, can we not simply loop through the channels?

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,156 @@
...
+            if file_data:
+                yield file_data
+
+    @requires_dependencies(["discord"], extras="discord")
+    def get_channel_file_data(self, channel_id: str, client) -> Optional[FileData]:
+        # Fetch channel metadata
+        date_created = dt.datetime.utcnow().isoformat()
+        identifier = channel_id
+        source_identifiers = SourceIdentifiers(
+            filename=channel_id,
```
rbiseck3, line 74: Will want to have the appropriate file extension appended to the channel id here.

rbiseck3:
This PR still needs to run the ingest fixtures update CI job to regenerate the expected output files.

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,156 @@
...
+
+CONNECTOR_TYPE = "discord"
+
+
+class DiscordAccessConfig(AccessConfig):
+    token: str = Field(description="Discord API token")
+
+
+class DiscordConnectionConfig(ConnectionConfig):
+    access_config: Secret[DiscordAccessConfig]
```
rbiseck3, line 31: For consistency, can we not make this `Optional`? And default it to a blank config, similar to the pinecone connector.

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,156 @@
...
+    def run(self, **kwargs: Any) -> Generator[FileData, None, None]:
+        client = self.get_client()
+        channels_to_process: set[str] = set(self.connection_config.channels or [])
+
+        for channel_id in list(channels_to_process):
+            file_data = self.get_channel_file_data(channel_id=channel_id, client=client)
+            if file_data:
+                yield file_data
+
+    @requires_dependencies(["discord"], extras="discord")
```
rbiseck3, line 63: Don't need to wrap this in the `@requires_dependencies` since that should be isolated to methods that actually make the import call.

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,156 @@
...
+    def get_channel_file_data(self, channel_id: str, client) -> Optional[FileData]:
+        # Fetch channel metadata
+        dt.datetime.utcnow().isoformat()
+        identifier = channel_id
+        channel_id = f"{channel_id}.txt"
+        source_identifiers = SourceIdentifiers(
+            filename=identifier,
```
rbiseck3, line 70: This filename won't contain the file extension.

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,156 @@
...
+
+        for channel_id in list(channels_to_process):
+            file_data = self.get_channel_file_data(channel_id=channel_id, client=client)
+            if file_data:
+                yield file_data
+
+    @requires_dependencies(["discord"], extras="discord")
+    def get_channel_file_data(self, channel_id: str, client) -> Optional[FileData]:
+        # Fetch channel metadata
+        dt.datetime.utcnow().isoformat()
```
rbiseck3, line 66: This is never being used anywhere, should probably remove.

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,156 @@
...
+            identifier=identifier,
+            connector_type=CONNECTOR_TYPE,
+            source_identifiers=source_identifiers,
+            metadata=metadata,
+        )
+
+
+class DiscordDownloaderConfig(DownloaderConfig):
+    @requires_dependencies(["discord"], extras="discord")
+    def get_client(self):
```
rbiseck3, line 88: Lets create the `get_client` method on the connection config since that is shared between the indexer and downloader and don't need to redefine it in both places.

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,156 @@
...
+        return discord.Client(intents=intents)
+
+
+@dataclass
+class DiscordDownloader(Downloader):
+    connection_config: DiscordConnectionConfig
+    download_config: DiscordDownloaderConfig
+
+    @requires_dependencies(["discord"], extras="discord")
+    def load_async(self):
```
rbiseck3, line 102: What is this function doing? I don't see it being called anywhere.

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,156 @@
...
+
+        intents = discord.Intents.default()
+        intents.message_content = True
+        return discord.Client(intents=intents)
+
+
+@dataclass
+class DiscordDownloader(Downloader):
+    connection_config: DiscordConnectionConfig
+    download_config: DiscordDownloaderConfig
```
rbiseck3, line 99: Need to add this for the connector to be detected as `async`:
```python
    def is_async(self) -> bool:
        return True
```

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,156 @@
...
+    def load_async(self):
+        import discord
+
+        return discord.Client, None
+
+    def run(self, file_data: FileData, **kwargs: Any) -> DownloadResponse:
+        # Synchronous run is not implemented
+        raise NotImplementedError()
+
+    async def run_async(self, file_data: FileData, **kwargs: Any) -> DownloadResponse:
```
rbiseck3, line 111: Since you're importing discord here, need to wrap it with the decorator. 
```suggestion
    @requires_dependencies(["discord"], extras="discord")
    async def run_async(self, file_data: FileData, **kwargs: Any) -> DownloadResponse:
```

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,156 @@
...
+            bot = commands.Bot(command_prefix=">", intents=client.intents)
+            download_path = self.get_download_path(file_data=file_data)
+            download_path.parent.mkdir(parents=True, exist_ok=True)
+
+            messages: list[discord.Message] = []
+
+            @bot.event
+            async def on_ready():
+                logging.info("Bot is ready")
+                record_locator["channel_id"] = record_locator["channel_id"].split(".")[0]
```
rbiseck3, line 128: What does this update do? And if we need it, it should be done when the channel id is set on the file data in the indexer. Mind posting an example of when the channel id has a period in it and we need to split on it?

unstructured_ingest/v2/processes/connectors/discord.py
mr-unstructured: @rbiseck3 That was on purpose, when doing filename = channel_id, I was having issues

rbiseck3: What issues were you having? This is the typical pattern we use elsewhere and might break how the uploader writes the file.

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,148 @@
...
+
+            await bot.start(self.connection_config.access_config.get_secret_value().token)
+
+            with open(download_path, "w") as file:
+                for message in messages:
+                    file.write(f"{message.content}\n")
+
+            return self.generate_download_response(file_data=file_data, download_path=download_path)
+        else:
+            raise ValueError("No channel id in file data record locator: {record_locator}")
```
rbiseck3, line 139:
```suggestion
            raise ValueError(f"No channel id in file data record locator: {record_locator}")
```

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,148 @@
...
+
+    @requires_dependencies(["discord"], extras="discord")
+    async def run_async(self, file_data: FileData, **kwargs: Any) -> DownloadResponse:
+        import discord
+        from discord.ext import commands
+
+        client = self.connection_config.get_client()
+        record_locator = file_data.metadata.record_locator
+
+        if "channel_id" in record_locator:
```
rbiseck3, line 110: To avoid a massive if-else, can we simply break early if this condition:
```suggestion
        if "channel_id" not in record_locator:
           raise ValueError("No channel id in file data record locator: {record_locator}")
```

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,148 @@
...
+class DiscordConnectionConfig(ConnectionConfig):
+    access_config: Secret[DiscordAccessConfig] = Field(
+        default=DiscordAccessConfig, validate_default=True
+    )
+    channels: Optional[list[str]] = Field(
+        default=None, description="List of Discord channel IDs to process"
+    )
+
+    @requires_dependencies(["discord"], extras="discord")
+    def get_client(self):
```
rbiseck3, line 39: Can this have a response type:
```suggestion
if TYPE_CHECKING:
    from discord import Client as DiscordClient
    ...

    def get_client(self) -> "DiscordClient":
```

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,148 @@
...
+    def run(self, **kwargs: Any) -> Generator[FileData, None, None]:
+        client = self.connection_config.get_client()
+        channels_to_process: set[str] = set(self.connection_config.channels or [])
+
+        for channel_id in list(channels_to_process):
+            file_data = self.get_channel_file_data(channel_id=channel_id, client=client)
+            if file_data:
+                yield file_data
+
+    def get_channel_file_data(self, channel_id: str, client) -> Optional[FileData]:
```
rbiseck3, line 65: The client input here isn't being used anywhere.

unstructured_ingest/v2/processes/connectors/discord.py
rbiseck3: This was never answered?

unstructured_ingest/v2/processes/connectors/discord.py
mr-unstructured: The structured-output gets written as `.txt.json` when doing the changes you requested.
image

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,151 @@
...
+
+class DiscordAccessConfig(AccessConfig):
+    token: str = Field(description="Discord API token")
+
+
+class DiscordConnectionConfig(ConnectionConfig):
+    access_config: Secret[DiscordAccessConfig] = Field(
+        default=DiscordAccessConfig, validate_default=True
+    )
+    channels: Optional[list[str]] = Field(
```
rbiseck3, line 37: This should be part of the indexer config since it's specific to what is being downloaded rather than the connection itself.

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,149 @@
...
+    def get_client(self) -> "DiscordClient":
+        import discord
+
+        intents = discord.Intents.default()
+        intents.message_content = True
+        return discord.Client(intents=intents)
+
+
+class DiscordIndexerConfig(IndexerConfig):
+    channels: Optional[list[str]] = Field(
```
rbiseck3, line 48: Do we want to enforce at least one channel is provided? Otherwise this won't pull in anything that that might be odd to the end user.

unstructured_ingest/v2/processes/connectors/discord.py
```diff
@@ -0,0 +1,149 @@
...
+        return FileData(
+            identifier=identifier,
+            connector_type=CONNECTOR_TYPE,
+            source_identifiers=source_identifiers,
+            metadata=metadata,
+        )
+
+
+class DiscordDownloaderConfig(DownloaderConfig):
+    limit: Optional[int] = Field(
```
rbiseck3, line 88: Do we want to support a time order on this? So if limit is applied, pull in the latest messages, or the oldest? If this isn't easy to do, I won't worry about it, just a thought.

test_e2e/expected-structured-output/discord/1099442333440802930.txt.json
```diff
@@ -1,50 +1,94 @@
 [
```
rbiseck3, line 1: Since we have custom logc on how we map the data into a file, should also validate the download content: [elasticsearch test](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/test/integration/connectors/elasticsearch/test_elasticsearch.py#L201)

unstructured_ingest/v2/processes/connectors/discord.py
hubert-rutkowski85: Good point. Then this list can't be optional. And for enforcing it being non-empty (at least one entry), the UI will have check it, because I don't see a way in backend code (Indexer or Downloader) to add a config check. Let me know if I'm missing sth :)

unstructured_ingest/v2/processes/connectors/discord.py
hubert-rutkowski85: Right, so by default it takes the latest. Fortunately the method used has a sensible API to control it:
```
    async def history(
        self,
        *,
        limit: Optional[int] = 100,
        before: Optional[SnowflakeTime] = None,
        after: Optional[SnowflakeTime] = None,
        around: Optional[SnowflakeTime] = None,
        oldest_first: Optional[bool] = None,
    ) -> AsyncIterator[Message]:
    ```

File: test_e2e/expected-structured-output/discord/1099442333440802930.txt.json (hunk 26 above)
Comment (hubert-rutkowski85) [reply test_e2e/expected-structured-output/discord/1099442333440802930.txt.json:1]: Ok, then I will have to add a new integration test for Discord, as now it has only e2e test..

File: test/integration/connectors/discord/test_discord.py [hunk 27]
```diff
@@ -0,0 +1,80 @@
...
+
+
+@pytest.fixture
+def discord_token() -> str:
+    return os.environ["DISCORD_TOKEN"]
+
+
+@pytest.mark.asyncio
+@pytest.mark.tags(CONNECTOR_TYPE, SOURCE_TAG)
+async def test_discord_source(discord_token: str, discord_channels: list[str]):
```
rbiseck3, line 36: Should also annotate it with `@required_env` if we're expecting those to exist: [utils.py#L6](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/test/integration/utils.py#L6)

rbiseck3 (approved):

micmarty-deepsense:
Received confirmation from @raiderrobert that the one failing test has nothing to do with this PR's changes, so we can go ahead and merge it.

PR #181, opened 2024-10-17
feat/refactor sql connectors

rbiseck3:
### Description
To better isolate the input values needed by both sqlite and postgres, these were broken up into distinct connectors based on the same base classes to encompass any shared logic. As part of this PR, the e2e tests for each were also migrated over to the new integration test framework.


potter-potter (approved):
looks good. one note on commenting the postgres test for reference

test/integration/connectors/test_postgres.py
```diff
@@ -0,0 +1,93 @@
...
+from unstructured_ingest.v2.processes.connectors.sql.postgres import (
+    CONNECTOR_TYPE,
+    PostgresAccessConfig,
+    PostgresConnectionConfig,
+    PostgresUploader,
+    PostgresUploadStager,
+)
+
+
+def validate_destination(
```
potter-potter, line 20: Please comment this up so we can use it as a reference test for new developers.

PR #183, opened 2024-10-17
bugfix/support optional access configs on connection configs

rbiseck3:
### Description
Having a connection config implemented with the access config typing set to `Optional` breaks the mapping from flat data. This updates that code to handle it.

### DIFF
--- a/test/unit/cli/test_utils_v2.py
+++ b/test/unit/cli/test_utils_v2.py
@@ -1,0 +1,58 @@
+import json
+from typing import Optional
+
+import pytest
+from pydantic import BaseModel, Secret, ValidationError
+
+from unstructured_ingest.v2.cli.utils.click import extract_config
+
+
+def test_extract_config_optional_access_config():
+    class MyAccessConfig(BaseModel):
+        host: str
+
+    class MyModel(BaseModel):
+        v: int
+        access_config: Optional[Secret[MyAccessConfig]]
+
+    data = {
+        "v": 4,
+        "host": "localhost",
+    }
+    extracted_config = extract_config(data, MyModel)
+    assert json.loads(extracted_config.model_dump_json()) == {"v": 4, "access_config": "**********"}
+
+    data = {"v": 4}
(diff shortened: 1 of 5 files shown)


test/unit/cli/test_utils_v2.py
```diff
@@ -0,0 +1,59 @@
...
+from unstructured_ingest.v2.interfaces import AccessConfig, ConnectionConfig
+
+
+def test_extract_config_optional_access_config():
+    class MyAccessConfig(AccessConfig):
+        host: str
+
+    class MyModel(ConnectionConfig):
+        v: int
+        access_config: Optional[Secret[MyAccessConfig]]
```
potter-potter, line 17: I was expecting the Optional to be in line 13 if we had multiple AccessConfigs and some are Optional.

rbiseck3: There should only ever be one access config

rbiseck3: We already have plenty of access configs that have optional fields and that works fine.

potter-potter: I meant multiple optional fields in the AccessConfigs.  But sounds like it works fine already.

rbiseck3: Example of one that already exists that works: [milvus.py](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/processes/connectors/milvus.py#L34)

unstructured_ingest/v2/cli/utils/click.py
```diff
@@ -112,13 +112,28 @@ def convert(
 BaseModelT = TypeVar("BaseModelT", bound=BaseModel)
 
 
+def unwrap_optional(val: Any) -> tuple[Any, bool]:
```
potter-potter, line 115: This definitely makes my eyes water.

potter-potter (approved):
lgtm

PR #184, opened 2024-10-18
feat/increase unit tests

rbiseck3:
### Description
To provide validation around all the various configs used by the concrete interface implementations, the unit tests were increased to begin a pattern or randonmly generating appropriate data to make sure all of these interfaces can be instantiated without error. For now, only the partitioner and chunker were implemented but this same approach should be extended to all v2 interfaces.


potter-potter (approved):
looks good

PR #185, opened 2024-10-18
feat/sql source

rbiseck3:
### Description
Update both the sqlite and postgres implementations to support a source connector.


test/integration/connectors/sql/test_postgres.py
```diff
@@ -1,21 +1,97 @@
...
+async def test_postgres_source():
+    connect_params = {
+        "host": "localhost",
+        "port": 5433,
+        "database": "test_db",
+        "user": "unstructured",
+        "password": "test",
+    }
+    with postgres_download_setup():
+        with tempfile.TemporaryDirectory() as tmpdir:
```
potter-potter, line 66: Eventually would love to have this configurable to go to TempDirectory or a specified directory like in the .sh ingest tests. Nice to have.

test/integration/connectors/sql/test_sqlite.py
```diff
@@ -3,39 +3,99 @@
...
+                download_config=SQLiteDownloaderConfig(
+                    fields=["car_id", "brand"], download_dir=Path(tmpdir)
+                ),
+            )
+            await source_connector_validation(
+                indexer=indexer,
+                downloader=downloader,
+                configs=ValidationConfigs(
+                    test_id="sqlite",
+                    expected_num_files=40,
```
potter-potter, line 79: This is great. Would love to have another validation called length of list or something like that so that we could indeed check that only `car_id` and `brand` came through.

potter-potter (approved):
looks good. had a few suggestions for future tests.

test/integration/connectors/sql/test_postgres.py
rbiseck3: Why would we want that? This is a cleaner version of that, making sure the data is always cleanup up since for the test, we never need the contents that are written here.

test/integration/connectors/sql/test_sqlite.py
rbiseck3: Nothing is stopping us from adding an additional test after that one, I can add it in here. This just wraps the most common functionality that is shared by all source connector tests.

test/integration/connectors/sql/test_postgres.py
potter-potter: I've found it valuable to see what is being downloaded from source connectors in the past. I like to look at the .json files from the indexer and the .txt files from the downloader to troubleshoot. I can always just change it from tempdir to a named folder when I need to so not a big deal. But I always tell new connector devs to run the tests and then look at the resulting files as an example of what their "new" connector needs to be doing in the many phases.

PR #186, opened 2024-10-19
bugfix/notion fix pagination in BlocksChildrenEndpoint and DatabasesEndpoint

hardchor:
I was wondering why the notion runner was running in circles when ingesting bigger pages (with more than 100 blocks) - turns out the recursive child block getters forgot to point at the next page / cursor


rbiseck3:
I had asked this get folded into an existing PR that's migrating the notion connector to v2: https://github.com/Unstructured-IO/unstructured-ingest/pull/162

potter-potter:
@hardchor Thanks so much! We are just about to merge in a new Notion connector (we're moving all connectors to V2) and will make sure this is in there.

bryan-unstructured:
issue has been fixed in PR https://github.com/Unstructured-IO/unstructured-ingest/pull/162

PR #188, opened 2024-10-21
feat/bump unstructured-client version and leverage new async support

rbiseck3:
### Description
This bumps the min version of unstructured-client to one that added `async` support and uses that rather than an explicit event loop. 

Also adds a new config `element_exclude` to filter the elements that have these types.


potter-potter (approved):
looks good

PR #190, opened 2024-10-22
feat/onedrive-destination

mateuszkuprowski:
Added OneDrive destination connector in V2 directly (no migration). 
Added integration test in the new test structure


mateuszkuprowski:
There's a major problem with the test - I can easily pass it locally where I have all the keys exported, but it seems integration test is missing MS_CLIENT_ID env var which is required. The same var is also used in the classic shell-based tests and it works. What's the problem here?

potter-potter:
The env var wasn't in that particular test ci container: 
https://github.com/Unstructured-IO/unstructured-ingest/actions/runs/11571186346/job/32208550641#step:7:13

test/integration/connectors/conftest.py
```diff
@@ -26,3 +33,67 @@ def temp_dir() -> Generator[Path, None, None]:
         logger.info(f"Created temp dir '{temp_path}'")
         yield temp_path
         logger.info(f"Removing temp dir '{temp_path}'")
+
+
+@pytest.fixture
```
rbiseck3, line 38: Can we add these to the one drive test file since they're so specific to that test?

unstructured_ingest/v2/processes/connectors/onedrive.py
```diff
@@ -224,10 +231,113 @@ def run(self, file_data: FileData, **kwargs: Any) -> download_responses:
...
+
+        # Get the OneDrive client
+        client: GraphClient = self.connection_config.get_client()
+        drive = client.users[self.connection_config.user_pname].drive
+
+        if file_data.source_identifiers and file_data.source_identifiers.fullpath:
+            destination_path = Path(file_data.source_identifiers.relative_path)
+        else:
+            # Use a default path or raise an error
+            raise ValueError("File data does not have a valid sour path")
```
rbiseck3, line 267:
```suggestion
            raise ValueError("File data does not have a valid source path")
```

test/integration/connectors/conftest.py
mateuszkuprowski: Moved all fixtures to the test file!

mateuszkuprowski:
Realized this test was missing old style E2E testing. Added the test and adjusted everything accordingly.

test_e2e/dest/onedrive.sh
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env bash
```
rbiseck3, line 1: Since you introduces the integration test, can we remove these e2e tests?

mateuszkuprowski: At first I didn't add them just went for integration test instead, but I understood from some meeting that we still need these so I added one. Not sure if that still holds up, but if it doesn't we could just remove it.

PR #191, opened 2024-10-22
feat/snowflake connector

rbiseck3:
### Description
Support both source and destination snowflake connectors

Currently the tests won't run in CI until we have a `LOCALSTACK_AUTH_TOKEN` associated with a localstack account.  For now this was validated locally using a trial period account.


potter-potter (approved):
looks good

PR #192, opened 2024-10-23
bugfix/support async indexing

rbiseck3:
### Description
`async` support exists for the indexer, but the pipeline was only calling it assuming it was never `async`. The code was update to handle both cases.


potter-potter (approved):
lgtm

PR #193, opened 2024-10-23
fix/Databricks example.

hubert-rutkowski85:
Fixes the filenames that were being saved as hashes by uploader.

Also removed old code.


potter-potter (approved):
looks good

potter-potter:
I'll bump changelog and merge this

PR #197, opened 2024-10-24
feat/add singlestore source connector

rbiseck3:
### Description
Refactor existing single store implementation to follow the established sql-based connector pattern and add a source connector with it. 

Note:
For some reason the docker image used for SingleStore requires a ton of memory so th github runners being used for the integration tests had to be bumped to `ubuntu-latest-m`.


potter-potter (approved):
looks good

PR #200, opened 2024-10-26
File system based indexers return record display name

six5532one:
This PR updates file system based indexers to return each file's path as its display name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
-## 0.2.1-dev2
+## 0.2.2
 
 ### Enhancements
 
+* **File system based indexers return a record display name**
 * **Add singlestore source connector**
 * **Astra DB V2 Source Connector** Create a v2 version of the Astra DB Source Connector.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.2.1-dev3
+## 0.2.1
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
-## 0.2.2
+## 0.2.1-dev3
+
+### Enhancements
+
+* **File system based indexers return a record display name**
+
+## 0.2.1-dev2
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,6 @@
 ### Enhancements
 
 * **File system based indexers return a record display name**
-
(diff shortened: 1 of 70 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
-## 0.2.1-dev2
+## 0.2.2
```
six5532one, line 1: I changed the version from a dev version to a release version and rolled the enhancements from the latest dev version into this new release version. Is that ok?

rbiseck3: Can we keep this dev so I can coordinate when the next release goes out?

unstructured_ingest/v2/processes/connectors/fsspec/fsspec.py
```diff
@@ -166,6 +166,7 @@ def run(self, **kwargs: Any) -> Generator[FileData, None, None]:
 
             additional_metadata = self.sterilize_info(file_data=file_data)
             additional_metadata["original_file_path"] = file_path
+            additional_metadata["record_display_name"] = file_path
```
rbiseck3, line 169: We should make this a concrete field on the FileData class, set to an optional string defaulted to None.

rbiseck3 (changes requested):

unstructured_ingest/v2/processes/connectors/fsspec/fsspec.py
six5532one: @rbiseck3 why is this preferred?

rbiseck3: It’ll provided a common interface that can be taken advantage of downstream from this. There’s no contractual guarantee on the data that lives in the additional metadata blob.

CHANGELOG.md
```diff
@@ -1,7 +1,14 @@
+## 0.2.1-dev3
```
rbiseck3, line 1: This should replace the one below rather then add a new one.

unstructured_ingest/v2/interfaces/file_data.py
```diff
@@ -43,6 +43,7 @@ class FileData(DataClassJsonMixin):
     additional_metadata: dict[str, Any] = field(default_factory=dict)
     reprocess: bool = False
     local_download_path: Optional[str] = None
+    record_display_name: Optional[str] = None
```
rbiseck3, line 46: Would this be more succinct? 
```suggestion
    display_name: Optional[str] = None
```

rbiseck3 (approved):

PR #202, opened 2024-10-29
upgrade unstructured client to 0.26.2

jordan-homan:
### Notes
Upgrades unstructured python client to 0.26.2


potter-potter:
we're already at 0.28

PR #166, opened 2024-10-14
fix: add 'in_trash' field to Notion block, database, and page models

hardchor:
Fixes https://github.com/Unstructured-IO/unstructured-ingest/issues/56

PR #208, opened 2024-10-31
Fix notion columnlist retrieval

hardchor:
(based on https://github.com/Unstructured-IO/unstructured-ingest/pull/166)

I noticed that content from `ColumnList` was missing, so decided to dive deep into what happened. In my search, I found that the current implementation only ever goes 1 level deep into the `Column`, and so any more complex structures (like an expanding `Heading` with children) would get lost.

I refactored the page extraction to enable recursive block extraction (within the current page). There are probably other elements where this could be applied to. 

I know you're working on the v2 migration right now and appreciate you might not want to merge this PR, but just wanted to leave this here for reference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,48 +1,54 @@
+## 0.2.2-dev0
+
+### Fixes
+
+- **Fix Notion Pagination** Iterate on Notion paginated results using the `next_cursor` and `start_cursor` properties.
+
 ## 0.2.1
 
 ### Enhancements
 
-* **File system based indexers return a record display name**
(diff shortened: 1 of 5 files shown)


hardchor:
@rbiseck3 @potter-potter FYI

potter-potter:
@hardchor Thanks so much for diving into this! We are right in the middle of coming out with the Notion V2 as you mentioned and will make sure the dev working on this sees it.

bryan-unstructured:
issue has been fixed in PR https://github.com/Unstructured-IO/unstructured-ingest/pull/186

PR #212, opened 2024-10-31
feat: Neo4j Destination Connector

ds-filipknefel:
Implementation of a Neo4j Destination Connector with a generic mapping from unstructured Elements to a graph.

Example Neo4j graph mapping the input file to the Elements 
image


test/integration/connectors/test_neo4j.py
```diff
@@ -0,0 +1,215 @@
...
+EXPECTED_DOCUMENT_COUNT = 1
+EXPECTED_NODES_COUNT = 65
+DOCKER_COMPOSE_FILEPATH = env_setup_path / "neo4j" / "docker-compose.yml"
+
+
+# NOTE: Precheck tests are read-only so we utilize the same container for all tests.
+# If new tests require clean neo4j container, this fixture's scope should be adjusted.
+@pytest.fixture(autouse=True, scope="module")
+def _neo4j_server():
+    with docker_compose_context(DOCKER_COMPOSE_FILEPATH):
```
rbiseck3, line 46: Since you only need one container, this would be better using the docker utility rather than the docker compose one: [docker.py](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/test/integration/connectors/utils/docker.py#L107). There are plenty of examples of that being used.

test/integration/connectors/env_setup/neo4j/docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+services:
```
rbiseck3, line 1: As mentioned in another comment, if you use the docker utlity, we won't need this file.

test/integration/connectors/test_neo4j.py
```diff
@@ -0,0 +1,215 @@
...
+    Neo4jUploadStager,
+    Relationship,
+)
+
+USERNAME = "neo4j"
+PASSWORD = "password"
+URI = "neo4j://localhost:7687"
+DATABASE = "neo4j"
+
+EXPECTED_ELEMENT_COUNT = 42
```
rbiseck3, line 35: We should derive these values from the upload file in case that ever changes.

test/integration/connectors/test_neo4j.py
ds-filipknefel: I've replaced it with `container_context` and removed the compose file and directory.

test/integration/connectors/env_setup/neo4j/docker-compose.yml
ds-filipknefel: Removed this file

test/integration/connectors/test_neo4j.py
ds-filipknefel: I've changed constants relating to the Elements in upload file to be calculated variables instead.

test/integration/connectors/test_neo4j.py
```diff
@@ -0,0 +1,236 @@
...
+    Relationship,
+)
+
+USERNAME = "neo4j"
+PASSWORD = "password"
+URI = "neo4j://localhost:7687"
+DATABASE = "neo4j"
+
+EXPECTED_DOCUMENT_COUNT = 1
+DOCKER_COMPOSE_FILEPATH = env_setup_path / "neo4j" / "docker-compose.yml"
```
rbiseck3, line 37: Don't need this anymore, right?

PR #215, opened 2024-11-03
Elliot wood/fix azure cognitive search index

ElliotWood:
Update to AzureCognitiveSearchUploadStager based on [feedback](https://github.com/Unstructured-IO/unstructured/pull/3311#issuecomment-2451520876). For ```AzureCognitiveSearch``` we desire the index to reuse items in the index, and update them if there is a change, therefore we will keep the ```element_id``` as the key for the index instead of a unique id each time which adds duplicate entries.

    def conform_dict(data: dict) -> dict:
        """
        updates the dictionary that is from each Element being converted into a dict/json
        into a dictionary that conforms to the schema expected by the
        Azure Cognitive Search index
        """
        
        # https://docs.unstructured.io/open-source/concepts/document-elements#element-id
        # By default, the element ID is a SHA-256 hash of the element’s text, its position on the page, page number it’s on, and the name of the document file - this is to ensure that the ID is deterministic and unique at the document level. 
        # To obtain globally unique IDs in the output (UUIDs), you can pass unique_element_ids=True into any of the partition functions. This can be helpful if you’d like to use the IDs as a primary key in a database, for example.
        # For AzureCognitiveSearch we desire the index to reuse items in the index, and update them if there is a change, therefore we will keep the element_id as the key for the index instead of a unique id each time which adds duplicate entries.
        data["id"] = data.get("element_id")
        #data["id"] = str(uuid.uuid4()) // Consider using this if unique_element_ids=True

### DIFF
--- a/unstructured_ingest/connector/azure_cognitive_search.py
+++ b/unstructured_ingest/connector/azure_cognitive_search.py
@@ -1,94 +1,83 @@
 import json
+import typing as t
 import uuid
 from dataclasses import dataclass, field
-from pathlib import Path
-from typing import TYPE_CHECKING, Any
 
-from pydantic import Field, Secret
-
+from unstructured_ingest.enhanced_dataclass import enhanced_field
 from unstructured_ingest.error import DestinationConnectionError, WriteError
-from unstructured_ingest.utils.data_prep import batch_generator
-from unstructured_ingest.utils.dep_check import requires_dependencies
-from unstructured_ingest.v2.interfaces import (
+from unstructured_ingest.interfaces import (
     AccessConfig,
-    ConnectionConfig,
-    FileData,
-    Uploader,
-    UploaderConfig,
-    UploadStager,
-    UploadStagerConfig,
+    BaseConnectorConfig,
+    BaseDestinationConnector,
+    WriteConfig,
 )
-from unstructured_ingest.v2.logger import logger
-from unstructured_ingest.v2.processes.connector_registry import (
-    DestinationRegistryEntry,
-)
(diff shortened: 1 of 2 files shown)


ElliotWood:
Ready for review - I copied v2 over v1 by accident but now fixed PR.
This PR contains a fix to the Azure Search Index for reuse of index elements rather than adding new element on each run causing many duplciates.

ElliotWood:
> The issue with this approach is that the element id is not unique. Meaning there might be two elements in the same document with the same id, and this would cause one of them to be lost when uploading.

Oh interesting I had used the assumption based on this line from doco "_the element ID is a SHA-256 hash of the element’s text, its position on the page, page number it’s on, and the name of the document file - this is to ensure that the ID is deterministic and unique at the document level_". 

Can I confirm that it won't be unique at the document level?

rbiseck3:
I actually have an open PR to solve this exact issue for pinecone by persisting an id associated with the original document: [PR219](https://github.com/Unstructured-IO/unstructured-ingest/pull/219). Until we have a more deterministic approach for elements, this would alleviate duplicates if the same document gets processed. This approach can be extended to azure cognitive search and any other databases that store element-level records.

ElliotWood:
Can anyone please reivew this branch and accept on the basis that elementID is deterministic and unique at the document level.
I have tested this and confirm it works as intended.

potter-potter:
@ElliotWood can we close this PR. Are your findings fixed. Also ACS was renamed to Azure AI Search.

potter-potter:
Closed because solved with record_id fix

PR #217, opened 2024-11-04
feat/created confluence source v2 connector

mateuszkuprowski:
Added V2 version of confluence source connector

Add integration tests had their fixtures updated causing the files changed in this PR to be as high as it is. This was done because a small error was exposed by the new confluence integration test.


potter-potter (approved):
lgtm

PR #219, opened 2024-11-05
feat/persist file id in metadata, use it to delete previous content

rbiseck3:
### Description
Currently the id for each vector entry in pinecone is a randomly generated ID. TO prevent duplicates in the destination index, a step was introduces to persist the record id from the `FileData` object in the metadata and delete using that during upload to get rid of duplicates.


unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -121,19 +125,23 @@ def conform_dict(self, element_dict: dict) -> dict:
                 }
             )
 
+        metadata = flatten_dict(
+            pinecone_metadata,
+            separator="-",
+            flatten_lists=True,
+            remove_none=True,
+        )
+        metadata["file_id"] = file_data.identifier
```
vangheem, line 134: would it make sense to call this `record_id`?

Whatever we choose, we should be consistent across all connectors.

unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -208,7 +270,16 @@ def run(self, path: Path, file_data: FileData, **kwargs: Any) -> None:
...
+        # Determine if serverless or pod based index
+        pinecone_client = self.connection_config.get_client()
+        index_description = pinecone_client.describe_index(name=self.connection_config.index_name)
+        if "serverless" in index_description.get("spec"):
+            self.serverless_delete_by_record_id(file_data=file_data)
+        elif "pod" in index_description.get("spec"):
+            self.pod_delete_by_record_id(file_data=file_data)
+        else:
+            raise ValueError(f"unexpected spec type in index description: {index_description}")
+        self.serverless_delete_by_record_id(file_data=file_data)
```
vangheem, line 282: is this potentially called twice?

unstructured_ingest/v2/processes/connectors/pinecone.py
rbiseck3: Ya I wasn't sure what the best label for this was, we don't want to clash with anything that might already exist in the metadata being flattened. I'm cool with `record_id`.

unstructured_ingest/v2/processes/connectors/pinecone.py
rbiseck3: Good catch.

unstructured_ingest/v2/processes/connectors/pinecone.py
rbiseck3: Created it as a constant in a new file that will be reused across other uploaders we add this functionality to.

CHANGELOG.md
```diff
@@ -1,9 +1,10 @@
-## 0.2.2-dev1
+## 0.2.2-dev2
 
 ### Enhancements
 
 * **Remove `overwrite` field** from fsspec and databricks connectors
 * **Added migration for GitLab Source V2**
+* **Persist file id in pinecone metadata, use it to delete previous content to prevent duplicates.**
```
vangheem, line 7:
```suggestion
* **Persist `record_id` in pinecone metadata, use it to delete previous content to prevent duplicates.**
```

vangheem (approved):

PR #222, opened 2024-11-08
feat/optimize test help script

rbiseck3:
### Description
Given how many new connectors are being added recently, the help check script is growing in duration because it's checking each one serially. This updates that to use a new script to generate all commands to test and fans it out via multiprocessing.


potter-potter:
I couldn't find where this file was used in the CI tests. I couldn't find the print statements being written out. Where does this show up?

test_e2e/test-help.py
```diff
@@ -0,0 +1,80 @@
...
+
+    for dest_command_label in dest_command_labels:
+        if local_code:
+            cmds.append(
+                f"PYTHONPATH={project_path} {main_file} {first_src_command_label} {dest_command_label} --help"  # noqa: E501
+            )
+        else:
+            cmds.append(
+                f"unstructured-ingest {first_src_command_label} {dest_command_label} --help"
+            )
```
hubert-rutkowski85, line 69: It's not immediately obvious why for dest commands the unstructured-ingest script should be called, and for src not. So it would be good to add a comment for that.

hubert-rutkowski85 (approved):
LGTM, apart from the single comment thing.

PR #229, opened 2024-11-13
fix: AstraDB precheck suceeds on non-existant collections

ds-filipknefel:
### Bug description
Precheck succeeds despite collection passed in parameters not existing

### Solution
Just getting collection with [get_collection()](https://docs.datastax.com/en/astra-api-docs/_attachments/python-client/astrapy/database.html#astrapy.database.AsyncDatabase.get_collection) doesn't validate existance

> an AsyncCollection instance, representing the desired collection (but without any form of validation).

so we run [options()](https://docs.datastax.com/en/astra-api-docs/_attachments/python-client/astrapy/collection.html#astrapy.collection.AsyncCollection.options) after getting a `Collection` which will fails if collection doesn't exist.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.2.3-dev0
+
+### Fixes
+* **Make AstraDB precheck fail on non-existant collections**
+
 ## 0.2.2
 
 ### Enhancements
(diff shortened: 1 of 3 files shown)


rbiseck3 (approved):

rbiseck3:
LGTM, just need to rebase.

PR #231, opened 2024-11-13
fix/delta table check region in precheck

mackurzawa:
Extending precheck functionality by verifying if the provided AWS Region matches the region of the provided S3 bucket


unstructured_ingest/v2/processes/connectors/delta_table.py
```diff
@@ -94,14 +94,17 @@ class DeltaTableUploader(Uploader):
...
     def precheck(self):
         secrets = self.connection_config.access_config.get_secret_value()
         if (
             self.connection_config.aws_region
             and secrets.aws_access_key_id
             and secrets.aws_secret_access_key
         ):
+            from urllib.parse import urlparse
+
+            from boto3 import client
```
mackurzawa, line 107: To the best of my knowledge, fsspec does not provide a function to check the region of an S3 bucket

unstructured_ingest/v2/processes/connectors/delta_table.py
```diff
@@ -94,14 +94,17 @@ class DeltaTableUploader(Uploader):
     connection_config: DeltaTableConnectionConfig
     connector_type: str = CONNECTOR_TYPE
 
-    @requires_dependencies(["s3fs", "fsspec"], extras="s3")
+    @requires_dependencies(["s3fs", "fsspec", "boto3"], extras="s3")
```
rbiseck3, line 97: This should not require the s3 extra for the delta table connector, also can we simply do all the checks with boto3 then? Rather than needing the fsspec dependency? 
```suggestion
    @requires_dependencies(["boto3"], extras="delta-table")
```

unstructured_ingest/v2/processes/connectors/delta_table.py
rbiseck3: Also `fsspec` abstracts `s3fs`. If we know that's all we need, simply depend on that and use it rather than pulling in all of `fsspec` as a dependency. Assuming we can't use boto3 for all of the check.

unstructured_ingest/v2/processes/connectors/delta_table.py
mackurzawa: Addressed suggestion in the commit

unstructured_ingest/v2/processes/connectors/delta_table.py
```diff
@@ -94,21 +94,34 @@ class DeltaTableUploader(Uploader):
...
+                s3_client = client(
+                    "s3",
+                    aws_access_key_id=secrets.aws_access_key_id,
+                    aws_secret_access_key=secrets.aws_secret_access_key,
                 )
-                fs.write_bytes(path=self.connection_config.table_uri, value=b"")
+                s3_client.put_object(Bucket=bucket_name, Key=dir_path, Body=b"")
+
+                response = s3_client.get_bucket_location(Bucket=bucket_name)
+                assert self.connection_config.aws_region == response.get(
```
rbiseck3, line 122: This should be raising a source connection error with this as the message. Look at any of the other prechecks. `assert` should be isolated to pytests.

unstructured_ingest/v2/processes/connectors/delta_table.py
```diff
@@ -94,21 +94,34 @@ class DeltaTableUploader(Uploader):
...
             self.connection_config.aws_region
             and secrets.aws_access_key_id
             and secrets.aws_secret_access_key
         ):
-            from fsspec import get_filesystem_class
+            from urllib.parse import urlparse
+
+            from boto3 import client
 
             try:
```
rbiseck3, line 109: We should isolate only the network call in the try-except, that way if urlparse breaks, that should be raised since it doesn't mean the connection failed. Or if this is something that could be caused by user input, wrap it in a different try-except and provide a useful message in the error that the table_uri was malformed.

unstructured_ingest/v2/processes/connectors/delta_table.py
mackurzawa: I believe you wanted to say destination connection error. I changed it so it is raising `ValueError` in the try-except, which is then caught by except and raised as `DestinationConnectionError`. Check if this implementation meets your expectations or should I raise `DestinationConnectionError` already in the try-except

unstructured_ingest/v2/processes/connectors/delta_table.py
```diff
@@ -94,21 +94,34 @@ class DeltaTableUploader(Uploader):
...
+    @requires_dependencies(["boto3"], extras="delta-table")
     def precheck(self):
         secrets = self.connection_config.access_config.get_secret_value()
         if (
             self.connection_config.aws_region
             and secrets.aws_access_key_id
             and secrets.aws_secret_access_key
         ):
-            from fsspec import get_filesystem_class
+            from urllib.parse import urlparse
```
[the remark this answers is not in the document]
mackurzawa: Done

unstructured_ingest/v2/processes/connectors/delta_table.py
mackurzawa: urlparse is now excluded from try-except. I believe it can't break, because of the regex on this field in the UI that doesn't allow for any error with parsing

rbiseck3 (approved):

PR #233, opened 2024-11-13
bugfix/Add logic to handle file being downloaded as directory

rbiseck3:
### Description
When extensions have odd characters, files are downloaded into a directory. Code added to shift it around to match expected behavior.

### DIFF
--- a/test/integration/connectors/expected_results/s3-specialchar/file_data/869bf15f-e840-51dc-a818-8d0b817817c9.json
+++ b/test/integration/connectors/expected_results/s3-specialchar/file_data/869bf15f-e840-51dc-a818-8d0b817817c9.json
@@ -1,0 +1,37 @@
+{
+    "identifier": "869bf15f-e840-51dc-a818-8d0b817817c9",
+    "connector_type": "s3",
+    "source_identifiers": {
+      "filename": "Why_is_the_sky_blue?.txt",
+      "fullpath": "utic-dev-tech-fixtures/special-characters/Why_is_the_sky_blue?.txt",
+      "rel_path": "Why_is_the_sky_blue?.txt"
+    },
+    "doc_type": "file",
+    "metadata": {
+      "url": "s3://utic-dev-tech-fixtures/special-characters/Why_is_the_sky_blue?.txt",
+      "version": "15344f385570b834e468ca05bc1cebd0",
+      "record_locator": {
+        "protocol": "s3",
+        "remote_file_path": "s3://utic-dev-tech-fixtures/special-characters/"
+      },
+      "date_created": "1731627148.0",
+      "date_modified": "1731627148.0",
+      "date_processed": "1730944104.616075",
+      "permissions_data": null,
+      "filesize_bytes": 14
+    },
+    "additional_metadata": {
+      "Key": "utic-dev-tech-fixtures/special-characters/Why_is_the_sky_blue?.txt",
+      "LastModified": "2024-11-14T23:32:28+00:00",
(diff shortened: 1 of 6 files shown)


unstructured_ingest/v2/processes/connectors/fsspec/fsspec.py
```diff
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
+import os.path
```
six5532one, line 3:
```suggestion
import os

File: unstructured_ingest/v2/processes/connectors/fsspec/fsspec.py [hunk 2]
```diff
@@ -207,12 +210,34 @@ def fs(self) -> "AbstractFileSystem":
             **self.connection_config.get_access_config(),
         )
 
+    def handle_directory_download(self, lpath: Path) -> None:
+        # When the extension has odd characters in it (i.e. ???,gz), it
+        # gets downloaded in a new directory rather than as a file. This reconciles
+        # that with what is expected.
```
six5532one, line 216: Can we modify this comment? Downloading into a directory also happens when the '?' appears elsewhere in the name, not just in the extension.

```suggestion
        # If the object's name contains certain characters (i.e. '?'), it
        # gets downloaded into a new directory of the same name. This
        # reconciles that with what is expected, which is to download it 
        # as a file that is not within a directory.

**Review by six5532one — CHANGES_REQUESTED** [review]
Thank you for this! 🙌

**Review by six5532one — APPROVED** [review]

File: test/integration/connectors/test_s3.py [hunk 3]
```diff
@@ -71,6 +71,29 @@ async def test_s3_source(anon_connection_config: S3ConnectionConfig):
...
+            downloader=downloader,
+            configs=ValidationConfigs(
+                test_id="s3-specialchar",
+                predownload_file_data_check=validate_predownload_file_data,
+                postdownload_file_data_check=validate_postdownload_file_data,
+                expected_num_files=1,
+            ),
+        )
+
+
```
six5532one, line 96: This validation code extracts only files (not directories) from the download directory, then asserts they match the expected directory structure:

* https://github.com/Unstructured-IO/unstructured-ingest/blob/ca90f0de698108171904b71abb628fbdf2cea2cb/test/integration/connectors/utils/validation.py#L287-L291
* https://github.com/Unstructured-IO/unstructured-ingest/blob/ca90f0de698108171904b71abb628fbdf2cea2cb/test/integration/connectors/utils/validation.py#L127

test/integration/connectors/expected_results/s3-specialchar/file_data/869bf15f-e840-51dc-a818-8d0b817817c9.json
```diff
@@ -0,0 +1,37 @@
+{
+    "identifier": "869bf15f-e840-51dc-a818-8d0b817817c9",
+    "connector_type": "s3",
+    "source_identifiers": {
+      "filename": "Why_is_the_sky_blue?.txt",
+      "fullpath": "utic-dev-tech-fixtures/special-characters/Why_is_the_sky_blue?.txt",
```
six5532one, line 6: I created a new text file and reproduced the bug with this file. Then, I uploaded it here:
https://us-east-2.console.aws.amazon.com/s3/object/utic-dev-tech-fixtures?region=us-east-2&bucketType=general&prefix=special-characters/Why_is_the_sky_blue%3F.txt

PR #237, opened 2024-11-14
feat/Add support to delete old content in azure cognitive search

rbiseck3:
### Description
To support deleting previous content, this depends on a record id field is present in the schema and is searchable. If this criteria does not match, it does not attempt to delete the old content. A check is performed by inspecting the fields on the index being written to and checking if that field exists and is searchable. If not, deletion does not take place but a warning is logged. 

Integration tests added (e2e tests removed) to check initial upload and on supplemental upload the document count doesn't double.


test/integration/connectors/test_azure_cog_search.py
```diff
@@ -0,0 +1,235 @@
...
+    uploader.precheck()
+    uploader.run(path=staged_filepath, file_data=file_data)
+
+    # Run validation
+    with staged_filepath.open() as f:
+        staged_elements = json.load(f)
+    expected_count = len(staged_elements)
+    search_client: SearchClient = uploader.connection_config.get_search_client()
+    validate_count(search_client=search_client, expected_count=expected_count)
+
```
potter-potter, line 232: Would be awesome to delete one of the documents here and make sure it returns expected_count - 1

potter-potter (approved):
lgtm

PR #238, opened 2024-11-14
bug/add batch doc type to FileData

rbiseck3:
### Description
Some of the batch indexers were not setting the batch label on the generated FileData. This was fixed.


potter-potter (approved):
lg

PR #243, opened 2024-11-18
feat/update weaviate to support upsert

rbiseck3:
### Description
As part of the update to support the `v4` client, this was also broken up into each distinct connection type supported: `local`, `embedded` and `cloud`. For now `custom` was excluded.


potter-potter (approved):
lg

PR #244, opened 2024-11-18
feat/redis destination connector

bryan-unstructured:
- supports Redis Open Source, Enterprise Edition and cloud
- supports connection to Redis through uri or password
- If Redis instances have stack extension, write output to Redis as JSON type, otherwise write as serialized string


unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,209 @@
...
+        self,
+        elements_filepath: Path,
+        file_data: FileData,
+        output_dir: Path,
+        output_filename: str,
+        **kwargs: Any,
+    ) -> Path:
+        output_path = Path(output_dir) / Path(f"{output_filename}.json")
+        shutil.copy(elements_filepath, output_path)
+        return output_path
```
ds-filipknefel, line 74: We don't need to implement stager if it doesn't actually do any transformations.
For reference [local](https://github.com/Unstructured-IO/unstructured-ingest/blob/0dc113ca963fe543297bb22a50c0e2f0e029c2d2/unstructured_ingest/v2/processes/connectors/local.py#L204) and [fsspec](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/processes/connectors/fsspec/fsspec.py) include just `Uploader` implementations.

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,209 @@
...
+
+        try:
+            client = self.create_client()
+            client.ping()
+        except Exception as e:
+            logger.error(f"failed to validate connection: {e}", exc_info=True)
+            raise DestinationConnectionError(f"failed to validate connection: {e}")
+
+    @requires_dependencies(["redis"], extras="redis")
+    def create_client(self, async_flag=Field(default=True)) -> "Redis":
```
ds-filipknefel, line 102: We delegate clients, connections etc. creation to the `ConnectionConfig`, so this method should belong there

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,209 @@
...
+        if async_flag:
+            from redis.asyncio import Redis, from_url
+        else:
+            from redis import Redis, from_url
```
ds-filipknefel, line 106: We prefer async over sync whenever possible so you don't need to handle sync cases here given this connector is supposed to support async.

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,209 @@
...
+                password=access_config.password,
+                ssl=self.connection_config.ssl,
+            )
+        else:
+            return Redis(
+                host=self.connection_config.host,
+                port=self.connection_config.port,
+                db=self.connection_config.database,
+                ssl=self.connection_config.ssl,
+            )
```
ds-filipknefel, line 126: When repeating multiple parameters like this you can alternatively do something along the lines of:
```py
options = {"host": self.connection_config.host, "port": self.connection_config.port, ...}
if access_config.password:
    options["password"] = access_config.password
return Redis(**options)
```
it's not necessary but something to consider.

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,209 @@
...
+
+    @requires_dependencies(["redis"], extras="redis")
+    def run(self, path: Path, file_data: FileData, **kwargs: Any) -> None:
```
ds-filipknefel, line 129: If it can run in async then you can just leave out sync implementation.

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,209 @@
...
+            client = await self.create_client(async_flag=True)
+            async with client.pipeline(transaction=True) as pipe:
+                first_element = elements_dict[0]
+                element_id = first_element["element_id"]
+                redis_stack = True
+                try:
+                    await pipe.json().set(element_id, "$", first_element).execute()
+                except redis_exceptions.ResponseError:
+                    # if redis server doesn't have stack extension,
+                    # then cannot save as JSON type, save as string instead
```
ds-filipknefel, line 188: Is this a specific error? Is it possible to be more explicit here?

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,209 @@
...
+
+                for chunk in batch_generator(elements_dict[1:], self.upload_config.batch_size):
+                    for element in chunk:
+                        element_id = element["element_id"]
+                        if redis_stack:
+                            pipe.json().set(element_id, "$", element)
+                        else:
+                            pipe.set(element_id, json.dumps(element))
+                    await pipe.execute()
+            await client.aclose()
```
ds-filipknefel, line 200: Given that Redis client has a `close` implement the client providing method as a context manager, check [how qdrant does it](https://github.com/Unstructured-IO/unstructured-ingest/blob/0dc113ca963fe543297bb22a50c0e2f0e029c2d2/unstructured_ingest/v2/processes/connectors/qdrant/qdrant.py#L45) for reference

ds-filipknefel (changes requested):

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,209 @@
...
+from unstructured_ingest.v2.processes.connector_registry import DestinationRegistryEntry
+
+if TYPE_CHECKING:
+    from redis.asyncio import Redis
+
+CONNECTOR_TYPE = "redis"
+SERVER_API_VERSION = "1"
+
+
+class RedisAccessConfig(AccessConfig):
```
rbiseck3, line 31: I thought we were creating a redis connector for each auth type supported?

bryan-unstructured: Well, of the platforms I have tested, there's actually just one auth type for now: password.
For Redis on Azure, it's either password (access key) or Microsoft Entra ID (this is in another ticket to be worked in the future).
For Redis on AWS ElastiCache, it's password.
For Upstash Redis, it's password.
For local docker container, it's password.
And it applies to Redis OpenSource, Enterprise, Cloud.
Do I miss anything here?

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,155 @@
...
+class RedisUploader(Uploader):
+    upload_config: RedisUploaderConfig
+    connection_config: RedisConnectionConfig
+    connector_type: str = CONNECTOR_TYPE
+
+    def is_async(self) -> bool:
+        return True
+
+    def precheck(self) -> None:
+        if not self.connection_config.access_config.uri and not self.connection_config.host:
```
rbiseck3, line 97: This check should be on the connection config class since it's specific to what needs to exist on it. Also `self.connection_config.access_config.uri` doesn't exist, you need to get the secret value first.

rbiseck3: This precheck should run some sort of check to make sure it can communicate with the redis cluster.

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,155 @@
...
+            raise ValueError("Please pass a hostname either directly or through uri")
+
+        async def check_connection():
+            async with self.connection_config.create_client() as async_client:
+                await async_client.ping()
```
rbiseck3, line 102: Can we also check if the database we want to write to exists in this precheck?

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,155 @@
...
+                            raise e
+
+                    for chunk in batch_generator(elements_dict[1:], self.upload_config.batch_size):
+                        for element in chunk:
+                            element_id = element["element_id"]
+                            if redis_stack:
+                                pipe.json().set(element_id, "$", element)
+                            else:
+                                pipe.set(element_id, json.dumps(element))
+                        await pipe.execute()
```
rbiseck3, line 148: At the moment, none of this code is doing any async functions in parallal, meaning it's be slower than simply doing it non-async in the pipeline. We should either update the code to send the commands in parallel using something like `async.gather` or convert it to a normal sync code and let the pipeline run this with a multiprocessing fan out.

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,155 @@
...
+            raise DestinationConnectionError(f"failed to validate connection: {e}")
+
+    @requires_dependencies(["redis"], extras="redis")
+    async def run_async(self, path: Path, file_data: FileData, **kwargs: Any) -> None:
+        from redis import exceptions as redis_exceptions
+
+        with path.open("r") as file:
+            elements_dict = json.load(file)
+
+        if elements_dict:
```
rbiseck3, line 117: I prefer exiting early and not needing the rest of the code to be indented. Just a stylistic thing. 
```suggestion
        if not elements_dict:
           return
      
```

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,155 @@
...
+                async with async_client.pipeline(transaction=True) as pipe:
+                    first_element = elements_dict[0]
+                    element_id = first_element["element_id"]
+                    redis_stack = True
+                    try:
+                        # Redis with stack extension supports JSON type
+                        await pipe.json().set(element_id, "$", first_element).execute()
+                    except redis_exceptions.ResponseError as e:
+                        message = str(e)
+                        if "unknown command `JSON.SET`" in message:
```
rbiseck3, line 133: there's no way to pull this from the redis cluster before hand? By querying metadata perhaps? Just wondering, if not, this is fine.

bryan-unstructured: I checked all the commands regarding the cluster, but don't see info whether it has stack extension or supports JSON.
https://redis.io/docs/latest/commands/cluster-info/

unstructured_ingest/v2/processes/connectors/redisdb.py
bryan-unstructured: The ping() command actually checks whether the db exists in Redis. If not exists, it would raise "redis.exceptions.ResponseError: DB index is out of range".
(Unlike other databases, Redis has a fixed amount of db. User cannot create db, it's set in Redis configuration file. Each db is identified by an int like 0, 1, etc.)

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,170 @@
...
+    connector_type: str = Field(default=CONNECTOR_TYPE, init=False)
+
+    @requires_dependencies(["redis"], extras="redis")
+    @asynccontextmanager
+    async def create_client(self) -> AsyncGenerator["Redis", None]:
+        from redis.asyncio import Redis, from_url
+
+        access_config = self.access_config.get_secret_value()
+
+        if not access_config.uri and not self.host:
```
rbiseck3, line 62: Can we add this as a validation on the access config pydantic model itself?

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,170 @@
...
+        if access_config.uri:
+            client = from_url(access_config.uri)
+        elif access_config.password:
+            options["password"] = access_config.password
+            client = Redis(**options)
+        else:
+            client = Redis(**options)
+
+        try:
+            yield client
```
rbiseck3, line 82: Does the client support a contextmanager itself?

rbiseck3: It does, we're going to want to use that:
```python
with Redis(**options) as client:
    yield client
```

unstructured_ingest/v2/processes/connectors/redisdb.py
bryan-unstructured: self.host is in ConnectionConfig, so moved it to ConnectionConfig with pydantic model_validator

test/integration/connectors/test_redis.py
```diff
@@ -0,0 +1,119 @@
...
+                tasks = [delete_record(client, element["element_id"]) for element in elements]
+                await asyncio.gather(*tasks)
+        else:
+            async with Redis(**connection_kwargs, password=password) as client:
+                tasks = [delete_record(client, element["element_id"]) for element in elements]
+                await asyncio.gather(*tasks)
+
+
+@pytest.mark.asyncio
+@pytest.mark.tags(REDIS_CONNECTOR_TYPE, DESTINATION_TAG, "redis")
```
rbiseck3, line 99: Don't need that since `REDIS_CONNECTOR_TYPE` is already `redis`
```suggestion
@pytest.mark.tags(REDIS_CONNECTOR_TYPE, DESTINATION_TAG)
```

rbiseck3 (approved):

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,170 @@
...
+        return True
+
+    def precheck(self) -> None:
+
+        async def check_connection():
+            async with self.connection_config.create_client() as async_client:
+                await async_client.ping()
+
+        try:
+            asyncio.run(check_connection())
```
rbiseck3, line 107: Ideally we don't want any async code being used in here because if the precheck is called from an async context, this breaks. Can we try to leverage an async and non-async client and use the non-async client in here?

rbiseck3 (changes requested):

unstructured_ingest/v2/processes/connectors/redisdb.py
```diff
@@ -0,0 +1,170 @@
...
+        except Exception as e:
+            logger.error(f"failed to validate connection: {e}", exc_info=True)
+            raise DestinationConnectionError(f"failed to validate connection: {e}")
+
+    @requires_dependencies(["redis"], extras="redis")
+    async def run_async(self, path: Path, file_data: FileData, **kwargs: Any) -> None:
```
rbiseck3, line 113: This was also updated recently to support taking in concrete data rather than a file path. Take a look at some of the other uploaders for an example.

rbiseck3 (approved):

PR #245, opened 2024-11-18
feat/upsert elasticsearch

rbiseck3:


test_e2e/dest/elasticsearch.sh
potter-potter, line 1: Are we just assuming we test only Opensearch now? I didn't see this replaced by a `test_elasticsearch.py`

potter-potter (approved):
works well

PR #247, opened 2024-11-19
feat: Add `precheck` to Milvus connector

ds-filipknefel:
Add `precheck` method to Milvus connector together with tests verifying that it will succeed on an existent collection and fail on a nonexistent one.


unstructured_ingest/v2/processes/connectors/milvus.py
```diff
@@ -174,6 +173,14 @@ class MilvusUploader(Uploader):
     upload_config: MilvusUploaderConfig
     connector_type: str = CONNECTOR_TYPE
 
+    @DestinationConnectionError.wrap
+    def precheck(self):
+        with self.get_client() as client:
+            if not client.has_collection(self.upload_config.collection_name):
```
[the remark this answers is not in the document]
ds-filipknefel: Looking at the function docs and it's code I don't have a good reason to think that Milvus would provision a collection on [has_collection](https://milvus.io/api-reference/pymilvus/v2.4.x/MilvusClient/Collections/has_collection.md) call.

It sends a describe collection request and checks if it succeeds.

test/integration/connectors/test_milvus.py
```diff
@@ -139,3 +143,26 @@ async def test_milvus_destination(
     uploader.run(path=staged_filepath, file_data=file_data)
     with uploader.get_client() as client:
         validate_count(client=client, expected_count=expected_count)
+
+
+@pytest.mark.tags(CONNECTOR_TYPE, DESTINATION_TAG)
+class TestPrecheck:
```
rbiseck3, line 149: Why are we introducing a class based approach here?

rbiseck3: Can we stick to single test methods to keep things consistent?

rbiseck3 (approved):
Small comment to fix but otherwise LGTM

test/integration/connectors/test_milvus.py
mateuszkuprowski: I've refactored this into two separate functions, good point!

PR #252, opened 2024-11-20
[DS-285]: Add id_name field to Couchbase connection_config

mpolomdeepsense:


test_e2e/src/couchbase.sh
```diff
@@ -51,6 +51,7 @@ python "$SCRIPT_DIR"/env_setup/couchbase/source_connector/ingest_source_setup_cl
   --bucket_name "$CB_BUCKET" \
   --scope_name "$CB_SCOPE" \
   --collection_name "$CB_COLLECTION" \
+  --id_name "$CB_ID_NAME" \
```
potter-potter, line 54: don't bother setting this as an env variable. We'd have to add it to the CI secrets. Just set it here as actual text

test_e2e/src/couchbase.sh
```diff
@@ -19,8 +19,8 @@ source "$SCRIPT_DIR"/cleanup.sh
 source "$SCRIPT_DIR"/env_setup/couchbase/common/constants.env
 
 # Check if all necessary environment variables are set
-if [ -z "$CB_USERNAME" ] || [ -z "$CB_PASSWORD" ] || [ -z "$CB_CONN_STR" ] || [ -z "$CB_BUCKET" ] || [ -z "$CB_SCOPE" ] || [ -z "$CB_COLLECTION" ]; then
-  echo "Error: One or more environment variables are not set. Please set CB_CONN_STR, CB_USERNAME, CB_PASSWORD, CB_BUCKET, CB_SCOPE, and CB_COLLECTION."
+if [ -z "$CB_USERNAME" ] || [ -z "$CB_PASSWORD" ] || [ -z "$CB_CONN_STR" ] || [ -z "$CB_BUCKET" ] || [ -z "$CB_SCOPE" ] || [ -z "$CB_COLLECTION" ] || [ -z "$CB_ID_NAME" ]; then
+  echo "Error: One or more environment variables are not set. Please set CB_CONN_STR, CB_USERNAME, CB_PASSWORD, CB_BUCKET, CB_SCOPE, CB_COLLECTION and CB_ID_NAME."
```
potter-potter, line 23: remove CB_ID_NAME this just set in the file

potter-potter: Ah. just saw that is in the constants file. Disregard the above.

test_e2e/src/couchbase.sh
potter-potter: Ah. just saw that is in the constants file. Disregard the above.

potter-potter (dismissed):
looks good. works well

unstructured_ingest/v2/processes/connectors/couchbase.py
```diff
@@ -219,6 +219,7 @@ def run(self, **kwargs: Any) -> Generator[FileData, None, None]:
 
 
 class CouchbaseDownloaderConfig(DownloaderConfig):
+    id_name: str = Field(default="id", description="The name of the id field in the collection")
```
potter-potter, line 222: @ryannikolaidis should this be id_name or maybe better id_field? We need to be able to identify which field is unique. Its not always "id"

ryannikolaidis: how about `collection_id`? particularly so as not to be confused with any other ids of content in couchbase.

potter-potter:
This looks good and works well. Just trying to think of the right name. If ryan says collection_id, that should be good. Though I wonder if people might confuse it with some sort of id to identify the collection.

unstructured_ingest/v2/processes/connectors/couchbase.py
mpolomdeepsense: Will change it to `collection_id` for now, but like @potter-potter says, it might be confusing since this implies the specific id of a collection instead of a parameter name. Maybe `collection_id_key`?

potter-potter (approved):
works well

PR #253, opened 2024-11-20
feat: Change LanceDB upload mode to upsert

ds-filipknefel:
Update LanceDB upload to remove old records from destination table if currently uploaded records are sourced from the same record.

Introduces requirement on a table to have column for tracking record id which is verified in `precheck`.
In the upload a warning is raised when the column is not present in the table.


unstructured_ingest/v2/processes/connectors/lancedb/lancedb.py
```diff
@@ -134,6 +136,7 @@ async def run_async(self, path, file_data, **kwargs):
         async with self.get_table() as table:
             schema = await table.schema()
             df = self._fit_to_schema(df, schema)
+            await table.delete(f'{RECORD_ID_LABEL} = "{file_data.identifier}"')
```
rbiseck3, line 139: Can we check if it can be deleted by seeing it the column exists and log the warning the same way we did it here: [azure_ai_search](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/processes/connectors/azure_ai_search.py#L281-L285)

ds-filipknefel: I logged in a different place but it makes more sense to do it here and not run the delete at all in that case.
So added the condition and moved the log.

unstructured_ingest/v2/processes/connectors/lancedb/lancedb.py
```diff
@@ -114,9 +117,17 @@ class LanceDBUploader(Uploader):
...
-            async with self.connection_config.get_async_connection() as conn:
-                table = await conn.open_table(self.upload_config.table_name)
-                table.close()
+            async with self.get_table() as table:
+                schema = await table.schema()
+                if (
+                    RECORD_ID_LABEL not in schema.names
+                    or schema.field(RECORD_ID_LABEL).type != pa.string()
+                ):
+                    raise DestinationConnectionError(
```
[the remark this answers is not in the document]
ds-filipknefel: Removed this validation from precheck, warning log is issued on write

PR #254, opened 2024-11-20
fix: LanceDB test databses conflict in CI

ds-filipknefel:
Make directories for test databases unique per test to avoid conflicts in CI


rbiseck3 (approved):

unstructured_ingest/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "0.3.2-dev1"  # pragma: no cover
+__version__ = "0.3.2-dev2  # pragma: no cover
```
ds-filipknefel, line 1:
```suggestion
__version__ = "0.3.2-dev2"  # pragma: no cover
```

PR #256, opened 2024-11-21
fix google drive relative path

potter-potter:


test_e2e/python/test-ingest-couchbase-output.py
```diff
@@ -73,6 +73,7 @@ def check(ctx, expected_docs):
     )
     # Tally up the embeddings
     query_result = cluster.query(f"Select * from {bucket_name}.{scope_name}.{collection_name}")
+    time.sleep(5)
```
rbiseck3, line 76: We should really move this over to a python integration test because taking into account these types of wait is easier there. But this works for now.

rbiseck3 (approved):

PR #257, opened 2024-11-21
fix/Kafka cloud source couldn't connect, add test

hubert-rutkowski85:
There were 2 config errors in the Kafka Cloud source connector.

Also adding test for it. To have it work, admin will have to set secrets KAFKA_API_KEY , KAFKA_SECRET and KAFKA_BOOTSTRAP_SERVER in repo.

### DIFF
--- a/test/integration/connectors/test_kafka.py
+++ b/test/integration/connectors/test_kafka.py
@@ -1,6 +1,7 @@
 import json
 import os
 import tempfile
+import time
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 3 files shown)


unstructured_ingest/v2/processes/connectors/kafka/cloud.py
```diff
@@ -27,7 +27,7 @@
 
 class CloudKafkaAccessConfig(KafkaAccessConfig):
     api_key: Optional[SecretStr] = Field(
-        description="Kafka API key to connect at the server", alias="kafka_api_key", default=None
+        description="Kafka API key to connect at the server", default=None
```
hubert-rutkowski85, line 30: The alias `kafka_api_key` was causing the api_key to be None, even though it was set directly:

>       assert connection_config.access_config.get_secret_value().api_key.get_secret_value() == "..."
E       AttributeError: 'NoneType' object has no attribute 'get_secret_value'

hubert-rutkowski85: Ouch, now see the CI complains about "ValueError: the following field name were reused, all must be unique: api_key" (https://github.com/Unstructured-IO/unstructured-ingest/actions/runs/11956245021/job/33332171861?pr=257). I think about renaming it to `kafka_api_key` (not alias, direct rename).

unstructured_ingest/v2/processes/connectors/kafka/kafka.py
```diff
@@ -157,7 +162,9 @@ async def run_async(self, file_data: FileData, **kwargs: Any) -> DownloadRespons
     def precheck(self):
         try:
             with self.get_consumer() as consumer:
-                cluster_meta = consumer.list_topics(timeout=self.index_config.timeout)
+                # timeout needs at least 3 secs, more info:
```
rbiseck3, line 165: We should enforce this at the field level by setting the greater-than parameters: 
```python
timeout: Optional[float] = Field(default=3.0, description="polling timeout", ge=3)
```

test/integration/connectors/test_kafka.py
```diff
@@ -47,20 +57,27 @@ def docker_compose_ctx():
         yield ctx
 
 
-def wait_for_topic(topic: str, retries: int = 10, interval: int = 1):
-    admin_client = get_admin_client()
+def wait_for_topic(
+    topic: str,
+    retries: int = 10,
+    interval: int = 1,
+    exists: bool = True,
```
rbiseck3, line 64: What purpose does adding this as an input help with? Seems odd to pass this in compared to the previous approach.

hubert-rutkowski85: It makes it possible to use this function to wait for both topic existance, and not-existance. I used it in the new test.

rbiseck3 (approved):

PR #258, opened 2024-11-21
feat: Refined box connector to actually use config JSON directly

mateuszkuprowski:
This connect have been utilisng path to JSON file with config, but was given full JSON as string from the CI. I changed the connector to use the JSON content from CI directly instead of file dumping.


unstructured_ingest/v2/processes/connectors/fsspec/box.py
```diff
@@ -34,7 +35,7 @@ class BoxIndexerConfig(FsspecIndexerConfig):
 
 class BoxAccessConfig(FsspecAccessConfig):
     box_app_config: Optional[str] = Field(
-        default=None, description="Path to Box app credentials as json file."
+        default=None, description="Box app credentials as a JSON string."
```
rbiseck3, line 38: Can use this utility to leveraeg pydantic to do the validation: [google_drive](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/processes/connectors/google_drive.py#L41)

mateuszkuprowski: Good point, fixed!

unstructured_ingest/v2/processes/connectors/fsspec/box.py
```diff
@@ -33,26 +36,43 @@ class BoxIndexerConfig(FsspecIndexerConfig):
 
 
 class BoxAccessConfig(FsspecAccessConfig):
-    box_app_config: Optional[str] = Field(
-        default=None, description="Path to Box app credentials as json file."
+    box_app_config: Annotated[dict, BeforeValidator(conform_string_to_dict)] = Field(
+        description="Box app credentials as a JSON string."
     )
 
+    def model_post_init(self, __context: Any) -> None:
```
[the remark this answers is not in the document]
mateuszkuprowski: Right, good point!

unstructured_ingest/v2/processes/connectors/fsspec/box.py
```diff
@@ -44,15 +47,28 @@ class BoxConnectionConfig(FsspecConnectionConfig):
...
-        # because it is not serializable.
         from boxsdk import JWTAuth
 
         ac = self.access_config.get_secret_value()
+
+        # Use the validated and converted box_app_config
+        settings_dict = ac.box_app_config
+
+        # Create the JWTAuth object
+        oauth = JWTAuth.from_settings_dictionary(settings_dict)
```
rbiseck3, line 58: This usually takes place in the precheck method to allow working with these classes without it failing on initialization.

mateuszkuprowski: I agree, but in this case we have a fsspec-type connector which in precheck simply does super().precheck() and the Fsspec indexer does call get_access_config() during precheck I think this is actually good place to do the check. What do you think?

rbiseck3: We should override the `precheck` method in this child class to add this in the check. The `get_access_config` should be to simply get the dictionary needed to instantiate the client.

mateuszkuprowski: Okay so I did a couople of back and forths with this idea which I think sounds good, separation of resposibility is great. However there's a slight problem with this system - boxfs requires an authenticated oauth object during initialization. Hence the simplest solution I came back to is the authentication must occur in get_access_config. I mean I did try with the precheck, but I'd have to pass around authenticated oauth somehow so doing a simple change like I did [here](https://github.com/Unstructured-IO/unstructured-ingest/pull/258/commits/5f4f1376f6b1a7ac95afd6ac398bfc7aa47a8438) doesn't really work. Adding Optional[Any] arg for the token in the connection config breaks the system, but it would solve a problem if I could correctly hold authenticated token in th connection config and made sure it's always there authenticated. Well, I could duplicate code to authenticate both in precheck and in get_access_config method, but then I feel my current solution fares better. 
Can you think of any better way of working around it?

rbiseck3 (approved):

PR #261, opened 2024-11-21
feat/generate json report for pytest, use it to append to github ci output

rbiseck3:
### Description
To make it more apparent when pytest tests are getting skipped (i.e. due to missing env vars), this adds a script to markdown in a github step output which can be seen in the Github CI UI.


potter-potter (approved):
super handy thanks

PR #272, opened 2024-11-26
chore: update weaviate example

ds-filipknefel:
Update weaviate example as the original got obsolete when weaviate was split into few classes.


potter-potter (approved):
thanks for doing this

PR #273, opened 2024-11-26
[DS-303] SQL Connectors prevent syntax errors and SQL injection

mpolomdeepsense:


unstructured_ingest/v2/processes/connectors/sql/snowflake.py
```diff
@@ -132,18 +133,22 @@ class SnowflakeUploader(PostgresUploader):
...
         for rows in pd.read_json(
             path, orient="records", lines=True, chunksize=self.upload_config.batch_size
         ):
+            rows.replace({np.nan: None}, inplace=True)
             with self.connection_config.get_cursor() as cursor:
                 values = self.prepare_data(columns, tuple(rows.itertuples(index=False, name=None)))
                 # TODO: executemany break on 'Binding data in type (list) is not supported'
                 for val in values:
+                    logger.debug(f"running query: {stmt}\nwith values: {val}")
                     cursor.execute(stmt, val)
```
mpolomdeepsense, line 152: @potter-potter @rbiseck3 
We load the whole file at the beginning to read columns names, but then we load it again in batches. I feel like we could optimize it and just read columns from batches.

rbiseck3: The logic was actually fixed in the root sql implementation but I forgot to update this one. Take a look at this and reproduce it in the snowflake file as well: [sql.py#L377](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/processes/connectors/sql/sql.py#L377)

rbiseck3 (approved):

PR #271, opened 2024-11-26
fix: deregister abstract weaviate destination entry

ds-filipknefel:
### Description

Pipeline with Weaviate cannot be created because a concrete implementation of the [abstract config](https://github.com/Unstructured-IO/unstructured-ingest/blob/87a90a0f009f470168c4af9d6a8ef4929e57623a/unstructured_ingest/v2/processes/connectors/weaviate/weaviate.py#L39) is counted as two configurations due to the fact that both abstract classes and concrete classes are registered as valid destination entries. The double registration occurs [here](https://github.com/Unstructured-IO/unstructured-ingest/blob/87a90a0f009f470168c4af9d6a8ef4929e57623a/unstructured_ingest/v2/pipeline/pipeline.py#L345) on `isinstance...`.

### Fix
The base classes shouldn't be registered as they cannot be instantiated as a configuration, this PR reverts registering them as a valid destination entry.

### How to reproduce
Run the following python code on main branch

```py
from unstructured_ingest.pipeline import Pipeline
from unstructured_ingest.processor import ProcessorConfig
from unstructured_ingest.v2.interfaces import ProcessorConfig
from unstructured_ingest.v2.pipeline.pipeline import Pipeline
from unstructured_ingest.v2.processes.chunker import ChunkerConfig
from unstructured_ingest.v2.processes.connectors.local import (
    LocalConnectionConfig,
    LocalDownloaderConfig,
    LocalIndexerConfig,
)
from unstructured_ingest.v2.processes.connectors.weaviate.cloud import (
    CloudWeaviateAccessConfig,
    CloudWeaviateConnectionConfig,
    CloudWeaviateUploaderConfig,
    CloudWeaviateUploadStagerConfig,
)
from unstructured_ingest.v2.processes.embedder import EmbedderConfig
from unstructured_ingest.v2.processes.partitioner import PartitionerConfig

Pipeline.from_configs(
    context=ProcessorConfig(),
    indexer_config=LocalIndexerConfig(input_path="."),
    downloader_config=LocalDownloaderConfig(),
    source_connection_config=LocalConnectionConfig(),
    partitioner_config=PartitionerConfig(),
    chunker_config=ChunkerConfig(chunking_strategy="by_title"),
    embedder_config=EmbedderConfig(embedding_provider="openai", embedding_api_key=""),
    destination_connection_config=CloudWeaviateConnectionConfig(
        access_config=CloudWeaviateAccessConfig(api_key=""),
        cluster_url="http://localhost:4444",
    ),
[...]

PR #277, opened 2024-11-26
Don't register classes that shouldn't be instantiated

six5532one:
This is a duplicate of https://github.com/Unstructured-IO/unstructured-ingest/pull/271


potter-potter (approved):
looks good

PR #279, opened 2024-11-28
fix: FSSPEC invalid metadata date field types

ds-filipknefel:
Few fsspec connectors: SFTP, Azure, Box and GCS have`date_modified` and `date_created` fields of `FileDataSourceMetadata` class were of type `float | None` instead of `str | None`, modified code creating the metadata to cast float timestamps to strings.


potter-potter (approved):
looks good

PR #281, opened 2024-11-30
Fix azure ai search error handling

vangheem:
`azure_ai_search_key` is not a property of `IndexingResult`: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/search/azure-search-documents/azure/search/documents/_generated/models/_models_py3.py#L622


potter-potter (approved):
Lg

PR #287, opened 2024-12-04
fix: DS-328 Snowflake Downloader error

mpolomdeepsense:
[DS-328](https://unstructured-ai.atlassian.net/browse/DS-328)

Fixes `'SnowflakeCursor' object has no attribute 'mogrify'` error.

This error was discovered during development of [DS-181](https://unstructured-ai.atlassian.net/browse/DS-181) ([PR](https://github.com/Unstructured-IO/platform-workflow-builder/pull/617))

[DS-328]: https://unstructured-ai.atlassian.net/browse/DS-328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DS-181]: https://unstructured-ai.atlassian.net/browse/DS-181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


unstructured_ingest/v2/processes/connectors/sql/snowflake.py
```diff
@@ -124,7 +153,7 @@ class SnowflakeUploaderConfig(PostgresUploaderConfig):
 
 
 @dataclass
-class SnowflakeUploader(PostgresUploader):
+class SnowflakeUploader(SQLUploader):
     upload_config: SnowflakeUploaderConfig = field(default_factory=SnowflakeUploaderConfig)
```
mpolomdeepsense, line 157: @rbiseck3 @potter-potter 
A lot of Snowflake classes inherit after `postgres` connector. I think it would be best to inherit from `sql` only.

unstructured_ingest/v2/processes/connectors/sql/snowflake.py
```diff
@@ -105,10 +110,34 @@ class SnowflakeDownloaderConfig(PostgresDownloaderConfig):
...
 
 @dataclass
-class SnowflakeDownloader(PostgresDownloader):
+class SnowflakeDownloader(SQLDownloader):
     connection_config: SnowflakeConnectionConfig
     download_config: SnowflakeDownloaderConfig
     connector_type: str = CONNECTOR_TYPE
+    values_delimiter: str = "?"
+
+    @requires_dependencies(["snowflake"], extras="snowflake")
```
potter-potter, line 119: this needs to be snowflake-connector-python. this is needed because platform-plugins are 3.12 and snowflake dependency by itself does not support 3.12 (at least it didn't in the past)

unstructured_ingest/v2/processes/connectors/sql/snowflake.py
potter-potter: Agree. Let's change that for all the source inherited classes if possible.

potter-potter (approved):
approved with requested changes. works well

unstructured_ingest/v2/processes/connectors/sql/snowflake.py
mpolomdeepsense: Getting `ImportError: Following dependencies are missing: snowflake-connector-python. Please install them using pip install "unstructured-ingest[snowflake]".` but when I do pip install unstructured-ingest[snowflake] or pip install snowflake-connector-python it shows me that it's already installed. Works when I use `snowflake` inside requires_dependency.

Looks like it uses `snowflake` as a module name, that's why it works.

unstructured_ingest/v2/processes/connectors/sql/snowflake.py
mpolomdeepsense: >  for all the source inherited classes

So only Indexer and Downloader (with config)?
If yes, it's changed

unstructured_ingest/v2/processes/connectors/sql/snowflake.py
potter-potter: This worked correctly for me when I uninstalled all the snowflake stuff that was already in my virtual environment.

potter-potter: At least it installed correctly. It does use snowflake as a module name. Thats why it is strange.

potter-potter: I did this:
 1053  pip unsinstall snowflake.core==1.0.0
 1054  pip uninstall snowflake.core==1.0.0
 1055  pip uninstall snowflake==1.0.0
 1056  pip uninstall snowflake-connector-python==3.12.3
 1057  pip freeze
 1058  pip uninstall snowflake._legacy==1.0.0
 
 Can you try in a fresh venv?

mpolomdeepsense: After some debugging session, it turns out that the module name inside `snowflake-connector-python` package is named `snowflake`, and it is imported as `snowflake`. Dependency check function imports by module name and not package name, that's why the error occurred.

For now we are staying with 
```py
@requires_dependencies(["snowflake"], extras="snowflake")
```
but this issue should be resolved in the future and dependency name changed to `snowflake-connector-python`.

unstructured_ingest/v2/processes/connectors/sql/snowflake.py
mpolomdeepsense: As per Slack conversation, changed inheritance for every Snowflake class.

PR #288, opened 2024-12-05
feat/ ollama embedder

potter-potter:
Thanks to @krish-adi we have a new ollama embedder!


ratzrattillo:
Would love to see this getting merged!

PR #295, opened 2024-12-10
bugfix/prevent pinecone being called too often when deleting content

rbiseck3:
### Description
A while loop was running as often as possible when checking if content was deleting, causing a ton of read units to be used while the database caught up to the delete request. A wait was added to give the database time to catch up before making the next call.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.3.8-dev0
+
+### Fixes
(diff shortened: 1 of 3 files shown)


potter-potter (approved):
looks good. works well.

potter-potter:
Can probably remove 
MAX_QUERY_RESULTS

PR #299, opened 2024-12-11
feat/ndjson support in stagers

rbiseck3:
### Description
In a move to alleviate memory usage, all stagers were updated to optionally support ndjson files as input. If ndjson files are used, rather than reading the entire file into memory, it reads/writes one line at a time. 

Tests were added to validate the staged output for both json and ndjson files. Most of the new files in this PR are because I added stager validation against the expected file which are not stored in the `expected_results` directory.


unstructured_ingest/v2/processes/connectors/weaviate/weaviate.py
```diff
@@ -159,25 +159,6 @@ def conform_dict(cls, data: dict, file_data: FileData) -> dict:
         working_data[RECORD_ID_LABEL] = file_data.identifier
         return working_data
 
-    def run(
```
potter-potter, line 162: yay! love that all these stager runs are being consolidated.

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -325,29 +325,6 @@ def conform_dict(self, element_dict: dict, file_data: FileData) -> dict:
             "metadata": element_dict,
         }
```
potter-potter, line 327: Just curious. Never seen this. Why does line 322 have a dollar sign before vector?

potter-potter:
This is looking good and mostly works. Seems there are some tests to fix in CI related to this, like Duckdb etc...

unstructured_ingest/v2/processes/connectors/astradb.py
potter-potter: I guess its just how the sdk wants it... :)

potter-potter (approved):
looks good

PR #300, opened 2024-12-12
bugfix/make sure all SDK clients get closed

rbiseck3:
### Description
Update all connection configs that have a `get_client()` method to leverage a context manager to make sure the client gets closed if the client itself supports that functionality.


potter-potter (approved):
looks good

PR #302, opened 2024-12-13
feat/improve file data batch content

rbiseck3:
### Description
Convert the `FileData` class to be a pydantic model to have better field validation on it. Also create inherited `FileData` classes for batch support and inter-index-downloader communication.


ctrahey (commented):
Deferring comments on the actual changes for a sec - the sheer proportion of the diff dedicated to innocuous changes to test fixtures is really striking me this time around. Have we explored pulling the artifacts (and perhaps even the scenarios themselves) into a different system? Manage them like artifacts - out of band from the code?

ctrahey:
My next comment, again not truly a code level conversation yet, but closer: How do you feel about collapsing the case of single and batch into a world where _everything_ is a batch, and it just so happens that sometimes that batch is of size 1?

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -53,6 +53,15 @@
 MAX_CONTENT_PARAM_BYTE_SIZE = 8000
 
 
+class FileDataMetadata(BaseModel):
+    collection_name: str
+    keyspace: Optional[str] = None
+
+
+class AstraDBBatchFileData(BatchFileData):
+    additional_metadata: FileDataMetadata
```
ds-filipknefel, line 62: I think the `FileDataMetadata` is fine, harder to mistake then when referring to a dict but

1. Name's a little confusing, we have both metadata and additional metadata for FileData and this is named like it should be the first one in my opinion
2. I don't like this subclass, we can just use `model_validate` and `model_dump` when we want to operate on `additional_metadata`. This completely changes that field's type and right now our typing doesn't even catch that in the code below.

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -264,24 +267,23 @@ def generate_download_response(self, result: dict, file_data: FileData) -> Downl
...
 
     async def run_async(self, file_data: FileData, **kwargs: Any) -> download_responses:
         # Get metadata from file_data
-        ids: list[str] = file_data.additional_metadata["ids"]
-        collection_name: str = file_data.additional_metadata["collection_name"]
-        keyspace: str = file_data.additional_metadata["keyspace"]
+        astra_file_data = AstraDBBatchFileData.cast(file_data=file_data)
+        ids: list[str] = [item.identifier for item in astra_file_data.batch_items]
+        collection_name: str = astra_file_data.additional_metadata.collection_name
+        keyspace: str = astra_file_data.additional_metadata.keyspace
```
ds-filipknefel, line 286: I don't understand that cast, because of they way the `cast` method is typed it doesn't actually consider `astra_file_data` a `AstraDBBatchFileData` instance. 

Is there a reason to this rather then changing the `file_data` argument type and adjusting the type in superclass to be a bound TypeVar?

ds-filipknefel (commented):
Typing around functions taking and returning `FileData` is not good enough in my opinion to support this subclassing right now.

I second @ctrahey that an alternative to this of `FileData` be default describing a number of files (and often that number being 1) might be preferable to this distinction (and doesn't require subclassing).

I'd argue that if there's a lot of common information between the Batch and the specific file then we could abstract that out as <placeholder-name> and have indexer provide:

```py
"<placeholder-name>: <placeholder-name-dict>, indexed_files: list[<file-specific-data>]"
```
and have the uploader build an object with <placeholder-name> data and file-specifc data. This would require more though

I don't quite understand why `BatchItem` has no place for file-specific data. What happens when there's some metadata we'd like to add on per-file basis in indexer? 
Do we shift this responsibility on the downloader as it produces file specific `FileData` instances?

Also what is the `version` in `BatchItem` for?

unstructured_ingest/v2/processes/connectors/astradb.py
rbiseck3: This enforces that the additional metadata astra requires (i.e. the collection name) exists.

vangheem:
> My next comment, again not truly a code level conversation yet, but closer: How do you feel about collapsing the case of single and batch into a world where everything is a batch, and it just so happens that sometimes that batch is of size 1?

I would be fine with this; however, I think you're proposing a breaking change if we went that route.

I was originally thinking that the indexer should emit records where NOTHING was a batch and everything be a single Record. HOWEVER, that would mean rewriting a lot of the connectors.

vangheem (approved):

PR #303, opened 2024-12-13
feat/drop unstructured constraint

rbiseck3:
### Description
Recently the unstructured dependency changed how NLTK data was being downloaded. This dependency is not a hard requirement for this repo and so the constaint on that dependency was lifted to be able to pull in the latest, but the use of it was isolated to only the unit test CI job to prevent collisions with underlying dependencies that are needed elsewhere, rather than have it as a shared dependency with everything else using pytest.


test/integration/chunkers/test_chunkers.py
```diff
@@ -29,14 +29,3 @@ async def test_chunker_api(chunker_file: Path, strategy: str):
     chunker = Chunker(config=chunker_config)
     results = await chunker.run_async(elements_filepath=chunker_file)
     assert results
-
-
-@pytest.mark.parametrize("chunker_file", chunker_files, ids=[path.name for path in chunker_files])
-@pytest.mark.parametrize("strategy", ["basic", "by_title"])
-def test_chunker_basic(chunker_file: Path, strategy: str):
```
potter-potter, line 36: any way this can be recreated in the original Unstructured repo or somewhere, if it's being pesky here due to nltk? It's always hard to give up tests

potter-potter (approved):
looks good.

PR #306, opened 2024-12-17
fix fsspec missing param error

mpolomdeepsense:


test/unit/v2/connectors/fsspec/test_fsspec.py
```diff
@@ -0,0 +1,90 @@
...
+from unstructured_ingest.v2.interfaces import FileData, SourceIdentifiers
+from unstructured_ingest.v2.processes.connectors.fsspec.fsspec import (
+    FsspecAccessConfig,
+    FsspecConnectionConfig,
+    FsspecUploader,
+    FsspecUploaderConfig,
+)
+
+
+def test_get_upload_path_with_source_identifiers():
```
mateuszkuprowski, line 12: For my particular case of debugging, could you add a test that would check how fsspec behaves when there's a space in the directory name in the path? I have a feeling this can break things. That particular case was for dropbox, so if possible you can use that protocol and path with spaces

mateuszkuprowski (commented):
Please add a test that would help me debug things

unstructured_ingest/v2/processes/connectors/fsspec/fsspec.py
```diff
@@ -303,10 +303,14 @@ def precheck(self) -> None:
...
-            / file_data.source_identifiers.relative_path
-        )
+        upload_path_name: str
+        if file_data.source_identifiers is not None:
+            upload_path_name = file_data.source_identifiers.relative_path.lstrip("/")
+        elif file_data.local_download_path is not None:
+            upload_path_name = Path(file_data.local_download_path).name
+        else:
+            upload_path_name = file_data.identifier
+        upload_path = Path(self.upload_config.path_without_protocol) / upload_path_name
```
mpolomdeepsense, line 313: @potter-potter @rbiseck3 

We had a short discussion with @mateusz-kuprowski about this solution. Shouldn't `source_identifiers` be required for all connectors? I'm guessing it would not be obvious to come up with a path for every connector (like SQL) but this way we would create a more certain way of defining output paths.

mpolomdeepsense:
Fixed by: https://github.com/Unstructured-IO/unstructured-ingest/pull/313

PR #307, opened 2024-12-17
fix: mitigate async logging problems

mateuszkuprowski:
Added explicit try catch in run function so the exception won't get packed into common asyncio error. Also offloaded blocking IO into threads not to block async operations which should also help with catching exceptions where they belong.


unstructured_ingest/v2/processes/connectors/onedrive.py
```diff
@@ -116,25 +117,34 @@ def precheck(self) -> None:
...
+
         folders = [d for d in drive_items if d.is_folder]
         for f in folders:
-            files.extend(self.list_objects(f, recursive))
+            # Use the sync method instead of the async one here
+            files.extend(self.list_objects_sync(f, recursive))
         return files
 
-    def get_root(self, client: "GraphClient") -> "DriveItem":
+    async def list_objects(self, folder: "DriveItem", recursive: bool) -> list["DriveItem"]:
```
pawel-kmiecik, line 132: I think it could be replaced with a decorator like:
```python
def asyncify(func):
    async def inner(*args, **kwargs):
        return await asyncio.to_thread(func, *args, **kwargs)
    return inner

@asyncify
def list_objects(...)
```
I'm just not sure what linters will think about it ;)

unstructured_ingest/v2/processes/connectors/onedrive.py
```diff
@@ -176,16 +189,42 @@ def drive_item_to_file_data(self, drive_item: "DriveItem") -> FileData:
...
+        client = await asyncio.to_thread(self.connection_config.get_client)
+        root = await self.get_root(client=client)
+        drive_items = await self.list_objects(folder=root, recursive=self.index_config.recursive)
+
+        results = []
         for drive_item in drive_items:
-            file_data = self.drive_item_to_file_data(drive_item=drive_item)
-            yield file_data
+            file_data = await self.drive_item_to_file_data(drive_item=drive_item)
+            results.append(file_data)
```
pawel-kmiecik, line 214: Is this memory-safe to keep all the data in a list? I think generators were used here for a purpose.
Here's an example how to convert `async` iterator into a standard one:
https://stackoverflow.com/a/79180265

mateuszkuprowski: Thank you for that remark, I'll prepare a fix for this right away. Very good catch!

unstructured_ingest/v2/processes/connectors/onedrive.py
mateuszkuprowski: I'll try and see! Looks better :)

pawel-kmiecik: If you do - please use `from functools import wraps` to make the decorator elegant :)

unstructured_ingest/v2/processes/connectors/onedrive.py
mateuszkuprowski: This should now be fixed, please check again

unstructured_ingest/v2/processes/connectors/onedrive.py
mateuszkuprowski: I've looked into it, feels like the wrap have to be a bit complicated, still looking into it

unstructured_ingest/v2/processes/connectors/onedrive.py
```diff
@@ -100,6 +101,27 @@ class OnedriveIndexerConfig(IndexerConfig):
     recursive: bool = False
 
 
+T = TypeVar("T")
+
+
+def async_iterable_to_sync_iterable(iterator: AsyncIterator[T]) -> Iterator[T]:
+    # This version works on Python 3.9 by manually handling the async iteration.
+    loop = asyncio.new_event_loop()
```
pawel-kmiecik, line 109: Hmm why do we need to create a new loop? :O can't we reuse the default one with `asyncio.get_event_loop()`?

pawel-kmiecik: Ok, makes sense.

pawel-kmiecik (approved):
LGTM!

PR #312, opened 2024-12-19
fix: handling path containing url unicode chars

mateuszkuprowski:
This fix should allow fsspec connector to correctly handle paths containing whitespace and other url like characters like: %20.
Also slightly improved error logging for debug purposes.

### DIFF
--- a/unstructured_ingest/v2/processes/connectors/fsspec/fsspec.py
+++ b/unstructured_ingest/v2/processes/connectors/fsspec/fsspec.py
@@ -4,6 +4,7 @@ import os
 import random
 import shutil
 import tempfile
+from urllib.parse import unquote
 from dataclasses import dataclass, field
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


test/unit/v2/connectors/test_fsspec_paths.py
```diff
@@ -0,0 +1,93 @@
...
+)
+
+
+@pytest.mark.parametrize(
+    ("remote_url", "expected_path"),
+    [
+        ("dropbox://da ta", "da ta"),
+        ("dropbox://da%20ta", "da ta"),
+    ],
+)
```
mpolomdeepsense, line 21: What if my dropbox folder is named "a%20b" (literal name, not encoded)? Will it return `dropbox://a%20b` or `dropbox://a b`?

ds-filipknefel:
@mateuszkuprowski I think we can close this PR now

PR #313, opened 2024-12-19
bugfix/fix sql download file data

rbiseck3:
### Description
Most uploaders depend on the source identifier content from the `FileData` produced by the downloader. Originally this was set as optional to support the batch use case, but now that we have a different model for that, we can make source identifiers required on the base `FileData` class. This also exposed missing identifiers in the sql connectors.


nikpocuca (approved):
This is aimed at resolving bug https://unstructured-ai.atlassian.net/browse/DS-420, 

workflows will be tested on platform after pulled into staging.

PR #318, opened 2024-12-20
feat/databricks delta table destination connector

rbiseck3:
### Description
Adds in support for writing to databricks delta tables in two different ways:
* Adds in Databricks delta table destination connector, building off of the existing sql foundation. 
* Adds support for supplemental table migration in the databricks volumes connector


micmarty-deepsense:
It seems that destination entry in `unstructured_ingest/v2/processes/connectors/sql/__init__.py` is missing

test/integration/connectors/env_setup/sql/databricks_delta_tables/destination/schema.sql
```diff
@@ -0,0 +1,48 @@
...
+    data_source_version STRING,
+    data_source_record_locator STRING,
+    category_depth INT,
+    parent_id STRING,
+    attached_filename STRING,
+    filetype STRING,
+    last_modified TIMESTAMP,
+    file_directory STRING,
+    filename STRING,
+    languages ARRAY<STRING>,
```
micmarty-deepsense, line 25: arrays are not correctly inserted; prepare_data converts lists into strings.
```
[upload] [DATATYPE_MISMATCH.CAST_WITHOUT_SUGGESTION] Cannot resolve "languages" due to data type mismatch: cannot cast "STRING" to "ARRAY<STRING>". SQLSTATE: 42K09; line 1 pos 0
```

The solution is to either change the SQL statement for insertion so that we can use `ARRAY` as a keyword, or we change the type to `STRING` (which worked for me)

micmarty-deepsense: actually there's also an issue with `embeddings` column
https://github.com/databricks/databricks-sql-python/blob/01e998cabe4a4972d5b45f7936219d924706c75c/docs/parameters.md#migrating-to-native-parameters

micmarty-deepsense: Unfortunately this SDK seems to not support array parameters:
image

I don't see any other way that would be safe (not prone to SQL injection) and would allow for having `cusor.execute` and list of parameters at the same time.

rbiseck3: Ya I've reached out to a contact at Databricks. Hopefully we'll hear a work around. Otherwise this entire approach might need to be redone.

unstructured_ingest/v2/processes/connectors/sql/sql.py
```diff
@@ -406,7 +422,7 @@ def upload_dataframe(self, df: pd.DataFrame, file_data: FileData) -> None:
                 cursor.executemany(stmt, values)
 
     def get_table_columns(self) -> list[str]:
-        with self.connection_config.get_cursor() as cursor:
+        with self.get_cursor() as cursor:
             cursor.execute(f"SELECT * from {self.upload_config.table_name}")
```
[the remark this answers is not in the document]
rbiseck3: I don't actually know when this was introduced and I'm more than happy to refactor to something more optimal to fetch column data.

rbiseck3:
@micmarty-deepsense this PR is still in the works as we wait to hear back from databricks on getting a solution for variable binding and using list data.

micmarty-deepsense:
I also experienced `Timestamp` is not JSON serializable errors. Changing [this line](https://github.com/Unstructured-IO/unstructured-ingest/blob/5914bd9c51af4a450af84daac840ab93d73453bd/unstructured_ingest/v2/interfaces/upload_stager.py#L28) to 
```python
json.dump(data, f, indent=2, default=str)
```
helped, but I believe it's not the best place to fix such issues

potter-potter (approved):
lg

PR #321, opened 2025-01-02
test: qdrant upload twice to test upsert behavior

ds-filipknefel:
Extend qdrant integration tests to upload the files twice to test the upsert behavior is working.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.3.12-dev3
+## 0.3.12-dev4
 
 ### Enhancements
 
--- a/test/integration/connectors/test_qdrant.py
+++ b/test/integration/connectors/test_qdrant.py
@@ -109,6 +109,14 @@ async def test_qdrant_destination_local(upload_file: Path, tmp_path: Path):
     async with qdrant_client(connection_kwargs) as client:
         await validate_upload(client=client, upload_file=upload_file)
 
+    # Repeat upload to test the upsert functionality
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 0.3.12-dev3
+## 0.3.12-dev4
 
 ### Enhancements
```
mpolomdeepsense, line 3: Don't we want to update changelog with this fix/enchancement?

I mean, it all depends on whether we want to include dev updates here too, or only user-facing ones.

test/integration/connectors/test_qdrant.py
```diff
@@ -144,10 +152,12 @@ async def test_qdrant_destination_server(upload_file: Path, tmp_path: Path, dock
...
-    if uploader.is_async():
-        await uploader.run_async(path=staged_upload_file, file_data=file_data)
-    else:
-        uploader.run(path=upload_file, file_data=file_data)
+    await uploader.run_async(path=staged_upload_file, file_data=file_data)
+    async with qdrant_client(connection_kwargs) as client:
+        await validate_upload(client=client, upload_file=upload_file)
+
+    # NOTE: Upload the second time to validate upsert behavior
+    await uploader.run_async(path=staged_upload_file, file_data=file_data)
```
mpolomdeepsense, line 160: Just curious, why the previous change includes both async and sync run (if first one fails) but here you removed that and use async only?

PR #322, opened 2025-01-03
fix: added possibility of utilising namespaces

mateuszkuprowski:
Per request enabled the possibility of utilisng namespace in pinecone connector

### DIFF
--- a/unstructured_ingest/v2/processes/connectors/pinecone.py
+++ b/unstructured_ingest/v2/processes/connectors/pinecone.py
@@ -105,7 +105,10 @@ class PineconeUploaderConfig(UploaderConfig):
     )
     namespace: Optional[str] = Field(
         default=None,
-        description="The namespace to write to. If not specified, the default namespace is used",
+        description=(
+            "The namespace to write to. If not specified (None), the Pinecone SDK "
+            "will fall back to the 'default' namespace automatically."
+        ),
     )
     record_id_key: str = Field(
         default=RECORD_ID_LABEL,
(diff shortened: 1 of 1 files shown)


rbiseck3:
I thought we already had support for namespaces? What does this PR solve?

mateuszkuprowski:
> I thought we already had support for namespaces? What does this PR solve?

We did have it yes, but we never used it from platform perspective, only relying on default namespace behaviour so I wanted to add a test before we put this up in production as a feature (alongside minor refactors I did). Turns out my test fails, so I wonder about the functionality as a whole, trying to find if I was the problem or our upsert.

mateuszkuprowski:
Added passing test verifying that namespaces work correctly. Plus minor refactor in the upsert function and modified descriptions.

unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -173,49 +176,56 @@ def precheck(self):
...
-                    logger.error(f"failed to delete batch of ids: {delete_resp}")
-            index.delete(**delete_kwargs)
+            if self.upload_config.namespace is not None:
+                delete_kwargs["namespace"] = self.upload_config.namespace
+            delete_resp = index.delete(**delete_kwargs)
+            if delete_resp:
+                logger.error(f"Failed to delete batch of IDs: {delete_resp}")
+
         logger.info(
-            f"deleted {deleted_ids} records with metadata "
```
rbiseck3, line 216: Why were these capitalized? I think throughout the repo, we don't capitalize thee debug statements. Just a pattern that would be nice to stick with. I don't care which way we go but lets stick to one.

unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -173,49 +176,56 @@ def precheck(self):
...
+            f"from Pinecone pod index"
         )
         index = self.connection_config.get_index(pool_threads=MAX_POOL_THREADS)
+
+        # Build the delete_kwargs, only include 'namespace' if it's not None
         delete_kwargs = {
-            "filter": {self.upload_config.record_id_key: {"$eq": file_data.identifier}}
+            "filter": {self.upload_config.record_id_key: {"$eq": file_data.identifier}},
         }
-        if namespace := self.upload_config.namespace:
```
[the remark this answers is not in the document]
mateuszkuprowski: Oh that was just personal preference, but you're right I'll just reverse this change

unstructured_ingest/v2/processes/connectors/pinecone.py
mateuszkuprowski: Agreed good point, will reverse

PR #325, opened 2025-01-08
feat(confluence): update authentication parameters

amadeusz-ds:


ds-filipknefel (commented):
Both integration tests utilize api token (according to the variable names at least) and looking at the reference you've linked with this example target the cloud.
```py
    url='https://your-domain.atlassian.net', # the example
```
 Yet they do not have `cloud=True` and connect just fine. Is it that they don't connect to the cloud or maybe this parameter is resolved based on credentials and thus not necessary?

amadeusz-ds:
> Both integration tests utilize api token (according to the variable names at least) and looking at the reference you've linked with this example target the cloud.
> 
> ```python
>     url='https://your-domain.atlassian.net', # the example
> ```
> 
> Yet they do not have `cloud=True` and connect just fine. Is it that they don't connect to the cloud or maybe this parameter is resolved based on credentials and thus not necessary?

Yes, I've noticed that in my tests. I did it according to the attlasian python SDK. Maybe in some cases the token won't work without this flag. I think it's best to follow the SDK and if it works with `cloud=False` (default) then it's ok.

ds-filipknefel:
> > Both integration tests utilize api token (according to the variable names at least) and looking at the reference you've linked with this example target the cloud.
> > ```python
> >     url='https://your-domain.atlassian.net', # the example
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Yet they do not have `cloud=True` and connect just fine. Is it that they don't connect to the cloud or maybe this parameter is resolved based on credentials and thus not necessary?
> 
> Yes, I've noticed that in my tests. I did it according to the attlasian python SDK. Maybe in some cases the token won't work without this flag. I think it's best to follow the SDK and if it works with `cloud=False` (default) then it's ok.

I think it'd be best to adjust the tests to reflect how it **should** be then but otherwise LGTM.

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-## 0.3.16-dev0
+## 0.4.0-dev1
```
rbiseck3, line 1:
```suggestion
## 0.4.0
```

rbiseck3 (approved):

PR #326, opened 2025-01-09
feat: experimental transfer sharepoint authentication to AzureAD

mateuszkuprowski:
Refined authentication method to get rid of deprecated login/password with sharepoint client in favour of EntraID auth.

### DIFF
--- a/test/integration/connectors/test_sharepoint.py
+++ b/test/integration/connectors/test_sharepoint.py
@@ -45,7 +45,7 @@ async def test_sharepoint_source(temp_dir):
         permissions_application_id=client_id,
         permissions_client_cred=SecretStr(client_cred),
         permissions_tenant=tenant,
-        authority_url=SecretStr("https://login.microsoftonline.com"),
+        authority_url=SecretStr("https://login.microsoftonline.com/"),
     )
 
     connection_config = SharepointConnectionConfig(
--- a/unstructured_ingest/v2/processes/connectors/sharepoint.py
+++ b/unstructured_ingest/v2/processes/connectors/sharepoint.py
@@ -98,7 +98,8 @@ class SharepointConnectionConfig(ConnectionConfig):
         from types import SimpleNamespace
 
         try:
-            token_result = self.get_permissions_token()
+            token_scope = f"{self.site}/.default"
+            token_result = self.get_permissions_token(token_scope=token_scope)
             if "access_token" not in token_result:
                 raise SourceConnectionNetworkError(
                     f"Failed to obtain token for SharePoint: \
@@ -117,18 +118,18 @@ class SharepointConnectionConfig(ConnectionConfig):
         return site_client
 
     @requires_dependencies(["msal"], extras="sharepoint")
-    def get_permissions_token(self):
+    def get_permissions_token(self, token_scope: Optional[str] = "https://graph.microsoft.com/.default"):
         from msal import ConfidentialClientApplication
 
         try:
             client_credential = self.permissions_config.permissions_client_cred.get_secret_value()
             app = ConfidentialClientApplication(
-                authority=f"{self.permissions_config.authority_url.get_secret_value()}/"
+                authority=f"{self.permissions_config.authority_url.get_secret_value()}"
                 f"{self.permissions_config.permissions_tenant}",
                 client_id=self.permissions_config.permissions_application_id,
                 client_credential=client_credential,
             )
-            token = app.acquire_token_for_client(scopes=["https://graph.microsoft.com/.default"])
+            token = app.acquire_token_for_client(scopes=[token_scope])
         except ValueError as exc:
             logger.error("Couldn't set up credentials for Sharepoint")
             raise exc


mateuszkuprowski:
I've taken the route of minimal changes to the whole structure. I've been reading and experimenting a bit and came up with this solution. It feels elegant and should cover what we need - changed the authorisation method without changing the whole code structure. The doownside of this is that tenant_id which is required when using sharepoint's REST API had to be taken from permissions, which is fine for tests, but we'd have to ensure that it works correctly with entire system via UI as well. 
Hence I'm asking for you opinion - can we stick with it as is or should we maybe create separate parameter for tenant_id besides permissions? That'd most likely duplicate which isn't ideal either, but would have to be this way since permissions are using graph api instead of REST api that has different capabilities.

test/integration/connectors/test_sharepoint.py
```diff
@@ -0,0 +1,88 @@
...
+    client_cred = os.getenv("SHAREPOINT_CRED")
+    tenant = os.getenv("SHAREPOINT_PERMISSIONS_TENANT")
+    site_url = os.getenv("SHAREPOINT_SITE")
+
+    access_config = SharepointAccessConfig(client_cred=client_cred)
+
+    permissions_config = SharepointPermissionsConfig(
+        permissions_application_id=None,
+        permissions_tenant=tenant,
+        permissions_client_cred=None,  # or SecretStr(...)
```
bryan-unstructured, line 47: should we provide with a value for permissions_client_cred and permissions_application_id here? otherwise this integration test would fail in CI pipeline

unstructured_ingest/v2/processes/connectors/sharepoint.py
```diff
@@ -91,16 +91,33 @@ class SharepointConnectionConfig(ConnectionConfig):
...
         from office365.sharepoint.client_context import ClientContext
 
         try:
-            credentials = ClientCredential(
-                self.client_id, self.access_config.get_secret_value().client_cred
+            # Acquire the token using MSAL, similar to get_permissions_token
+            app = ConfidentialClientApplication(
+                authority=f"{self.permissions_config.authority_url.get_secret_value()}/"
+                f"{self.permissions_config.permissions_tenant}",
+                client_id=self.permissions_config.permissions_application_id,
```
bryan-unstructured, line 104: should client_id be self.client_id instead of permissions_application_id?

unstructured_ingest/v2/processes/connectors/sharepoint.py
```diff
@@ -91,16 +91,33 @@ class SharepointConnectionConfig(ConnectionConfig):
...
+                client_credential=self.permissions_config.permissions_client_cred.get_secret_value(),  # noqa: E501
+            )
+            token_result = app.acquire_token_for_client(
+                scopes=[
+                    f"https://{self.permissions_config.permissions_tenant}.sharepoint.com/.default"
```
bryan-unstructured, line 109: when I use SHAREPOINT_PERMISSIONS_TENANT value in Keeper, this line reports an error. After I remove the second part of the string and only keeps the first part, it works.

bryan-unstructured:
> I've taken the route of minimal changes to the whole structure. I've been reading and experimenting a bit and came up with this solution. It feels elegant and should cover what we need - changed the authorisation method without changing the whole code structure. The doownside of this is that tenant_id which is required when using sharepoint's REST API had to be taken from permissions, which is fine for tests, but we'd have to ensure that it works correctly with entire system via UI as well. Hence I'm asking for you opinion - can we stick with it as is or should we maybe create separate parameter for tenant_id besides permissions? That'd most likely duplicate which isn't ideal either, but would have to be this way since permissions are using graph api instead of REST api that has different capabilities.

on the platform UI, what's differences between permissions and AccessConfig? Aren't both of them shown on the UI?

test/integration/connectors/test_sharepoint.py
mateuszkuprowski: It won't actually fail (and it doesn't) as this connector can work in 2 modes, either with or without permissions. There's a separate e2e test for that in the test_e2e/dest/sharepoint.sh and sharepoint_with_permissions.sh

unstructured_ingest/v2/processes/connectors/sharepoint.py
mateuszkuprowski: Oh! what a great catch, I must have been moving things around and missed it!

mateuszkuprowski: Corrected

unstructured_ingest/v2/processes/connectors/sharepoint.py
mateuszkuprowski: I might need a little more explanation on this not sure which part do you mean? Like it works without the /.default/ but it doesn't with it?

mateuszkuprowski:
> > I've taken the route of minimal changes to the whole structure. I've been reading and experimenting a bit and came up with this solution. It feels elegant and should cover what we need - changed the authorisation method without changing the whole code structure. The doownside of this is that tenant_id which is required when using sharepoint's REST API had to be taken from permissions, which is fine for tests, but we'd have to ensure that it works correctly with entire system via UI as well. Hence I'm asking for you opinion - can we stick with it as is or should we maybe create separate parameter for tenant_id besides permissions? That'd most likely duplicate which isn't ideal either, but would have to be this way since permissions are using graph api instead of REST api that has different capabilities.
> 
> on the platform UI, what's differences between permissions and AccessConfig? Aren't both of them shown on the UI?

Unfortunately there's no permission config to be setup in the UI. All I need is tenant for this to work. Actually you indirectly input a tenant when you give link to the proper site, I was wondering if simple copy that info to tenant wouldn't work. In tests it's exactly like that.

PR #333, opened 2025-01-15
feat/break up int tests for CI

rbiseck3:
### Description
By splitting up the CI tests into these categories provides the following benefits:
* Speed up in CI since now more runners can be used to run the tests in parallel
* Secrets are better organized for each test since the same provider is now tested in one job rather than being split between src/dest. 

This PR also adds in a check that all tests defined in the integration connectors directory are properly tagged


potter-potter (approved):
looks great

PR #343, opened 2025-01-21
feat/fsspec file data from uri

rbiseck3:


ctrahey:
@rbiseck3 do you think its possible to move this up from just fsspec to being available on all connectors? I understand some of them might be meaningless or close to it, but if we embrace the universality of URIs we should be able to take advantage of this capability for more than just fsspec. 

I think the major complication would be returning many records optionally. I sorta think we should always return a list anyway, which would then allow a connector to decide how many records a URI refers to. Similar with the hydration method - it should take a list in and put a list out, and not necessarily of the same length.

PR #346, opened 2025-01-21
feat/html file extraction

rbiseck3:
### Description
Adds a utility to parse out all hrefs from html content and downloads them as new files to be processed.


potter-potter (approved):
looks good

PR #348, opened 2025-01-23
feat/vastdb connector

potter-potter:
Connector for vast db big data database.


unstructured_ingest/v2/processes/connectors/sql/vastdb.py
```diff
@@ -0,0 +1,281 @@
...
+                table.insert(pa_table)
+
+    def get_table_columns(self) -> list[str]:
+        if self._columns is None:
+            with self.get_cursor() as cursor:
+                table = self.connection_config.get_table(cursor, self.upload_config.table_name)
+                self._columns = table.columns().names
+        return self._columns
+
+    @requires_dependencies(["vastdb", "ibis", "pyarrow"], extras="vastdb")
```
rbiseck3, line 248:
```suggestion
    @requires_dependencies(["ibis"], extras="vastdb")
```
Since we're only importing the `ibis` dependency in this method.

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
```diff
@@ -0,0 +1,281 @@
...
+
+class VastdbConnectionConfig(SQLConnectionConfig):
+    access_config: Secret[VastdbAccessConfig] = Field(
+        default=VastdbAccessConfig(), validate_default=True
+    )
+    vastdb_bucket: str
+    vastdb_schema: str
+    connector_type: str = Field(default=CONNECTOR_TYPE, init=False)
+
+    @requires_dependencies(["vastdb", "ibis", "pyarrow"], extras="vastdb")
```
rbiseck3, line 61:
```suggestion
    @requires_dependencies(["vastdb"], extras="vastdb")
```
Since we're only importing `vastdb` in this method.

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
```diff
@@ -0,0 +1,281 @@
...
+class VastdbIndexer(SQLIndexer):
+    connection_config: VastdbConnectionConfig
+    index_config: VastdbIndexerConfig
+    connector_type: str = CONNECTOR_TYPE
+
+    def _get_doc_ids(self) -> list[str]:
+        with self.get_cursor() as cursor:
+            table = self.connection_config.get_table(cursor, self.index_config.table_name)
+            reader = table.select(columns=[self.index_config.id_column])
+            results = reader.read_all()  # Build a PyArrow Table from the RecordBatchReader
```
rbiseck3, line 97: Is there no way to limit this to fetching only IDs? Otherwise we're going to be reading the entire table in twice, once to index, and once to download.

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
```diff
@@ -0,0 +1,281 @@
...
+class VastdbConnectionConfig(SQLConnectionConfig):
+    access_config: Secret[VastdbAccessConfig] = Field(
+        default=VastdbAccessConfig(), validate_default=True
+    )
+    vastdb_bucket: str
+    vastdb_schema: str
+    connector_type: str = Field(default=CONNECTOR_TYPE, init=False)
+
+    @requires_dependencies(["vastdb", "ibis", "pyarrow"], extras="vastdb")
+    def get_connection(self) -> "VastdbConnect":
```
rbiseck3, line 62: These need `@contextmanager` to match the parent class implementation and yield them.

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
```diff
@@ -0,0 +1,281 @@
...
+        data.update(coordinates)
+
+        data["id"] = get_enhanced_element_id(element_dict=data, file_data=file_data)
+
+        # remove extraneous, not supported columns
+        # but also allow for additional columns
+        approved_columns = set(list(_COLUMNS) + self.upload_stager_config.additional_columns)
+        element = {k: v for k, v in data.items() if k in approved_columns}
+        element[RECORD_ID_LABEL] = file_data.identifier
+        return element
```
rbiseck3, line 171: Most of this code can be reused from the parent class implementation. Looks like this only adds the use if the additional colunmns: 
```suggestion
    def conform_dict(self, element_dict: dict, file_data: FileData) -> dict:
        conformed_dict = super().conform_dict(element_dict=element_dict, file_data=file_data)
        additional_keys = {k: v for k, v in element_dict.items() if k in self.upload_stager_config.additional_columns}
        conformed_dict.update(additional_keys)
        return conformed_dict
```

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
```diff
@@ -0,0 +1,281 @@
...
+                lambda x: json.dumps(x) if isinstance(x, (list, dict)) else None
+            )
+        for column in filter(
+            lambda x: x in df.columns,
+            ("version", "page_number", "regex_metadata"),
+        ):
+            df[column] = df[column].apply(str)
+        if self.upload_stager_config.rename_columns_map:
+            df.rename(columns=self.upload_stager_config.rename_columns_map, inplace=True)
+        return df
```
rbiseck3, line 190: Same concept here, let's reuse as much of the parent implementation as possible:
```suggestion
    def conform_dataframe(self, df: pd.DataFrame) -> pd.DataFrame:
        df = super().conform_dataframe(df=df)
        if self.upload_stager_config.rename_columns_map:
            df.rename(columns=self.upload_stager_config.rename_columns_map, inplace=True)
        return df
```

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
```diff
@@ -0,0 +1,281 @@
...
+class VastdbUploader(SQLUploader):
+    upload_config: VastdbUploaderConfig = field(default_factory=VastdbUploaderConfig)
+    connection_config: VastdbConnectionConfig
+    connector_type: str = CONNECTOR_TYPE
+
+    def precheck(self) -> None:
+        try:
+            with self.get_cursor() as cursor:
+                table = self.connection_config.get_table(cursor, self.upload_config.table_name)
+                table.select()
```
rbiseck3, line 207: Does running `table.select()` act the same as `cursor.execute("SELECT 1;")`? We want to make sure it's not making a heavy request as part of the precheck.

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
```diff
@@ -0,0 +1,281 @@
...
+            f"writing a total of {len(df)} elements via"
+            f" document batches to destination"
+            f" table named {self.upload_config.table_name}"
+            f" with batch size {self.upload_config.batch_size}"
+        )
+
+        for rows in split_dataframe(df=df, chunk_size=self.upload_config.batch_size):
+
+            with self.get_cursor() as cursor:
+                pa_table = pa.Table.from_pandas(df)
```
rbiseck3, line 237: I assume this needs to be rows, otherwise you're uploading the entire table for each batch. 
```suggestion
                pa_table = pa.Table.from_pandas(rows)
```

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
```diff
@@ -0,0 +1,281 @@
...
+            f"from table {self.upload_config.table_name}"
+        )
+        predicate = _[self.upload_config.record_id_key].isin([file_data.identifier])
+        with self.get_cursor() as cursor:
+            table = self.connection_config.get_table(cursor, self.upload_config.table_name)
+            # Get the internal row id
+            rows_to_delete = table.select(
+                columns=[], predicate=predicate, internal_row_id=True
+            ).read_all()
+            table.delete(rows_to_delete)
```
rbiseck3, line 264: Does this return a count of how many rows were deleted? Having that as a log statement is nice to have as a sanity check.

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
potter-potter: I couldn't find anything to limit 1. I definitely tried. I think its because it is a columnar instead of table storage the concept is not really there.

I believe as long as we just select and not table.read_all() then it is lazy. It never actually pulls down the records.

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
potter-potter: thank you!

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
rbiseck3: Then does this actually validate anything if it's lazy? Have you tried running this with the wrong table name and checking that if fails?

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
potter-potter: It returns none. Bummer.
It does work at least :) 
But the crap thing is when I do a row count after delete, includes deleted rows. I guess they have some internal row id even after they are deleted.

potter-potter: https://vastdb-sdk.readthedocs.io/en/v1.1.0/table.html#vastdb.table.Table.delete

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
potter-potter: I like it but we would still have to "flatten" element_dict again since the metadata needs to be flattened.

I will probably keep it. Unless we want to make the flattening a whole different method... in the sql file

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
potter-potter: oh yes id def fails if it doesn't find table.

potter-potter: I double checked. Works as expected.

unstructured_ingest/v2/processes/connectors/sql/vastdb.py
potter-potter: row_count does "eventually" update to what we would expect. Works just like count() eventually.

unstructured_ingest/v2/processes/connectors/sql/sql.py
```diff
@@ -398,7 +398,7 @@ def upload_dataframe(self, df: pd.DataFrame, file_data: FileData) -> None:
                 f"{self.upload_config.record_id_key}, skipping delete"
             )
         df.replace({np.nan: None}, inplace=True)
-        self._fit_to_schema(df=df)
+        df = self._fit_to_schema(df=df)
```
potter-potter, line 401: So fit_to_schema never worked. until I added this line. (pandas dataframes have this weird inplace inconsistency)
Now the tests are failing. It has to do with the date fields.

potter-potter: fixed in the prepare_data. date comes in as a pandas float but a none. so a nan

rbiseck3 (approved):

PR #349, opened 2025-01-23
feat(DS-412): fix databricks volume delta tables uploader

mpolomdeepsense:
This fix fixes an issue with the Databricks Volume Delta Tables uploader. When uploading data, it did not select a given database; it always picked a default one.


Paul-Cornell:
Looks like there are several dependent files that start with a class name of `DatabrickDelta` (no `s`). It should be `DatabricksDelta`.

mpolomdeepsense:
> Looks like there are several dependent files that start with a class name of DatabrickDelta (no s). It should be DatabricksDelta.

Changed

amadeusz-ds (approved):
LGTM

PR #350, opened 2025-01-23
bugfix/use columns from database to adjust data to upload

rbiseck3:
### Description
The `_fit_to_schema` method on the sql connector base class was never taking into account the columns on the destination table. These were pulled in to drop any columns that would cause the upload to fail.


potter-potter (approved):
Looks good

PR #351, opened 2025-01-23
feat/add support for allow list when downloading from raw urls

rbiseck3:


unstructured_ingest/utils/html.py
```diff
@@ -1,109 +1,157 @@
...
-        result_file_data.metadata.url = source_url
-        result_file_data.metadata.record_locator["parent_url"] = url
-        result_file_data.identifier = str(
-            uuid5(NAMESPACE_DNS, source_url + original_filedata.identifier)
+        return response.content
+
+    def can_download(self, url_to_download: str, original_url: str) -> bool:
+        parsed_original_url = urlparse(original_url)
+        base_url = parsed_original_url.scheme + "://" + parsed_original_url.netloc
+        allow_list = self.allow_list or [base_url]
```
potter-potter, line 62: if their allow list is populated, but they didn't include their base_url in that list, will it download images from their base url?

rbiseck3: Nope. Otherwise a user should add the base url in that list.

potter-potter (approved):
looks good

PR #355, opened 2025-01-28
feat/add code to weaviate uploader to create default collection

rbiseck3:
### Description
Start adding support for uploaders to create the destination index/collection/etc. on behalf of a user with default configs.


unstructured_ingest/v2/interfaces/uploader.py
```diff
@@ -51,3 +56,11 @@ def run_data(self, data: list[dict], file_data: FileData, **kwargs: Any) -> None
 
     async def run_data_async(self, data: list[dict], file_data: FileData, **kwargs: Any) -> None:
         return self.run_data(data=data, file_data=file_data, **kwargs)
+
+
+@dataclass
+class VectorDBUploader(Uploader, ABC):
+    def create_destination(
+        self, destination_name: str = "elements", vector_length: Optional[int] = None, **kwargs: Any
```
shreyanid, line 64: `dimension` would be preferred, but `vector_length` is fine

unstructured_ingest/v2/processes/connectors/assets/weaviate_collection_config.json
```diff
@@ -0,0 +1,76 @@
...
+        "enabled": false
+    },
+    "properties": [
+        {
+            "dataType": [
+                "text"
+            ],
+            "indexFilterable": true,
+            "indexSearchable": true,
+            "name": "record_id",
```
shreyanid, line 24: Can we simplify this collection config to just the record_id property? since the other minimum items will be populated by the uploader. Something more in line with https://docs.unstructured.io/platform/destinations/weaviate#weaviate to keep it minimal

unstructured_ingest/v2/processes/embedder.py
```diff
@@ -184,6 +184,9 @@ def get_embedder(self) -> "BaseEmbeddingEncoder":
 class Embedder(BaseProcess, ABC):
     config: EmbedderConfig
 
+    def init(self, *kwargs: Any) -> None:
```
shreyanid, line 187: does this impact the current functionality? I thought the embedder was not invoked/involved here

shreyanid (commented):
looks good, the test is also hanging for me too though so will wait to approve. couple small comments/questions

unstructured_ingest/v2/interfaces/uploader.py
vangheem, line 64: `vector_length` should that be `dimension` instead?

unstructured_ingest/v2/processes/connectors/weaviate/weaviate.py
```diff
@@ -205,17 +208,54 @@ def get_batch_client(self, client: "WeaviateClient") -> Generator["BatchClient",
 
 
 @dataclass
-class WeaviateUploader(Uploader, ABC):
+class WeaviateUploader(VectorDBUploader, ABC):
     upload_config: WeaviateUploaderConfig
     connection_config: WeaviateConnectionConfig
 
+    def _existing_collections(self) -> list[str]:
```
vangheem, line 215: I'm assuming there is no api to just get a single collection to see if it exists?

unstructured_ingest/v2/interfaces/uploader.py
vangheem: Why not change if it is preferred?

rbiseck3: I chose `vector_length` only to make it more aligned with a vector database, since usually in their docs you're setting vector configs. But we can change it to `embedding_dimensions` if we really want to. This makes it more aligned with our process since the assumption is that the vector being written is the embeddings.

unstructured_ingest/v2/processes/embedder.py
rbiseck3: This was something that was already being done but I had a `#TODO` to move it to a dedicated `init` method for a long time, it was just never added until now: [pipeline.py#L94](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/pipeline/pipeline.py#L94)

unstructured_ingest/v2/processes/connectors/assets/weaviate_collection_config.json
rbiseck3: I figured only content with text would be written to a vector db, since the embedding is based off of the text, so it might make sense that the text is also indexed.

unstructured_ingest/v2/processes/connectors/weaviate/weaviate.py
```diff
@@ -205,17 +208,50 @@ def get_batch_client(self, client: "WeaviateClient") -> Generator["BatchClient",
...
+        collection_name = self.upload_config.collection or destination_name
+        self.upload_config.collection = collection_name
+        connectors_dir = Path(__file__).parents[1]
+        collection_config_file = connectors_dir / "assets" / "weaviate_collection_config.json"
+        with collection_config_file.open() as f:
+            collection_config = json.load(f)
+        collection_config["class"] = collection_name
+        if not self._collection_exists():
+            logger.info(
+                f"creating default weaviate collection '{collection_name}' with default configs"
```
shreyanid, line 247: we don't want to create a user's collection, only our "destination_name". we could check for `if collection_name == destination_name`

shreyanid (approved):

PR #357, opened 2025-01-28
Proposal for Neo4j Uploader Improvments

jexp:
- use neo4j-rust-ext instead of plain neo4j driver for 10x perf improvement
- always match on a single node label (equivalent to the constraint), never blank matches
- group relationships by type, source- and target-type
- increase batch size
- use vector property procedure to set fp32 instead of p64
- method to select the main label for a node
- TODO: create vector index would need information from the embedder (dimension) and similarity function (from config)


unstructured_ingest/v2/processes/connectors/neo4j.py
```diff
@@ -257,6 +261,8 @@ async def run_async(self, path: Path, file_data: FileData, **kwargs) -> None:  #
         graph_data = _GraphData.model_validate(staged_data)
         async with self.connection_config.get_client() as client:
             await self._create_uniqueness_constraints(client)
+            # TODO need chunker config
```
rbiseck3, line 264: We could get the dimensions from the content by inspecting the length of the embeddings.

unstructured_ingest/v2/processes/connectors/neo4j.py
```diff
@@ -274,48 +280,73 @@ async def _create_uniqueness_constraints(self, client: AsyncDriver) -> None:
                 """
             )
 
+    async def _create_vector_index(self, client: AsyncDriver, dimensions : int, similarity_function: str) -> None:
+        label = Label.CHUNK
+        logger.info(
+            f"Adding id uniqueness constraint for nodes labeled '{label.value}'"
+            " if it does not already exist."
+        )
```
ds-filipknefel, line 288: This log doesn't seem to describe what's happening, we create a vector index but log we're creating id uniqueness constraint?

unstructured_ingest/v2/processes/connectors/neo4j.py
```diff
@@ -274,48 +280,73 @@ async def _create_uniqueness_constraints(self, client: AsyncDriver) -> None:
...
         )
 
+    def _main_label(self, labels: list[Label]) -> Label:
+        if labels is None or len(labels) == 0: return None
+
+        for label in Label:
+            if label in labels:
+                return label
+            else:
+                return labels[0]
```
ds-filipknefel, line 320: I believe this function may not do what it was intended to, right now it:
- Returns first enum value of `Label` (where order is arbitrary) if that label is present in provided `labels` parameter
- Returns first label from provided `labels` parameter otherwise

1. What was the intention of the function?
2. If the idea is that each node should **necessarily** have a main label that it will be searched by then I'd argue that we should represent that in the abstraction, that is to:
- add a `main_label: Label` field to the node - use that for searches
- have `secondary_labels: list[Label]` field for the rest of the labels

_Names subject to change, just sketching the idea_

unstructured_ingest/v2/processes/connectors/neo4j.py
```diff
@@ -328,45 +359,47 @@ async def _execute_queries(
...
             UNWIND $nodes AS node
-            MERGE (n: {labels_string} {{id: node.id}})
+            MERGE (n: `{label.value}` {{id: node.id}})
             SET n += node.properties
+            SET n:$(node.labels)
+            WITH * WHERE node.vector IS NOT NULL
+            CALL db.create.setNodeVectorProperty(n, 'embedding', node.vector)
             """
-        parameters = {"nodes": [{"id": node.id_, "properties": node.properties} for node in nodes]}
+        parameters = {"nodes": [{"id": node.id_, "labels":[l.value for l in node.labels if l != label],"vector":node.properties.pop('embedding', None), "properties": node.properties} for node in nodes]}
```
ds-filipknefel, line 392: Do we want to allow exactly one label per Node?

unstructured_ingest/v2/processes/connectors/neo4j.py
```diff
@@ -328,45 +359,47 @@ async def _execute_queries(
...
+    def _create_edges_query(edges: list[_Edge], relationship: tuple[Relationship,Label,Label]) -> tuple[str, dict]:
         logger.info(f"Preparing MERGE query for {len(edges)} {relationship} relationships.")
         query_string = f"""
             UNWIND $edges AS edge
-            MATCH (u {{id: edge.source}})
-            MATCH (v {{id: edge.destination}})
-            MERGE (u)-[:{relationship.value}]->(v)
+            MATCH (u: `{relationship[1].value}` {{id: edge.source}})
+            MATCH (v: `{relationship[2].value}` {{id: edge.destination}})
+            MERGE (u)-[:`{relationship[0].value}`]->(v)
```
ds-filipknefel, line 402: Does this imply that a relationship should always be defined in terms of connecting two specified labels?

unstructured_ingest/v2/processes/connectors/neo4j.py
```diff
@@ -274,48 +280,73 @@ async def _create_uniqueness_constraints(self, client: AsyncDriver) -> None:
...
+        logger.info(
+            f"Adding id uniqueness constraint for nodes labeled '{label.value}'"
+            " if it does not already exist."
+        )
+        index_name = f"{label.value.lower()}_vector"
+        await client.execute_query(
+            f"""
+            CREATE VECTOR INDEX {index_name} IF NOT EXISTS
+            FOR (n:{label.value}) ON n.embedding
+            OPTIONS {{`vector.similarity_function`: '{similarity_function}', `vector.dimensions`: {dimensions}}}
```
ds-filipknefel, line 294: If we want the function to be configurable by user when creating the destination connector then such field should be added to `Neo4jUploaderConfig` class

ds-filipknefel (commented):
Few question and comments to inform further review

unstructured_ingest/v2/processes/connectors/neo4j.py
jalakoo: Agreed, this should read, "Adding vector index for nodes labeled '{label.value}'" instead.

unstructured_ingest/v2/processes/connectors/neo4j.py
jalakoo: I believe this is setup to try and avoid any naming collisions when additional indexes are added. Making the index name configurable might have unexpected consequences later.

unstructured_ingest/v2/processes/connectors/neo4j.py
jalakoo: I like this idea

unstructured_ingest/v2/processes/connectors/neo4j.py
jalakoo: Certainly would be simpler. But would dramatically alter the data model for graphs containing multi-labeled nodes, resulting in a lot of duplicated nodes (with the same properties) and relationships to get the same effect.  My gut says no, but it's an idea worth exploring if the ingested data is coming entirely from unstructured.

unstructured_ingest/v2/processes/connectors/neo4j.py
jalakoo: A Match could be found with just the ids, but by adding labels, the match part the of the query will be more performant.

ahmetmeleq (approved):

PR #361, opened 2025-01-31
feat: add destination creation support to pinecone

ahmetmeleq:


unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -159,18 +159,64 @@ def conform_dict(self, element_dict: dict, file_data: FileData) -> dict:
...
+
+            logger.info(f"creating pinecone index {index_name}")
+
+            pc = self.connection_config.get_client()
+
+            pc.create_index(
+                name=destination_name,
+                dimension=vector_length,
+                metric=metric,
+                spec=ServerlessSpec(cloud=cloud, region=region),
```
rbiseck3, line 212: This will need to be either serverless or pod spec based on the type on deployment this is pointing to. I think we can get this by fetching some metadata first.

unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -159,18 +159,64 @@ def conform_dict(self, element_dict: dict, file_data: FileData) -> dict:
...
             raise DestinationConnectionError(f"failed to validate connection: {e}")
 
+    def create_destination(
+        self,
+        vector_length: int,
+        # TODO: make name required to either be in config, or here, then remove default
+        destination_name: Optional[str] = "unstructured-pinecone-default-index",
+        cloud: Optional[str] = "aws",
+        region: Optional[str] = "us-east-1",
+        metric: Optional[str] = "cosine",
```
rbiseck3, line 194: I think these would be required if creating a new destination. Having the default will make sure of that. 
```suggestion
        cloud: str = "aws",
        region: str = "us-east-1",
        metric: str = "cosine",
```

unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -159,18 +159,96 @@ def conform_dict(self, element_dict: dict, file_data: FileData) -> dict:
...
+                    f"index {self.connection_config.index_name} does not exist"
+                )
         except Exception as e:
             logger.error(f"failed to validate connection: {e}", exc_info=True)
             raise DestinationConnectionError(f"failed to validate connection: {e}")
 
+    def create_destination(
+        self,
+        destination_name: str,
+        vector_length: int,
```
shreyanid, line 205: can we have a default vector length, since a dimension is required for creating an index in pinecone? I believe 3072 is the default easy button embedder dimension

unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -160,18 +160,94 @@ def conform_dict(self, element_dict: dict, file_data: FileData) -> dict:
...
+                    f"index {self.connection_config.index_name} does not exist"
+                )
         except Exception as e:
             logger.error(f"failed to validate connection: {e}", exc_info=True)
             raise DestinationConnectionError(f"failed to validate connection: {e}")
 
+    def create_destination(
+        self,
+        destination_name: str,
+        vector_length: int = 3072,
```
rbiseck3, line 204: How did we decide on this default value?

shreyanid: it's the default value of the embedding model chosen through the easy buttons ([docs](https://docs.unstructured.io/platform/workflows#basic)), openAI's `text-embedding-3-large`

rbiseck3: To me, I feel like this shouldn't be a default in open source then since that knowledge isn't relevant to this repo. Personally, I think this should be left without a default. But that would also break the init behavior since nothing would be passing in the embedder dimensions. It might also make sense to do this lazily and create it after reading in a single record and inspecting the length of the embeddings. That would make sure it's the most accurate for open source.

unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -160,18 +160,94 @@ def conform_dict(self, element_dict: dict, file_data: FileData) -> dict:
 
 
 @dataclass
-class PineconeUploader(Uploader):
+class PineconeUploader(VectorDBUploader):
     upload_config: PineconeUploaderConfig
     connection_config: PineconeConnectionConfig
     connector_type: str = CONNECTOR_TYPE
 
+    # TODO: question this
```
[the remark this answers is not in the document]
ahmetmeleq: Why does open source have init behavior on creating an index? Shouldn't this behavior be implemented downstream?

Is the reason that we want to keep downstream agnostic of custom logic? In that case I understand and agree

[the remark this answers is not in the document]
ahmetmeleq: But if we're gonna lazily and set the embedding dimensionality after reading in a single record and inspecting the length of the embeddings, then I think we should also put the destination creation logic to the same place instead of init here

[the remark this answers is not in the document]
ahmetmeleq: So I think we should put both assigning embedding dimensionality, and, destination creation, [in the run_data method](https://github.com/Unstructured-IO/unstructured-ingest/blob/364fc9bea9ea34ff729293f325416b1182ff675d/unstructured_ingest/v2/processes/connectors/pinecone.py#L261-L276). It'd check data for embedding dims, set the embedding dimensionality, and create a destination if needed

unstructured_ingest/v2/processes/connectors/pinecone.py
ahmetmeleq: Cross posting

I actually think we should put both assigning embedding dimensionality, and, destination creation, [in the run_data method](https://github.com/Unstructured-IO/unstructured-ingest/blob/364fc9bea9ea34ff729293f325416b1182ff675d/unstructured_ingest/v2/processes/connectors/pinecone.py#L261-L276). It'd check data for embedding dims, set the embedding dimensionality, and create a destination if needed. 

As opposed to having a default dimensionality assigned, nor, creating a destination at init

test/integration/connectors/test_pinecone.py
```diff
@@ -351,3 +351,36 @@ def test_pinecone_stager(
         stager=stager,
         tmp_dir=tmp_path,
     )
+
+
+@pytest.mark.tags(CONNECTOR_TYPE, DESTINATION_TAG, VECTOR_DB_TAG)
```
shreyanid, line 356: need to add a requires_env decorator

shreyanid (approved):
lgtm! just one note about a test decorator

PR #368, opened 2025-02-03
Fix sharepoint for new Microsoft auth

potter-potter:


unstructured_ingest/v2/pipeline/pipeline.py
```diff
@@ -329,9 +329,9 @@ def from_configs(
         source_entry = {
             k: v
             for k, v in source_registry.items()
```
potter-potter, line 331: @rbiseck3 whats your thoughts here. `isinstance` was pulling in Sharepoint and Onedrive (as subclass)

rbiseck3 (approved):

PR #369, opened 2025-02-03
support custom metadata in sql connector

vangheem:
The SQL connector has been updated to detect which columns the destination table is configured from.

We should just use that to prep to control the data we insert instead of defining a set of columns.


potter-potter (approved):
works well

PR #373, opened 2025-02-04
feat(DS-547) zendesk source connector

nikpocuca:


unstructured_ingest/v2/processes/connectors/zendesk.py
```diff
@@ -0,0 +1,227 @@
...
+            "token": access_config.api_token,
+        }
+
+        client = zenpy.Zenpy(**options)
+        yield client
+
+
+class ZendeskIndexerConfig(IndexerConfig):
+    batch_size: int = Field(
+        default="1",
```
ds-filipknefel, line 68:
```suggestion
        default=1,
```

unstructured_ingest/v2/processes/connectors/zendesk.py
```diff
@@ -0,0 +1,227 @@
...
+class ZendeskIndexer(Indexer):
+    connection_config: ZendeskConnectionConfig
+    index_config: ZendeskIndexerConfig
+    connector_type: str = CONNECTOR_TYPE
+
+    def precheck(self) -> None:
+        """Validates connection to Zendesk api"""
+        try:
+            # there is no context manager method for Zenpy.
+            with self.connection_config.get_client() as client:
```
ds-filipknefel, line 83: I don't think this comment is useful, even if there was a context manager we'd do it the same way as you do here

unstructured_ingest/v2/processes/connectors/zendesk.py
```diff
@@ -0,0 +1,227 @@
...
+
+                if client.users()[:] == []:
+                    raise SourceConnectionError(
+                        f"users do not exist in subdomain {self.connection_config.sub_domain}"
+                    )
```
ds-filipknefel, line 88: Could we check immediately for some more specific information? e.g. whether the provided email is associated with one of existing users? Or is there no such relationship?

unstructured_ingest/v2/processes/connectors/zendesk.py
```diff
@@ -0,0 +1,227 @@
...
+            tickets = client.tickets()
+            return tickets
+
+    def _list_comments(self, ticket_generator, ticket_id: int):
+        return ticket_generator.comments(ticket=ticket_id)
+
+    def _generate_fullpath(self, ticket) -> Path:
+        return Path(hashlib.sha256(str(ticket.id).encode("utf-8")).hexdigest()[:16] + ".txt")
+
+    # require dependency zenpy
```
ds-filipknefel, line 111: Don't think this comment is very useful, it's the case with all connectors

unstructured_ingest/v2/processes/connectors/zendesk.py
```diff
@@ -0,0 +1,227 @@
...
+    def is_async(self) -> bool:
+        return False
+
+    def _list_tickets(self):
+        with self.connection_config.get_client() as client:
+            tickets = client.tickets()
+            return tickets
+
+    def _list_comments(self, ticket_generator, ticket_id: int):
+        return ticket_generator.comments(ticket=ticket_id)
```
ds-filipknefel, line 106: Can you add return types to both of these?

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
```diff
@@ -0,0 +1,146 @@
...
+@dataclass
+class ZendeskTicket:
+    id: str
+    subject: str
+    description: str
+    generated_ts: int
+    metadata: dict
+
+
+class ZendeskWrapper:
```
rbiseck3, line 25:
```suggestion
class ZendeskClient:
```

rbiseck3:
Why didn't we use any of the supported python SDKs for zendesk rather than writing our own wrapper? [Python](https://developer.zendesk.com/documentation/ticketing/api-clients/python/)

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
```diff
@@ -0,0 +1,146 @@
...
+
+    z = ZendeskWrapper(token=API_TOKEN, subdomain=ENDPOINT, email="test@unstructured.io")
+
+    tickets = z.get_tickets()
+
+    first_ticket = tickets[0]
+
+    comments = z.get_comments(ticket_id=first_ticket.id)
+
+    breakpoint()
```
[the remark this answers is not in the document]
nikpocuca: woops i left this thing in there

nikpocuca:
@rbiseck3 
> Why didn't we use any of the supported python SDKs for zendesk rather than writing our own wrapper? [Python](https://developer.zendesk.com/documentation/ticketing/api-clients/python/)

There was a conversation about this on our channel, basically of the three python libraries, 

Zenpy - is under copy-left license so we cannot use it according to our policies. 
Zdesk - has not been updated in several years. 
Eventbrite - is not directly supported. 

Before I had an implementation using Zenpy, and I had to quickly write my own wrapper.

unstructured_ingest/v2/processes/connectors/zendesk.py
nikpocuca: The only guaranteed outcome that you have with zendesk is that at least 1 user exists. 

The other requires that emails provided may be removed etc, for a pre-check i think this is sufficient. 

The token alone supports authentication, my only interest with this test is to make sure that there is data coming from zendesk.

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
```diff
@@ -0,0 +1,146 @@
+from dataclasses import dataclass
```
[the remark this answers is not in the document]
nikpocuca: I would agree with this statement but in my opinion, 

 client and connector should be in seperate files. There are a few features for zendesk that need to be added including:
 
-  parsing of attachments to tickets,
-  batching 
-  other useful metadata. 

In addition, we have to maintain both the connector and the client overtime if Zendesk decides to change its api it should be treated as an independent feature. 
- wrapper/client 
- connector (name.py)

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,219 @@
...
+    def is_async(self) -> bool:
+        return False
+
+    def _list_tickets(self) -> List[ZendeskTicket]:
+        with self.connection_config.get_client() as client:
+            tickets = client.get_tickets()
+            return tickets
+
+    def _list_comments(self, ticket_generator, ticket_id: int) -> List[Comment]:
+        return ticket_generator.get_comments(ticket_id)
```
ds-filipknefel, line 103: This is unused right?

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+class ZendeskAccessConfig(AccessConfig):
+    api_token: str = Field(
+        description="API token for zendesk generated under Apps and Integrations"
+    )
+
+
+class ZendeskConnectionConfig(ConnectionConfig):
+    sub_domain: str = Field(description="Subdomain for zendesk site, <sub-domain>.company.com")
```
ds-filipknefel, line 43: I'd rename to `subdomain`, it's a single word and in few other places spelled this way

nikpocuca: done.

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
```diff
@@ -0,0 +1,124 @@
...
+                    author_id=entry["author_id"],
+                    body=entry["body"],
+                    metadata=entry,
+                    parent_ticket_id=ticket_id,
+                )
+
+                comments.append(comment)
+        else:
+            raise RuntimeError(
+                f"Comments for ticket id:{ticket_id} could not be acquried from url: {comments_url}"
```
rbiseck3, line 68:
```suggestion
                f"Comments for ticket id:{ticket_id} could not be acquired from url: {comments_url}"
```

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
```diff
@@ -0,0 +1,124 @@
...
+
+            # TODO(convert this into a dataclass later, right now just set it as a list of dicts)
+            users = users_in_response
+
+        else:
+            raise RuntimeError(f"Users could not be acquried from url: {users_url}")
+
+        return users
+
+    def get_tickets(self, ticket_id=None) -> List[ZendeskTicket]:
```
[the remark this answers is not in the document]
rbiseck3: Actually are we not doing anything is a ticket id is passed in? Why support that as a input then?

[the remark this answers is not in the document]
nikpocuca: i was going to update it later

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,219 @@
...
+                    f.write("comment")
+                    f.write("\n")
+                    f.write(str(comment["comment_id"]))
+                    f.write("\n")
+                    f.write(str(comment["author_id"]))
+                    f.write("\n")
+                    f.write(comment["body"])
+                    f.write("\n")
+                    f.write(comment["date_created"])
+                    f.write("\n")
```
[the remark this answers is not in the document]
nikpocuca: updated to add ticket stuff.

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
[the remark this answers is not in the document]
nikpocuca: removed ticket id.

unstructured_ingest/v2/processes/connectors/zendesk/__init__.py
```diff
@@ -3,7 +3,28 @@
...
 
 from .zendesk import (
+    ZendeskAccessConfig,
+    ZendeskClient,
+    ZendeskConnectionConfig,
+    ZendeskDownloader,
+    ZendeskDownloaderConfig,
+    ZendeskIndexer,
+    ZendeskIndexerConfig,
+    ZendeskTicket,
```
ds-filipknefel, line 13: I am suspicious of why are these imports necessary.

nikpocuca: you want to make the visible to zendesk module etc.

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
```diff
@@ -0,0 +1,339 @@
...
+
+        for entry in tickets_in_response:
+            ticket = ZendeskTicket(
+                id=entry["id"],
+                subject=entry["subject"],
+                description=entry["description"],
+                generated_ts=entry["generated_timestamp"],
+                metadata=entry,
+            )
+            tickets.append(ticket)
```
mpolomdeepsense, line 217: Why not use list comprehension? (multiple places where this could be applied)

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
```diff
@@ -0,0 +1,339 @@
...
+                    # Encode the file content as base64 bytes
+                    encoded_content = base64.b64encode(download_response.content).decode("utf-8")
+                    print(f"Encoded attachment {attachment['file_name']} successfully.")
+
+                    # You can store the encoded content as part of the attachment data
+                    encoded_string_injection = f"data:{attachment_data["content_type"]};base64,{encoded_content}"
+                    attachment_data["encoded_content"] = encoded_string_injection
+
+                else:
+                    print(f"Failed to download attachment {attachment['file_name']}.")
```
mpolomdeepsense, line 290: Replace print with logger (multiple places)

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,494 @@
...
+            f.write(file_data.identifier + "\n")
+            f.write(file_data.metadata.record_locator["subject"] + "\n")
+            f.write(file_data.metadata.record_locator["description"] + "\n")
+            f.write(first_date + "\n")
+            for comment in comments:
+                f.write("comment\n")
+                f.write(str(comment["comment_id"]) + "\n")
+                f.write(str(comment["author_id"]) + "\n")
+                f.write(comment["body"] + "\n")
+                f.write(comment["date_created"] + "\n")
```
mpolomdeepsense, line 349: wouldn't it be better to create the whole string first and just write it once?

unstructured_ingest/v2/processes/connectors/zendesk/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+    "ZendeskClient",
+    "ZendeskConnectionConfig",
+    "ZendeskDownloader",
+    "ZendeskDownloaderConfig",
+    "ZendeskIndexer",
+    "ZendeskIndexerConfig",
+    "ZendeskTicket",
+]
+
+add_source_entry(source_type="zendesk", entry=zendesk_source_entry)
```
mpolomdeepsense, line 30: Would be better to import `CONNECTOR_TYPE` from zendesk connector file and use it for `source_type` instead.

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
nikpocuca: I would normally agree with you but im not a fan of list comprehension in this case because its harder to debug  

```
tickets = [
    ZendeskTicket(
        id=entry["id"],
        subject=entry["subject"],
        description=entry["description"],
        generated_ts=entry["generated_timestamp"],
        metadata=entry,
    )
    for entry in tickets_in_response
]
```

Suppose one of these tickets was of interest from a debugging perspective, how would you do list apprehension + debug?

mpolomdeepsense: I would set the breakpoint after the list and look through it in order to find the object i'm interested in. I see your point since now you can set breakpoint inside the for loop, but in this case it changes very little because you still have to go through the whole loop in order to find the item you are interested in. IMO it doesn't change much for this simple case.

The main reason why I brought this up is because list comprehension were design for such simple cases like a single append of an object, and also in general they are faster than for loops.

test/integration/connectors/test_zendesk.py
```diff
@@ -0,0 +1,72 @@
+import os
+from pathlib import Path
+from typing import Optional
+
+import pytest
+
```
potter-potter, line 6: I think we need 
from test.integration.connectors.utils.constants import SOURCE_TAG, UNCATEGORIZED_TAG
and the appropriate tags on the actual tests so that it gets picked up by CI

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
```diff
@@ -0,0 +1,339 @@
...
+                attachment_url = attachment['content_url']
+                download_response = requests.get(attachment_url, auth=self._auth)
+
+                if download_response.status_code == 200:
+                    # Encode the file content as base64 bytes
+                    encoded_content = base64.b64encode(download_response.content).decode("utf-8")
+                    logger.debug(f"Encoded attachment {attachment['file_name']} successfully.")
+
+                    # You can store the encoded content as part of the attachment data
+                    encoded_string_injection = f"data:{attachment_data["content_type"]};base64,{encoded_content}"
```
potter-potter, line 286: This fstring wont work. content_type needs single quotes

test/integration/connectors/test_zendesk.py
```diff
@@ -0,0 +1,72 @@
...
+    )
+
+
+@pytest.mark.asyncio
+@requires_env("ZENDESK_ENDPOINT", "ZENDESK_TOKEN")
+async def test_zendesk_source(temp_dir):
+    await zendesk_source_test(
+        tmp_path=temp_dir,
+        token=os.environ["ZENDESK_TOKEN"],
+        endpoint=os.environ["ZENDESK_ENDPOINT"],
```
potter-potter, line 69: I'm not understanding where this being used. I don't see an `endpoint` in the actual ZendeskConnectionConfig in the main file.

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
nikpocuca: it has to be there because its their own API, if it isnt there, it means their api changed.

test/integration/connectors/test_zendesk.py
nikpocuca: agreed, this was an artifact from a previous rendition.

test/integration/connectors/test_zendesk.py
nikpocuca: image
tagged and bagged.

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
```diff
@@ -0,0 +1,350 @@
+import base64
```
rbiseck3, line 1: Can we call this file `client.py` to match the code?

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
```diff
@@ -0,0 +1,350 @@
...
+                subject=entry["subject"],
+                description=entry["description"],
+                generated_ts=entry["generated_timestamp"],
+                metadata=entry,
+            )
+            tickets.append(ticket)
+
+        return tickets
+
+    def get_tickets(self) -> List[ZendeskTicket]:
```
rbiseck3, line 221: Do we need async and non async implementations of each method?

unstructured_ingest/v2/processes/connectors/zendesk/wrapper.py
```diff
@@ -0,0 +1,350 @@
...
+                subject=entry["subject"],
+                description=entry["description"],
+                generated_ts=entry["generated_timestamp"],
+                metadata=entry,
+            )
+            tickets.append(ticket)
+
+        return tickets
+
+    def get_article_attachments(self, article_id: str):
```
rbiseck3, line 247: Why is this one not async?

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,641 @@
...
+        description="Type of item from zendesk to parse, can only be `tickets` or `articles`.",
+    )
+
+
+@dataclass
+class ZendeskIndexer(Indexer):
+    connection_config: ZendeskConnectionConfig
+    index_config: ZendeskIndexerConfig
+    connector_type: str = CONNECTOR_TYPE
+
```
rbiseck3, line 128: Will need to add this:
```python
    def is_async(self) -> bool:
        return True
```

Otherwise it defaults to using the non async code.

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,641 @@
...
+                        f"{comment.get('date_created', '')}\n"
+                    )
+
+                await f.write(content)
+
+        return super().generate_download_response(
+            file_data=batch_file_data, download_path=download_path
+        )
+
+    @SourceConnectionError.wrap
```
rbiseck3, line 604: We're moving away from using `SourceConnectionError`, instead can we use the errors here: [unstructured_ingest/v2/errors.py](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/errors.py)

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,641 @@
...
+        # Process each ticket in the batch
+        async with aiofiles.open(download_path, "a", encoding="utf8") as f:
+            for batch_item in batch_file_data.batch_items:
+                ticket_identifier = batch_item.identifier
+                first_date = None
+                comments: List[dict] = []
+
+                # Fetch comments asynchronously
+                comments_list = await client.get_comments_async(
+                    ticket_id=ticket_identifier
```
rbiseck3, line 557: There's a typing mismatch, `batch_item.identifier` is a `str` being passed in for `ticket_id` which is typed as an `int`

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,641 @@
...
+            "email": self.email,
+            "subdomain": self.subdomain,
+            "token": access_config.api_token,
+        }
+
+        client = ZendeskClient(**options)
+        yield client
+
+    @asynccontextmanager
+    async def get_client_async(self) -> AsyncGenerator["ZendeskClient", None]:
```
rbiseck3, line 98: We don't need to wrap this in a context manager is there's no closing logic needed.

rbiseck3: This is no different than: 
```python
    async def get_client_async(self) -> "ZendeskClient":
        """Provides an async context manager for ZendeskClient."""
        access_config = self.access_config.get_secret_value()

        options = {
            "email": self.email,
            "subdomain": self.subdomain,
            "token": access_config.api_token,
        }

        client = ZendeskClient(**options)
        return client
```

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,641 @@
...
+        client = ZendeskClient(**options)
+        yield client
+
+
+class ZendeskIndexerConfig(IndexerConfig):
+    batch_size: int = Field(
+        default=1,
+        description="Number of tickets or articles.",
+    )
+    item_type: str = Field(
```
rbiseck3, line 117: If we know it can only be a subset of strings, we can user `Literal` to add that for type checking. 
```suggestion
    item_type: Literal["tickets", "articles"] = Field(
```

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,641 @@
...
+    def precheck(self) -> None:
+        """Validates connection to Zendesk API."""
+        try:
+            with self.connection_config.get_client() as client:
+                if not client.get_users():
+                    subdomain_endpoint = f"{self.connection_config.subdomain}.zendesk.com"
+                    raise SourceConnectionError(
+                        f"Users do not exist in subdomain {subdomain_endpoint}"
+                    )
+        except SourceConnectionError as e:
```
rbiseck3, line 138: Same comment I posted elsewhere, let's not use these errors but the ones here instead: [unstructured_ingest/v2/errors.py](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/errors.py)

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,641 @@
...
+            logger.error(f"Source connection error: {e}", exc_info=True)
+            raise
+        except Exception as e:
+            logger.error(f"Failed to validate connection to Zendesk: {e}", exc_info=True)
+            raise SourceConnectionError(f"Failed to validate connection: {e}")
+
+    def is_async(self) -> bool:
```
rbiseck3, line 145: Usually the indexer doesn't need to be `async` if it's just making a single call/request, since the indexer runs once per pipeline. No harm in making it `async`, but not as much optimization as the downloader.

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,641 @@
...
+
+            full_path = self._generate_fullpath(str(article_batch[0].id))
+            source_identifiers = SourceIdentifiers(filename=full_path.name, fullpath=str(full_path))
+
+            batched_file_data = ZendeskBatchFileData(
+                connector_type=self.connector_type,
+                metadata=metadata,
+                batch_items=batch_items,
+                additional_metadata=additional_metadata,
+                source_identifiers=source_identifiers,
```
rbiseck3, line 205: Missing `identifier` which is required in all `FileData`

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,641 @@
...
+            yield batched_file_data
+
+    async def handle_articles_async(
+        self, articles: List[ZendeskArticle], batch_size: int
+    ) -> AsyncGenerator[ZendeskBatchFileData, None]:
+        """Parses articles from a list and yields FileData objects asynchronously in batches."""
+        for article_batch in batch_generator(articles, batch_size=batch_size):
+            additional_metadata = ZendeskAdditionalMetadata(
+                item_type="articles",
+                leading_id=str(article_batch[0].id),
```
rbiseck3, line 217: I think these would need to be sorted first to use this logic for leading and tail ids.

rbiseck3: Sorting should probably be done before the call to `batch_generator`

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
rbiseck3: Also do we not want to support pulling in all this data in a single call? Otherwise they need to run 2 different pipelines if they want to ingest tickets and articles.

unstructured_ingest/v2/processes/connectors/zendesk/client.py
```diff
@@ -0,0 +1,369 @@
...
+            response = requests.get(url_to_check, auth=auth)
+
+            http_code = response.status_code
+
+            if 400 <= http_code < 500:
+                message = (
+                    f"Failed to connect to {url_to_check} using zendesk response"
+                    f"status code {http_code}"
+                )
+                return UserError(message)
```
rbiseck3, line 60:
```suggestion
                raise UserError(message)
```

unstructured_ingest/v2/processes/connectors/zendesk/client.py
```diff
@@ -0,0 +1,369 @@
...
+
+            if http_code >= 500:
+                return ProviderError(f"Failed to connect to {url_to_check} using zendesk response")
```
rbiseck3, line 63:
```suggestion
                raise ProviderError(f"Failed to connect to {url_to_check} using zendesk response")
```

unstructured_ingest/v2/processes/connectors/zendesk/client.py
```diff
@@ -0,0 +1,369 @@
...
+
+            if http_code != 200:
+                raise Exception(f"Failed to connect to {url_to_check} using zendesk response")
+
+        except Exception as e:
+            raise RuntimeError(f"Failed to instantiate response: {e}") from e
```
rbiseck3, line 69: This is going to catch everything you raise above and wrap it in this generic `RuntimeError` given the way the code is written right now.

unstructured_ingest/v2/processes/connectors/zendesk/client.py
```diff
@@ -0,0 +1,369 @@
...
+                message = (
+                    f"Failed to connect to {url_to_check} using zendesk response"
+                    f"status code {http_code}"
+                )
+                return UserError(message)
+
+            if http_code >= 500:
+                return ProviderError(f"Failed to connect to {url_to_check} using zendesk response")
+
+            if http_code != 200:
```
rbiseck3, line 65: Let the request library raise the error if it's one we didn't explicitly catch by checking the status codes. 
```suggestion
            response.raise_for_status()
```

unstructured_ingest/v2/processes/connectors/zendesk/client.py
```diff
@@ -0,0 +1,369 @@
...
+        retrieves article content from zendesk using requests
+        """
+
+        articles: List[ZendeskArticle] = []
+
+        article_url = f"https://{self._subdomain}.zendesk.com/api/v2/help_center/articles.json"
+
+        response = requests.get(article_url, auth=self._auth)
+
+        if response.status_code == 200:
```
rbiseck3, line 118: Should wrap the response in the same error handling you have in the init method. Let's create a method similar to the one used elsewhere (wrap_error) on this client and leverage it whenever we make a request.

unstructured_ingest/v2/processes/connectors/zendesk/client.py
```diff
@@ -0,0 +1,369 @@
...
+                for entry in articles_in_response
+            ]
+            return articles
+
+        raise RuntimeError(
+            f"Articles were not able to be acquired from url: {article_url}."
+            f"Status Code: {response.status_code}"
+        )
+
+    def get_articles(self) -> List[ZendeskArticle]:
```
rbiseck3, line 107: I don't see this non async method being used anywhere.

unstructured_ingest/v2/processes/connectors/zendesk/client.py
```diff
@@ -0,0 +1,369 @@
...
+        if response.status_code == 200:
+            users_in_response: List[dict] = response.json()["users"]
+            users = users_in_response
+
+        else:
+            raise RuntimeError(f"Users could not be acquried from url: {users_url}")
+
+        return users
+
+    async def get_users_async(self) -> List[dict]:
```
rbiseck3, line 201: I don't see this method being used anywhere

unstructured_ingest/v2/processes/connectors/zendesk/client.py
```diff
@@ -0,0 +1,369 @@
...
+                subject=entry["subject"],
+                description=entry["description"],
+                generated_ts=entry["generated_timestamp"],
+                metadata=entry,
+            )
+            tickets.append(ticket)
+
+        return tickets
+
+    def get_article_attachments(self, article_id: str):
```
rbiseck3, line 266: I don't see this non-async method being used anywhere.

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -0,0 +1,402 @@
...
+    def _generate_fullpath(self, identifier: str) -> Path:
+        return Path(hashlib.sha256(identifier.encode("utf-8")).hexdigest()[:16] + ".txt")
+
+    async def handle_articles_async(
+        self, articles: List[ZendeskArticle], batch_size: int
+    ) -> AsyncGenerator[ZendeskBatchFileData, None]:
+        """Parses articles from a list and yields FileData objects asynchronously in batches."""
+        for article_batch in batch_generator(articles, batch_size=batch_size):
+
+            article_batch = sorted(article_batch)
```
rbiseck3, line 150: I don't think this is going to sort it the way we think since `articles_batch` isn't a list of concrete types, such as ints. We'll need to tell `sorted` how to sort it. In this case, sort it by `id`. And this is assuming `id` is a sortable list. I'm not sure what form the id is coming in but we need it sortable since we're leveraging it for pagination given the `leading_id` and `tail_id` fields we use when fetching the content.

nikpocuca: I have added logic on how these classes/objects are sorted within the method. 

```python

@dataclass
class ZendeskTicket:
    id: str
    subject: str
    description: str
    generated_ts: int
    metadata: dict

    def __lt__(self, other):
        return self.id < other.id


@dataclass
class ZendeskArticle:
    id: str
    author_id: str
    title: str
    content: str

    def __lt__(self, other):
        return self.id < other.id

```

nikpocuca: I can add int(self.id) < int(other.id) ?

rbiseck3: But is `id` an int? If so, the `id` field itself should be typed as an `int` rather than us casting it in the comparator. Can you post an example id you're getting back from their api?

nikpocuca: (Pdb) type(tickets_in_response[0]['id'])
<class 'int'>

looks like an int

rbiseck3: great, then lets update the typing for that field in the dataclass.

unstructured_ingest/v2/processes/connectors/zendesk/client.py
nikpocuca: new method defined and wrapped.

unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
nikpocuca: updated

unstructured_ingest/v2/processes/connectors/zendesk/client.py
nikpocuca: fixed it with error handling

unstructured_ingest/v2/processes/connectors/zendesk/client.py
```diff
@@ -0,0 +1,209 @@
...
+
+        response = httpx.get(url_to_check, auth=auth)
+
+        self.handle_response(response, url_to_check)
+        self._token = token
+        self._subdomain = subdomain
+        self._email = email
+        self._auth = auth
+
+    def handle_response(self, response: httpx.Response, url: Optional[str] = None):
```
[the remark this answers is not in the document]
rbiseck3: You'll need to update the code elsewhere to raise the error explicitly and use this to wrap any exceptions that are caught

rbiseck3 (approved):

PR #374, opened 2025-02-05
feat/warning on file unsupported

rbiseck3:
### Description
By default, only raise a warning when a file that can't be processed throws a file unsuppported error.


potter-potter (approved):
looks good

PR #376, opened 2025-02-07
feat: Improved google drive precheck

mateuszkuprowski:
Precheck can now actually verify much more things hence it would be more informative for the user what went wrong when trying to use the connector. It will hopefully catch more cases that would fail at precheck than at runtime.


Paul-Cornell:
Any way to check whether the user also has the Google Drive API enabled in their account? At least two customers reported this as an issue that, when they enabled it, caused their workflows to run as expected.

unstructured_ingest/v2/processes/connectors/google_drive.py
```diff
@@ -132,12 +132,104 @@ class GoogleDriveIndexer(Indexer):
...
+
+            # If the target is a folder, perform file count check.
+            if self.is_dir(root_info):
+                if self.index_config.recursive:
+                    file_count = self.count_files_recursively(
+                        client,
+                        self.connection_config.drive_id,
+                        extensions=self.index_config.extensions,
+                    )
+                    if file_count == 0:
```
potter-potter, line 200: Are we sure we want to raise here?  I could see someone setting up the connector. And then filling the folder. Or if it is a scheduled job that runs every hour and the folder gets emptied and refilled every 6 hours.

mateuszkuprowski: Good point, I was thinking more from the perspective that we have when we usually fire one time jobs to process something specific. However this warrants more discussion I think, because me chaning this a warning in ingest will only hide it from the user on the platform as they can only see errors during setup.

mateuszkuprowski:
> Any way to check whether the user also has the Google Drive API enabled in their account? At least two customers reported this as an issue that, when they enabled it, caused their workflows to run as expected.

Absolutely I'll separate a check for that. With my current fixes it would still fail when API was not enabled, but I'll try to make a simple check to produce a separate error specifically when API is not enabled.

unstructured_ingest/v2/processes/connectors/google_drive.py
potter-potter: I passed it by sudarshan. He agrees that it doesn't seem like a reason to fail. But please double check with Product.

I believe most users of platform will be running it as a scheduled service. I don't think its a good idea to fail if there are no documents. The precheck will get run with every pipeline run now I believe.

potter-potter:
This is a great improvement over our previous google drive precheck.

bryan-unstructured (approved):

PR #379, opened 2025-02-10
fix: Prepare for breaking change in upcoming unstructured-client

awalker4:
In the unreleased 0.30.0 of `unstructured-client`, handling of the `server_url` is changing. We need to do this to integrate the platform API functions. Now, different parts of the SDK are talking to different backend services, and so it's preferred that we set the `server_url` per endpoint.

Note that for compatibility with the current version, we need to strip the `/general/v0/general` path off of the URL. We have logic to do this in the SDK, but in versions before 0.30.0 this only applies to the `server_url` passed to the client instance. So, it's a bit of a chicken and egg. Once 0.30.0 is published and pulled in this repo, we can remove these lines if we want.

Related core library fix: https://github.com/Unstructured-IO/unstructured/pull/3916

### DIFF
--- a/unstructured_ingest/v2/unstructured_api.py
+++ b/unstructured_ingest/v2/unstructured_api.py
@@ -87,13 +87,16 @@ async def call_api_async(
     """
     from unstructured_client import UnstructuredClient
 
+    # Note(austin) - the sdk takes the base url, but users may pass the full endpoint
+    # For consistency, strip off the path when it's given
+    base_url = server_url[:-19] if "/general/v0/general" in server_url else server_url
+
     client = UnstructuredClient(
-        server_url=server_url,
         api_key_auth=api_key,
     )
     partition_request = create_partition_request(filename=filename, parameters_dict=api_parameters)
     try:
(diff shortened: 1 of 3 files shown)


unstructured_ingest/v2/unstructured_api.py
```diff
@@ -87,13 +87,16 @@ async def call_api_async(
     """
     from unstructured_client import UnstructuredClient
 
+    # Note(austin) - the sdk takes the base url, but users may pass the full endpoint
+    # For consistency, strip off the path when it's given
+    base_url = server_url[:-19] if "/general/v0/general" in server_url else server_url
```
PastelStorm, line 92: Would be safer to split on `/general/v0/general` in case there are trailing spaces or tabs.

PastelStorm: Also `server_url` is an optional parameter, so it would make sense to check if it exists.

awalker4:
Closing in preference of an upstream fix: https://github.com/Unstructured-IO/unstructured-python-client/pull/226

PR #390, opened 2025-02-19
chore: add destination name format for weaviate

shreyanid:


unstructured_ingest/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "0.5.5"  # pragma: no cover
+__version__ = "0.5.6-dev0"  # pragma: no cover
```
vangheem, line 1: Do you want to just make a release right away? Maybe just do `0.5.6` instead of `-dev0`

vangheem (approved):

PR #391, opened 2025-02-19
feat: opinionated writes for AstraDB destination

shreyanid:


test/integration/connectors/test_astradb.py
```diff
@@ -85,6 +86,7 @@ def test_precheck_fails_indexer(connection_config: AstraDBConnectionConfig):
         indexer.precheck()
 
 
+@pytest.mark.skip("Returning success")
```
vangheem, line 89: what's up here?

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -337,25 +350,79 @@ class AstraDBUploader(Uploader):
...
+
     def precheck(self) -> None:
         try:
-            get_astra_collection(
-                connection_config=self.connection_config,
-                collection_name=self.upload_config.collection_name,
-                keyspace=self.upload_config.keyspace,
-            ).options()
+            if self.upload_config.collection_name:
+                print("im here", self.upload_config.collection_name)
```
vangheem, line 359: think you want to remove this :)

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -337,25 +350,79 @@ class AstraDBUploader(Uploader):
...
+            collection_name=collection_name or self.upload_config.collection_name,
             keyspace=self.upload_config.keyspace,
         )
 
+    def _collection_exists(self, collection_name: str):
+        astra_db = get_astra_db(
+            connection_config=self.connection_config,
+            keyspace=self.upload_config.keyspace,
+        )
+        return collection_name in astra_db.list_collection_names(
```
vangheem, line 387: Maybe just do a GET on the specific collection instead of listing all the collections?

vangheem (approved):
I think you have a few things to clean up bur approving assuming you'll fix them before merge.

unstructured_ingest/v2/processes/connectors/astradb.py
shreyanid: I didn't use just `get_collection` because according to the [docs](https://docs.datastax.com/en/astra-db-serverless/api-reference/collection-methods/get-collection.html), that method will always return a collection object (even if it doesn't exist). I could then try `collection_object.options()` and except the `CollectionNotFoundException` as a way to determine if the collection exists - I've updated the code to do this instead of listing all names.

I do see that `create_collection` has a check_exists flag param too, but we want separate functions to check for the collection and create one.

unstructured_ingest/v2/processes/connectors/astradb.py
```diff
@@ -103,9 +104,27 @@ def get_astra_collection(
...
+) -> "AstraDBCollection":
+    from astrapy.exceptions import CollectionNotFoundException
+
+    astra_db = get_astra_db(connection_config=connection_config, keyspace=keyspace)
 
-    # Connect to the collection
     astra_db_collection = astra_db.get_collection(name=collection_name)
+    # astradb will return a collection object in all cases (even if it doesn't exist)
+    try:
+        astra_db_collection.options()
```
shreyanid, line 122: question for reviewer: should this check go here in this function, or should it just return a (potentially invalid) collection that the caller will then be responsible for verifying?
- leaving it in this get_astra_collection: the collection returned will be guaranteed to exist
- moving it to each caller's logic: only verified if the caller cares / doesn't already have it verified (ex. the indexer already has this logic in its precheck, so its double checked at the moment)

note that this logic gets similarly repeated in the uploader `_collection_exists` method, so if I move the logic out I can use this get_astra_collection method

...in typing this i think i convinced myself to move it out. please hold :)

vangheem (approved):

PR #395, opened 2025-02-20
fix: pinecone index creation

ahmetmeleq:
also: 
- fix potential weaviate name formatting bug
- add identity to default index name for opinionated writes
- add reference test script for validation
- change step name due to potentially confusing log
- change method name due to potentially confusing naming


unstructured_ingest/v2/interfaces/uploader.py
```diff
@@ -38,7 +38,7 @@ def is_batch(self) -> bool:
     def run_batch(self, contents: list[UploadContent], **kwargs: Any) -> None:
         raise NotImplementedError()
 
-    def create_destination(self, destination_name: str = "elements", **kwargs: Any) -> bool:
+    def create_destination(self, destination_name: str = "unstructuredautocreated", **kwargs: Any) -> bool:
```
ahmetmeleq, line 41: @shreyanid thoughts on this? It feels much better to declare identity while creating these

unstructured_ingest/v2/pipeline/steps/index.py
```diff
@@ -11,7 +11,7 @@
 
 IndexerT = TypeVar("IndexerT", bound=Indexer)
 
-STEP_ID = "index"
+STEP_ID = "indexer"
```
ahmetmeleq, line 14: @rbiseck3 thoughts on this? It is annoying but otherwise we get logs that might really confuse people:

`2025-02-21 02:45:32,101 MainProcess INFO     created index with configs: {"input_path": "example-docs/book-war-and-peace-1225p.txt", "recursive": false}, connection configs: {"access_config": "**********"}`

the above says "creating index" but it actually means the pipeline step. but it might be super confusing. this tries to address that

unstructured_ingest/v2/interfaces/uploader.py
shreyanid: fine with me! Length looks ok, and the formatting functions for each connector will take care of the rest

unstructured_ingest/v2/processes/connectors/pinecone.py
```diff
@@ -219,7 +219,7 @@ def create_destination(
     ) -> bool:
         from pinecone import PodSpec, ServerlessSpec
 
-        index_name = destination_name or self.connection_config.index_name
+        index_name = self.connection_config.index_name or destination_name
```
shreyanid, line 222: good catch!

unstructured_ingest/v2/processes/connectors/weaviate/weaviate.py
```diff
@@ -241,10 +241,13 @@ def format_destination_name(self, destination_name: str) -> str:
...
     def create_destination(
-        self, destination_name: str = "elements", vector_length: Optional[int] = None, **kwargs: Any
+        self,
+        destination_name: str = "unstructuredautocreated",
+        vector_length: Optional[int] = None,
+        **kwargs: Any,
     ) -> bool:
-        destination_name = self.format_destination_name(destination_name)
         collection_name = self.upload_config.collection or destination_name
+        collection_name = self.format_destination_name(collection_name)
```
shreyanid, line 250: hm, i had put it first because i didn't want us to be reformatting the user's inputted name, but on second thought since we are writing the user's index / want things to "just work" more often I am good with this change

PR #399, opened 2025-02-25
feat: Jira v1 to v2 migration

mpolomdeepsense:
- Migrates Jira source from v1 to v2
- Adds Jira source integration tests
- Adds Jira source unit tests

## Note

Because of failing google-drive e2e test, jira e2e test is not executed. Here's the result of successful Jira e2e test (when google-drive e2e test is disabled): https://github.com/Unstructured-IO/unstructured-ingest/actions/runs/13518362476/job/37771837198?pr=399#step:6:3892


bryan-unstructured (commented):
initial round of review.
Will take a closer look if PR is not merged yet tomorrow.

unstructured_ingest/v2/processes/connectors/jira.py
```diff
@@ -0,0 +1,457 @@
...
+
+@dataclass
+class JiraIndexer(Indexer):
+    connection_config: JiraConnectionConfig
+    index_config: JiraIndexerConfig
+    connector_type: str = CONNECTOR_TYPE
+
+    def precheck(self) -> None:
+        try:
+            client: "Jira"
```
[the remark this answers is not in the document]
mpolomdeepsense: Only for developing purposes. The linter doesn't properly recognize the type of the client, thus not giving the hints for methods etc. Do you think it's better to remove it?

unstructured_ingest/v2/processes/connectors/jira.py
```diff
@@ -0,0 +1,457 @@
...
+                self._get_id_fields_for_issue(issue),
+                self._get_project_fields_for_issue(parsed_fields),
+                self._get_dropdown_fields_for_issue(parsed_fields),
+                self._get_subtasks_for_issue(parsed_fields),
+                self._get_comments_for_issue(parsed_fields),
+                self._get_text_fields_for_issue(parsed_fields),
+            ],
+        )
+
+    def update_file_data(self, file_data: FileData, issue: dict) -> None:
```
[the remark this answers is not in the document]
mpolomdeepsense: Good catch, I wanted to use it at the end of `run` function, but I probably forgot to include it there. Changed

unstructured_ingest/v2/processes/connectors/jira.py
```diff
@@ -0,0 +1,457 @@
...
+            [self._get_fields_for_comment(comment) for comment in issue["comment"]["comments"]],
+        )
+
+    def _get_fields_for_comment(
+        self, comment, c_sep: str = DEFAULT_C_SEP, r_sep: str = DEFAULT_R_SEP
+    ) -> str:
+        return f"{comment['author']['displayName']}{c_sep}{comment['body']}{r_sep}"
+
+    def form_templated_string(
+        self, issue: dict, parsed_fields, c_sep: str = "|||", r_sep: str = "\n\n\n"
```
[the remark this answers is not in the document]
mpolomdeepsense: Fixed

unstructured_ingest/v2/processes/connectors/jira.py
```diff
@@ -0,0 +1,457 @@
...
+
+class FieldGetter(dict):
+    def __getitem__(self, key):
+        value = super().__getitem__(key) if key in self else None
+        if value is None:
+            value = FieldGetter({})
+        return value
+
+
+def nested_object_to_field_getter(object: dict) -> Union[FieldGetter, dict]:
```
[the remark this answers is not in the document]
mpolomdeepsense: fixed

unstructured_ingest/v2/processes/connectors/jira.py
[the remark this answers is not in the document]
bryan-unstructured: if only for developing purposes, I think we should remove them. the type hinting usually exists in function or class declaration.

unstructured_ingest/v2/processes/connectors/jira.py
```diff
@@ -0,0 +1,461 @@
...
+        for _ in range(num_iterations):
+            response = func(*args, **kwargs)
+            if isinstance(response, list):
+                all_results += response
+            elif isinstance(response, dict):
+                if results_key not in response:
+                    raise KeyError(f'Response object is missing "{results_key}" key.')
+                all_results += response[results_key]
+            else:
+                raise TypeError(
```
bryan-unstructured, line 115: just thought for food here, I'm interested to hear others' opinion:
even if 1 out of 10 responses has wrong type, the job would fail. Is it better to still complete the job, and log that 1 invalid response instead of raising exception?

unstructured_ingest/v2/processes/connectors/jira.py
```diff
@@ -0,0 +1,461 @@
...
+            with self.connection_config.get_client() as client:
+                # TODO: (Marek Połom) - what do we check here?
+                response = client.get_permissions("BROWSE_PROJECTS")
+                permitted = response["permissions"]["BROWSE_PROJECTS"]["havePermission"]
+        except Exception as e:
+            logger.error(f"Failed to connect to Jira: {e}", exc_info=True)
+            raise SourceConnectionError(f"Failed to connect to Jira: {e}")
+        if not permitted:
+            raise ValueError(
+                """The user with the provided *user_email* and the *api_token*
```
bryan-unstructured, line 205: the error message only mentions api_token, is it possible due to wrong password?

bryan-unstructured:
I've completed my review. left a few more comments.

test_e2e/src/jira.sh
```diff
@@ -64,8 +70,8 @@ PYTHONPATH=${PYTHONPATH:-.} "$RUN_SCRIPT" \
   --output-dir "$OUTPUT_DIR" \
   --verbose \
   --url https://unstructured-jira-connector-test.atlassian.net \
-  --user-email "$JIRA_INGEST_USER_EMAIL" \
-  --api-token "$JIRA_INGEST_API_TOKEN" \
+  --username "$JIRA_INGEST_USER_EMAIL" \
```
potter-potter, line 73: switching user_email and api-token to username and password are breaking changes for users. (luckily not for platform since we don't have the connector yet)

What was the reasoning for this? Not that it is a big deal.

unstructured_ingest/v2/processes/connectors/jira.py
potter-potter: I think a hard fail is fine. We usually build the connectors the best that we can. If the response is something other than a list or dict then we need to understand why and accommodate it.

test_e2e/src/jira.sh
mpolomdeepsense: I looked into the `atlassian-python-api` documentation and noticed that the Confluence client uses the same parameters, so I figured it would be best if we stuck to the same pattern here, too.

This is also how it changed for Confluence connector between v1 and v2.

unstructured_ingest/v2/processes/connectors/jira.py
mpolomdeepsense: changed

unstructured_ingest/v2/processes/connectors/jira.py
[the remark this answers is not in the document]
mpolomdeepsense: changed

bryan-unstructured (approved):

bryan-unstructured:
there are failures in CI, but they seem unrelated to Jira connector.

test_e2e/src/jira.sh
potter-potter: Cool.

Though I personally hate to do it, lets remove this test since we are moving towards only tests in test/integration/connectors/test_jira.py They are much faster since they don't do all the partitioning etc.

potter-potter (approved):
looks good.
just remove the test_e2e test

PR #400, opened 2025-02-25
fix: dropbox connector can now refresh token by itself

mateuszkuprowski:
Dropbox connector was using only short-lived (4h) access token which for testing purposes was generated out of the long lived refresh token and user creds. Now this method is incorporated in the dropbox connector itself so we don't have to keep refreshing configuration.

### DIFF
--- a/unstructured_ingest/v2/processes/connectors/fsspec/dropbox.py
+++ b/unstructured_ingest/v2/processes/connectors/fsspec/dropbox.py
@@ -46,7 +46,10 @@ class DropboxIndexerConfig(FsspecIndexerConfig):
 
 
 class DropboxAccessConfig(FsspecAccessConfig):
-    token: Optional[str] = Field(default=None, description="Dropbox access token.")
+    token: Optional[str] = Field(default=None, description="Dropbox access token.") #This is the short lived (4h) token that needs to be generated anew each time.
+    app_key: Optional[str] = Field(default=None, description="Dropbox app key.")
+    app_secret: Optional[str] = Field(default=None, description="Dropbox app secret.")
+    refresh_token: Optional[str] = Field(default=None, description="Dropbox refresh token.") #This is the long lived token that doesn't expire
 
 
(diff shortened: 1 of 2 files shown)


unstructured_ingest/v2/processes/connectors/fsspec/dropbox.py
```diff
@@ -46,32 +45,95 @@ def model_post_init(self, __context):
...
+                app_key=app_key,
+                app_secret=app_secret,
+            )
+            if not new_token:
+                raise ValueError(
+                    "Unable to retrieve an access token from Dropbox. "
+                    "Please check that your refresh token, app key, and secret are valid."
+                )
+            base_conf["token"] = new_token
+        elif not base_conf.get("token"):  # we might already have an access token from outside
```
bryan-unstructured, line 114: this check works, just want point out an alternative:
use model_post_init() in DropboxAccessConfig to check whether both access token and refresh token are None

unstructured_ingest/v2/processes/connectors/fsspec/dropbox.py
```diff
@@ -46,32 +45,95 @@ def model_post_init(self, __context):
...
+        # Standard scenario - we have refresh a token and creds provided
+        # which we're going to use to retrieve access token
+        if refresh_token and app_key and app_secret:
+            logger.debug("Attempting to generate access token from refresh token...")
+            new_token = self.get_dropbox_access_token_from_refresh(
+                refresh_token=refresh_token,
+                app_key=app_key,
+                app_secret=app_secret,
+            )
+            if not new_token:
```
bryan-unstructured, line 108: when getting access token from refresh token fails, I'm thinking there's still a possibility that an access token is provided by user and that could work?

bryan-unstructured: I see, in that sense it's all good.

bryan-unstructured (approved):

PR #415, opened 2025-03-06
feat: add source permission metadata for google drive

shreyanid:
This PR retrieves the permissions available in the file data of each google drive document and normalizes the format into lists of users and groups with each type of role (read/update/delete etc).
Permission data was already being emitted from this connector, so this PR is just the type normalization step in the indexer.


vangheem (approved):
lg

PR #418, opened 2025-03-07
Zendesk Deps Update

nikpocuca:
Changes for zendesk deps.


unstructured_ingest/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "0.5.11"  # pragma: no cover
```
potter-potter, line 1: this should be 0.5.12-dev0

test/integration/connectors/test_zendesk.py
```diff
@@ -29,6 +29,11 @@ async def zendesk_source_test(
     subdomain: Optional[str] = None,
 ):
 
+    # tmp_path=temp_dir,
```
potter-potter, line 32: you can probably delete these comments

potter-potter (approved):
looks good. potentially remove unneeded comments

PR #424, opened 2025-03-13
feat/update confluence connector

potter-potter:
Confluence accepts either password or api_token for the password field in its SDK. We left that as password field.

But this was causing confusion in the front end UI. We need separate entries. After discussion with Steven we thought it best to make that distinction in OSS. And then present it as 2 different options in the UI.


unstructured_ingest/v2/processes/connectors/confluence.py
```diff
@@ -57,7 +61,12 @@ class ConfluenceConnectionConfig(ConnectionConfig):
 
     def model_post_init(self, __context):
         access_configs = self.access_config.get_secret_value()
-        basic_auth = self.username and access_configs.password
+        if access_configs.password and access_configs.api_token:
+            raise ValueError(
+                "both password and api_token provided, only one allowed, "
+                "see: https://atlassian-python-api.readthedocs.io/"
+            )
+        basic_auth = bool(self.username and (access_configs.password or access_configs.api_token))
```
rbiseck3, line 69: We should also add a check that at least one was provided. I think this won't break if both are left as `None`

rbiseck3 (approved):

PR #426, opened 2025-03-13
feat/add support for ndjson files

rbiseck3:


unstructured_ingest/utils/data_prep.py
```diff
@@ -163,7 +163,10 @@ def write_data(path: Path, data: list[dict], indent: Optional[int] = 2) -> None:
             raise IOError("Unsupported file type: {path}")
 
 
-def get_data(path: Path) -> list[dict]:
+def get_data(path: Union[Path, str]) -> list[dict]:
+    if isinstance(path, str):
+        path = Path(path)
+    path = Path(path)
```
potter-potter, line 169: this seems a little strange

rbiseck3: Oh ya that's a bit redundant. I can remove the second one.

PR #428, opened 2025-03-18
feat: IBM watsonx.data Destination connector

mpolomdeepsense:


potter-potter:
I'll check this out later today.

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
```diff
@@ -0,0 +1,298 @@
...
+    from pyiceberg.table import Table, Transaction
+
+CONNECTOR_TYPE = "ibm_watsonx_data"
+
+DEFAULT_IBM_CLOUD_AUTH_URL = "https://iam.cloud.ibm.com/identity/token"
+DEFAULT_ICEBERG_URI_PATH = "/mds/iceberg"
+DEFAULT_ICEBERG_CATALOG_TYPE = "rest"
+
+
+@requires_dependencies(["pyiceberg"], extras="ibm-watsonx-data")
```
potter-potter, line 41: we are missing the other pieces to allow installing this dependency correctly. like in setup.py (search for pinecone and see where these extra pieces live)

mpolomdeepsense: Found only one place where it was missing (`setup.py`). Unless we want to add an example into `unstructured_ingest/v2/examples`.

rbiseck3:
Is `ibm_watson` sufficient? I feel like `ibm_watson_data` is overly verbose and redundant.

rbiseck3 (changes requested):
Left some comments

rbiseck3:
I'm assuming this new connector wasn't registered, when I run the cli I don't see watson listed under the destination connectors.

rbiseck3:
Can we wrap errors using these rather than the legacy `DestinationConnectionError` :[unstructured_ingest/v2/errors.py](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/errors.py)

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
```diff
@@ -0,0 +1,298 @@
...
+            table = catalog.load_table(self.upload_config.table_identifier)
+            yield table
+
+    @requires_dependencies(["pyarrow"], extras="ibm-watsonx-data")
+    def _get_data_table(self, table: "Table", path: Path) -> Any:
+        import pyarrow as pa
+
+        df = get_data_df(path)
+        table_column_names = table.schema().column_names
+        df_column_names = list(df.columns.values)
```
rbiseck3, line 249:
```suggestion
        df_column_names = df.columns.values.tolist()
```

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
```diff
@@ -0,0 +1,298 @@
...
+
+CONNECTOR_TYPE = "ibm_watsonx_data"
+
+DEFAULT_IBM_CLOUD_AUTH_URL = "https://iam.cloud.ibm.com/identity/token"
+DEFAULT_ICEBERG_URI_PATH = "/mds/iceberg"
+DEFAULT_ICEBERG_CATALOG_TYPE = "rest"
+
+
+@requires_dependencies(["pyiceberg"], extras="ibm-watsonx-data")
+def _transaction_wrapper(table: "Table", fn: Callable, max_retries: int, **kwargs: Any) -> None:
```
rbiseck3, line 42: Can we potentially use this library to handle the retry logic? We use this elsewhere in our org repos: [tenacity](https://tenacity.readthedocs.io/en/latest/)

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+class IbmWatsonxDataConnectionConfig(ConnectionConfig):
+    access_config: Secret[IbmWatsonxDataAccessConfig]
+    iceberg_endpoint: str = Field(description="Iceberg REST endpoint")
+    object_storage_endpoint: str = Field(description="Cloud Object Storage public endpoint")
+    object_storage_region: str = Field(description="Cloud Object Storage region")
+    catalog: str = Field(description="Catalog name")
+
+    _bearer_token: Union[dict[str, Any], None] = None
```
rbiseck3, line 92:
```suggestion
    _bearer_token: Optional[dict[str, Any]] = None
```

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
```diff
@@ -0,0 +1,298 @@
...
+                )
+
+    @contextmanager
+    def get_table(self) -> Generator["Table", None, None]:
+        with self.connection_config.get_catalog() as catalog:
+            table = catalog.load_table(self.upload_config.table_identifier)
+            yield table
+
+    @requires_dependencies(["pyarrow"], extras="ibm-watsonx-data")
+    def _get_data_table(self, table: "Table", path: Path) -> Any:
```
rbiseck3, line 244: Can we reuse the logic from the sql connector? [_fit_to_schema()](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/processes/connectors/sql/sql.py#L326)

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
```diff
@@ -0,0 +1,298 @@
...
+        """
+        Uploads data to a specified table.
+
+        This method handles the process of uploading data to a table, including
+        deleting existing records if necessary and appending new data. It uses
+        a transaction wrapper to ensure the operation is retried if it fails.
+        """
+        from pyiceberg.expressions import EqualTo
+
+        def _upload_data(transaction: "Transaction", data_table: Any, file_data: FileData) -> None:
```
rbiseck3, line 271: Should this be typed as a dataframe? 
```suggestion
        def _upload_data(transaction: "Transaction", data_table: pd.DataFrame, file_data: FileData) -> None:
```

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
```diff
@@ -0,0 +1,298 @@
...
+            if self.upload_config.record_id_key in transaction._table.schema().column_names:
```
rbiseck3, line 272: Can we isolate the delete logic and call it as part of the `run` method rather than having it as part of the `_upload_data` method to make that more explicit?

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
```diff
@@ -0,0 +1,298 @@
...
+    connector_type: str = CONNECTOR_TYPE
+
+    def precheck(self) -> None:
+        with self.connection_config.get_catalog() as catalog:
+            if not catalog.namespace_exists(self.upload_config.namespace):
+                raise DestinationConnectionError(
+                    f"Namespace '{self.upload_config.namespace}' does not exist"
+                )
+            if not catalog.table_exists(self.upload_config.table_identifier):
+                raise DestinationConnectionError(
```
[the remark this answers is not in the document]
mpolomdeepsense: changed

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
mpolomdeepsense: changed

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
mpolomdeepsense: Changed, though now the number of retries can't be controlled by the user. However, it's not a big deal.

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
mpolomdeepsense: changed

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
mpolomdeepsense: Good idea, done. Also made the whole uploader subclass from SQL and made the whole uploader flow similar to the SQL one.

Should I also move the whole connector into `sql` directory?

mpolomdeepsense: I added an additional argument to `_fit_to_schema` called `add_missing_columns = True`. Iceberg automatically fills missing columns with null and when you try to input them manually it throws an error `[upload] Failed to upload data to table: Column 'links' has an unsupported type: null`. From what I understand it accepts null values but can't identify the right type from dataframe and match it to the table schema.

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
mpolomdeepsense: changed

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
mpolomdeepsense: I wanted to make delete and append using a single transaction, that's why it uses the same logic.

unstructured_ingest/v2/processes/connectors/ibm_watsonx_data.py
mpolomdeepsense: Nvm, after tenacity dependency fix max_retires is back.

unstructured_ingest/v2/processes/connectors/ibm_watsonx.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+class IbmWatsonxConnectionConfig(ConnectionConfig):
+    access_config: Secret[IbmWatsonxAccessConfig]
+    iceberg_endpoint: str = Field(description="Iceberg REST endpoint")
+    object_storage_endpoint: str = Field(description="Cloud Object Storage public endpoint")
+    object_storage_region: str = Field(description="Cloud Object Storage region")
+    catalog: str = Field(description="Catalog name")
+
+    _bearer_token: Optional[dict[str, Any]] = None
```
rbiseck3, line 60:
```suggestion
    _bearer_token: Optional[dict[str, Any]] = Field(init=False, default=None)
```

unstructured_ingest/v2/processes/connectors/ibm_watsonx.py
```diff
@@ -0,0 +1,297 @@
...
+        bearer_token = self.bearer_token
+        try:
+            catalog = load_catalog(
+                self.catalog,
+                **{
+                    "type": DEFAULT_ICEBERG_CATALOG_TYPE,
+                    "uri": self.iceberg_url,
+                    "token": bearer_token,
+                    "warehouse": self.catalog,
+                    "s3.endpoint": self.object_storage_url,
```
rbiseck3, line 147: Is ibm watson always backed by s3? Is this something we want configurable for other blob stores from other cloud providers if that's supported?

unstructured_ingest/v2/processes/connectors/ibm_watsonx.py
```diff
@@ -0,0 +1,297 @@
...
+                    self._delete(transaction, file_data.identifier)
+                    self._append(transaction, data_table)
+            except CommitFailedException as e:
+                table.refresh()
+                logger.debug(e)
+                raise IcebergCommitFailedException(e)
+            except Exception as e:
+                raise ProviderError(f"Failed to upload data to table: {e}")
+
+        return _upload_data_table(table, data_table, file_data)
```
rbiseck3, line 274: Should also wrap this in a try-except in case the reties were exhausted and threw the original `CommitFailedException`, in which case we should wrap this in a `ProviderError` as well.

unstructured_ingest/v2/processes/connectors/ibm_watsonx.py
mpolomdeepsense: I think you can make it use different type of storage but from what I understand the specific example they sent us and they need implementation for is using S3.

unstructured_ingest/v2/processes/connectors/ibm_watsonx.py
```diff
@@ -0,0 +1,297 @@
...
+        @retry(
+            stop=stop_after_attempt(self.upload_config.max_retries),
+            wait=wait_random(),
+            retry=retry_if_exception_type(IcebergCommitFailedException),
+            before=before_log(logger, logging.DEBUG),
+        )
+        def _upload_data_table(table: "Table", data_table: "ArrowTable", file_data: FileData):
+            try:
+                with table.transaction() as transaction:
+                    self._delete(transaction, file_data.identifier)
```
rbiseck3, line 265: We might want to break this up into a delete transaction and an append transaction where we wrap each in their own tenacity retry loop. Otherwise this is going to retry the delete each time even if that already passed or doesn't apply.

unstructured_ingest/v2/processes/connectors/ibm_watsonx.py
```diff
@@ -0,0 +1,297 @@
...
+    def upload_data_table(
+        self, table: "Table", data_table: "ArrowTable", file_data: FileData
+    ) -> None:
+        from pyiceberg.exceptions import CommitFailedException
+        from tenacity import (
+            before_log,
+            retry,
+            retry_if_exception_type,
+            stop_after_attempt,
+            wait_random,
```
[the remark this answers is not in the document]
mpolomdeepsense: nice catch

unstructured_ingest/v2/processes/connectors/ibm_watsonx.py
mpolomdeepsense: @raiderrobert

unstructured_ingest/v2/processes/connectors/ibm_watsonx.py
mpolomdeepsense: Getting this error:
`NameError: Fields must not use names with leading underscores; e.g., use 'bearer_token' instead of '_bearer_token'.`

Reverting this change

mpolomdeepsense: This is a private parameter, not intended to be accessed by the user, I don't see any point of making this a Field.

unstructured_ingest/v2/processes/connectors/ibm_watsonx.py
mpolomdeepsense: I think it's a bad idea. I ran some tests and:

1. Because of the way Iceberg works this is going to create a lot of retries.
2. Tried partitioning and uploading 4 files with separate transactions and it fails (on delete) when using `max_retries=5` for 1 file (so 3 out of 4 files got uploaded). Worked after changing it to `max_retries=10`
3. Because of so many retries it makes this uploader very slow.
4. Why even make transaction? Isn't the point of transaction to rollback the changes in case of a fail? After split we have two single operations that shouldn't require transaction.

unstructured_ingest/v2/processes/connectors/ibm_watsonx.py
```diff
@@ -0,0 +1,297 @@
...
+        if self.can_delete():
+            transaction.delete(delete_filter=EqualTo(self.upload_config.record_id_key, identifier))
+        else:
+            logger.warning(
+                f"Table doesn't contain expected "
+                f"record id column "
+                f"{self.upload_config.record_id_key}, skipping delete"
+            )
+
+    def _append(self, transaction: "Transaction", data_table: "ArrowTable") -> None:
```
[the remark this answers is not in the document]
mpolomdeepsense: changed

unstructured_ingest/v2/processes/connectors/ibm_watsonx.py
mpolomdeepsense: changed

unstructured_ingest/v2/processes/connectors/ibm_watsonx.py
rbiseck3: Then we should follow the same pattern we have for other connectors that can be backed by various blob stores, so we should have an s3 suffix for this one and create a method, i.e. `def get catalog_configs()-> dict` that gets called to generate this on a base class. For now, we don't need to introduce this base class, but at least add that method so if we introduce other blob stores, we could easily support those. Take a look at databricks volumes for an example of a connector supporting multiple blob stores.

mpolomdeepsense: Actually, it could even be more precise `ibm-watsonx-iceberg-s3`, but let's stick with `ibm-watsonx-s3` for now, we can always change it later.

PR #429, opened 2025-03-19
Zendesk fix

nikpocuca:
Trying to debug whats going on with platform and this missing content variable for articles.

### DIFF
--- a/unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
+++ b/unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
@@ -48,10 +48,11 @@ class ZendeskFileDataSourceMetadata(FileDataSourceMetadata):
     as the source metadata.
     """
 
+
 class ZendeskBatchFileData(BatchFileData):
     additional_metadata: ZendeskAdditionalMetadata
     metadata: ZendeskFileDataSourceMetadata
-    batch_items: List[Union[ZendeskBatchItemArticle,ZendeskBatchItemTicket]]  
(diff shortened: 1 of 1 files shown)


unstructured_ingest/v2/processes/connectors/zendesk/zendesk.py
```diff
@@ -310,6 +314,7 @@ async def handle_articles_async(
 
         async with aiofiles.open(download_path, "a", encoding="utf8") as f:
             for article in batch_file_data.batch_items:
+
```
rbiseck3, line 317: We should be using the same approach we have elsewhere regarding casting the file data when there's a specific implementation: [sql.py#L201](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/processes/connectors/sql/sql.py#L201)

rbiseck3: Because that was never set on the new implementation of `BatchFileData`, this should work: 
```
class ZendeskBatchFileData(BatchFileData):
    additional_metadata: ZendeskAdditionalMetadata
    metadata: ZendeskAdditionalMetadata
```

rbiseck3: But I also don't see anything new defined in the new implementation of `FileDataSourceMetadata` here, what's the purpose? 
```
class ZendeskFileDataSourceMetadata(FileDataSourceMetadata):
    """
    inherits metadata object as tickets and articles
    are treated in single batch, we need to denote indices ticket/article
    as the source metadata.
    """
```

amadeusz-ds (approved):

PR #434, opened 2025-03-20
feat: add Confluence source permission metadata

shreyanid:
Retrieves permission data from each Confluence document and its parent space, normalizes the format (see below), and adds it to the `permissions_data` field in the metadata.

The output format is a dictionary of roles (read, update, delete) with a lists of "groups" and users" having that permission under each.

```
"permissions_data": [{
     "read": {
       "groups": [...,],
       "users": [...,], 
     },
     "update": {
       "groups": [...,],
       "users": [...,], 
     },
     "delete": {
       "groups": [...,],
       "users": [...,], 
     },
 }]
```
The output type is currently wrapped in a list to match the existing FileDataSourceMetadata type [definition](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/interfaces/file_data.py#L32) shared by all connectors. TODO to update that as the other connectors get normalized

The current implementation limits the results to approximately 250 permission entities.

Further improvement items:
- expand users and groups (apply group permissions as enabled for user of the group)
- retrieve email based on ID to populate the email_address field
- enforce the normalized format via the type of permissions_data in FileDataSourceMetadata (instead of a generic [list[dict]](https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/v2/interfaces/file_data.py#L32))

This PR also clarifies "space_id" (int) vs "space_key" (str). The previous `space_id` used throughout the code was actually the space key. Does not change the user facing experience (the space_id field returned in metadata is still the space key as before)


unstructured_ingest/v2/processes/connectors/confluence.py
```diff
@@ -134,54 +133,201 @@ def precheck(self) -> bool:
...
                 status=None,
             )
-        doc_ids = [{"space_id": space_id, "doc_id": page["id"]} for page in pages]
+        doc_ids = [{"space_id": space_key, "doc_id": page["id"]} for page in pages]
         return doc_ids
 
+    def _get_permissions_for_doc(self, space_id: int, doc_id: str) -> Optional[List[dict]]:
+        from requests.exceptions import HTTPError
+
+        def parse_permissions(doc_permissions: dict, space_permissions: list):
```
vangheem, line 173: Does this function need to be inline like this? Maybe it's just me but it makes it a little harder to read like this.

unstructured_ingest/v2/processes/connectors/confluence.py
```diff
@@ -134,54 +133,201 @@ def precheck(self) -> bool:
...
+                space_target_type = space_perm["operation"]["targetType"]
+                space_entity_id = space_perm["principal"]["id"]
+                space_entity_type = space_perm["principal"]["type"]
+
+                if space_target_type == "space":
+                    if space_operation == "read" and not page_view_restricted:
+                        # Apply space-level view permissions if no page restrictions exist
+                        perm_entry = normalized_perms_dict.setdefault(
+                            space_entity_id,
+                            {
```
vangheem, line 247: Can we concretely define the unified format we want to store permissions in somewhere outside of this module?

unstructured_ingest/v2/processes/connectors/confluence.py
```diff
@@ -134,54 +133,201 @@ def precheck(self) -> bool:
...
                 expand=None,
-                content_type="page",
+                content_type="page",  # blogpost and comment types not currently supported
                 status=None,
             )
-        doc_ids = [{"space_id": space_id, "doc_id": page["id"]} for page in pages]
+        doc_ids = [{"space_id": space_key, "doc_id": page["id"]} for page in pages]
         return doc_ids
 
+    def _get_permissions_for_doc(self, space_id: int, doc_id: str) -> Optional[List[dict]]:
```
vangheem, line 170: can we fully define the output type here instead of `dict`?

I think we need a normalized/unified way to export permissions

unstructured_ingest/v2/processes/connectors/confluence.py
```diff
@@ -134,54 +133,201 @@ def precheck(self) -> bool:
...
+            for perm_data in normalized_perms_dict.values():
+                perm_data["roles"] = sorted(perm_data["roles"])
+
+            return list(normalized_perms_dict.values())
+
+        with self.connection_config.get_client() as client:
+            try:
+                # space permissions
+                space_permissions = []
+                space_permissions_result = client.get(f"/api/v2/spaces/{space_id}/permissions")
```
vangheem, line 290: Just asking(maybe not possible), is there any way to get any of this data in the original request to download the doc? (just to reduce the number of api calls)

vangheem (changes requested):

unstructured_ingest/v2/processes/connectors/confluence.py
shreyanid: moved it out!

unstructured_ingest/v2/processes/connectors/confluence.py
```diff
@@ -136,54 +138,200 @@ def precheck(self) -> bool:
...
+        # TODO adjust permissions_data FileDataSourceMetadata interface type to match/enforce format
+        return [parsed_permissions_dict]
+
     def run(self) -> Generator[FileData, None, None]:
         from time import time
 
-        space_ids = self._get_space_ids()
-        for space_id in space_ids:
-            doc_ids = self._get_docs_ids_within_one_space(space_id)
+        space_ids_and_keys = self._get_space_ids_and_keys()
```
vangheem, line 311: Should this be cached?

unstructured_ingest/v2/processes/connectors/confluence.py
```diff
@@ -136,54 +138,200 @@ def precheck(self) -> bool:
...
+        for space_key, space_id in space_ids_and_keys:
+            doc_ids = self._get_docs_ids_within_one_space(space_key)
+            space_perm = self._get_permissions_for_space(space_id)
```
vangheem, line 314: Can we cache this across calls to the run method?

The way I read this, we could call this for every file when it's not necessary

vangheem (commented):
I think we need some more caching here yet.

unstructured_ingest/v2/processes/connectors/confluence.py
shreyanid: am now passing the space_id through additional_metadata through to the downloader, moved all the permissions logic to the downloader, and am caching the function to get space_permissions across runs. 

let me know if you recommend a different cache size (the resulting space permission return value could get fairly large so maybe we should't have too large a cache)

unstructured_ingest/v2/processes/connectors/confluence.py
```diff
@@ -234,6 +248,143 @@ def download_embedded_files(
...
+        # turn sets into sorted lists for consistency and json serialization
+        for role_dict in permissions_by_role.values():
+            for key in role_dict:
+                role_dict[key] = sorted(role_dict[key])
+
+        return permissions_by_role
+
+    def _get_permissions_for_space(self, space_id: int) -> Optional[List[dict]]:
+        @lru_cache(maxsize=128)
+        def cached(_space_id: int):
```
vangheem, line 352: Why is this a function nested in here like this?

Doing this will mean you'll never actually cache anything because the cache/function will be redefined every time this is called.

vangheem (changes requested):

unstructured_ingest/v2/processes/connectors/confluence.py
shreyanid: hmm. but then if I do it on the outer function, `self` is not a valid hashable argument for building the cache, and I'm getting an error. 
```
  File "/Users/shreyanid/Documents/all-unstructured/unstructured-ingest/unstructured_ingest/v2/processes/connectors/confluence.py", line 425, in run
    space_perm = self._get_permissions_for_space(space_id)  # must be the id, NOT the space key
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: unhashable type: 'ConfluenceDownloader'

File: unstructured_ingest/processes/connectors/confluence.py [hunk 8]
```diff
@@ -234,6 +250,147 @@ def download_embedded_files(
...
+                    space_permissions_result = client.get(f"/api/v2/spaces/{space_id}/permissions")
+                    space_permissions.extend(space_permissions_result["results"])
+                    if space_permissions_result["_links"].get("next"):  # pagination
+                        while space_permissions_result.get("next"):
+                            space_permissions_result = client.get(space_permissions_result["next"])
+                            space_permissions.extend(space_permissions_result["results"])
+
+                    if len(self._permissions_cache) >= self._permissions_cache_max_size:
+                        self._permissions_cache.popitem(last=False)  # FIFO eviction
+                    self._permissions_cache[space_id] = space_permissions
```
vangheem, line 371: since we're using `OrderedDict`, I guess we can also call `self._permissions_cache.move_to_end(space_id)` to make it behave like a LRU cache.

unstructured_ingest/processes/connectors/confluence.py
```diff
@@ -234,6 +250,147 @@ def download_embedded_files(
...
+
+                    return space_permissions
+                except Exception as e:
+                    logger.debug(f"Could not retrieve permissions for space {space_id}: {e}")
+                    return None
+
+    def _parse_permissions_for_doc(self, doc_id: str, space_permissions: list) -> Optional[dict]:
+        from requests.exceptions import HTTPError
```
vangheem, line 379: Why is this import nested here? I think you can put that up top.

unstructured_ingest/processes/connectors/confluence.py
```diff
@@ -234,6 +250,147 @@ def download_embedded_files(
...
+                doc_permissions = client.get_all_restrictions_for_content(content_id=doc_id)
+                parsed_permissions_dict = self.parse_permissions(doc_permissions, space_permissions)
+
+            except HTTPError as e:
+                # skip writing any permission metadata
+                logger.debug(f"Could not retrieve permissions for doc {doc_id}: {e}")
+                return None
+
+        # TODO adjust permissions_data FileDataSourceMetadata interface type to match/enforce format
+        return [parsed_permissions_dict]
```
vangheem, line 392: I think we should return a dict here instead of a single item list.

vangheem (changes requested):

unstructured_ingest/processes/connectors/confluence.py
shreyanid: added in cache hit case

unstructured_ingest/processes/connectors/confluence.py
shreyanid: moved

unstructured_ingest/processes/connectors/confluence.py
shreyanid: updated the type

unstructured_ingest/processes/connectors/confluence.py
```diff
@@ -233,6 +250,146 @@ def download_embedded_files(
...
+                    if len(self._permissions_cache) >= self._permissions_cache_max_size:
+                        self._permissions_cache.popitem(last=False)  # LRU/FIFO eviction
+                    self._permissions_cache[space_id] = space_permissions
+
+                    return space_permissions
+                except Exception as e:
+                    logger.debug(f"Could not retrieve permissions for space {space_id}: {e}")
+                    return None
+
+    @requires_dependencies(["requests"], extras="confluence")
```
vangheem, line 379: You refactored this to no longer use requests. I guess it's fine to still use requests and put the import inline.

vangheem (approved):

unstructured_ingest/processes/connectors/confluence.py
shreyanid: either works right? I went ahead with catching a generic exception since we really don't want this to causing jobs to err, so ill remove this dep

PR #445, opened 2025-03-25
feat/fix databricks delta tables

rbiseck3:
### Description
This pulls in the columns for the destination table to be uploaded to and makes sure the data to be uploaded matches the destination schema.


potter-potter (approved):
lgtm

PR #449, opened 2025-03-27
feat/databricks opinionated writes

rbiseck3:
### Description
Updates to the databricks connectors:
* Volumes connector now emits user agent
* databricks delta tables has support for opinionated writes, creating the table on a user's behalf
* databricks delta table leverages the blob support for metadata and no longer flattens the content
* databricks delta table connector deletes previous content if the proper column exists in the table


unstructured_ingest/v2/processes/connectors/databricks/volumes_table.py
```diff
@@ -1,43 +1,71 @@
...
     DatabricksDeltaTablesConnectionConfig,
-    DatabricksDeltaTablesUploadStager,
     DatabricksDeltaTablesUploadStagerConfig,
 )
 from unstructured_ingest.v2.types.file_data import FileData
+from unstructured_ingest.v2.utils import get_enhanced_element_id
 
 CONNECTOR_TYPE = "databricks_volume_delta_tables"
 
 if TYPE_CHECKING:
```
vangheem, line 32: Let's just remove this instead of a pass here

vangheem (approved):
just one tweak

PR #457, opened 2025-03-31
feat/use latest astradb client version

rbiseck3:


vangheem (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.0.2
+
+### Features
+
+* **Update astra source connector to use new astrapy client**
```
shreyanid, line 5: this updates both the source and destination astra connector, right?

potter-potter: @shreyanid It covers the entire astradb connector

potter-potter: Because above 2.0 is causing me an error

rbiseck3: Above `2.0.0` is where the client was updated causing the breaking changes. By pinning it, we make sure out implementation will work with the dependency that gets pulled in.

PR #459, opened 2025-04-02
revert back to 1.0.2

shreyanid:
This reverts commit cb1277a, 81d6425, and c70ee0b.


shreyanid:
partial revert / fix was made in #461 instead

PR #460, opened 2025-04-02
Fix stripping extensions in google drive

mateuszkuprowski:
I noticed current Google Drive config is not cutting leading dots properly leading to fails when extensions are provided with a dot. This is because wrong type of post init was used, dataclass approach instead of pydantic was used and the extension fix was beeing omitted.


pawel-kmiecik (approved):
LGTM!

PR #461, opened 2025-04-02
bugfix/fix permissions

rbiseck3:
### Description
The `FileData` schema requires that permission data come as a list. The recent changes to introduce permission data to the google drive and confluence connectors didn't conform to that schema. This fixes that.


test/integration/connectors/expected_results/google_drive_source/downloads/ingest_integration_test_do_not_change
potter-potter, line 1: What is this file? curious

potter-potter (approved):
lg

PR #466, opened 2025-04-08
feat: google drive internal formats support

mateuszkuprowski:
This implementation replaces the older approach using the Google Drive API's
    `files.export()` and `files.get_media()` methods, which were limited by export size
    restrictions (~10MB hard cap for Google-native files).

    **New mechanism:**
    - Authenticates using a service account and generates a bearer token via `google-auth`.
    - Uses `files().get(fileId=..., fields="exportLinks,webContentLink")` to fetch download URLs:
        - `exportLinks` is a dictionary of export MIME types (e.g., docx, pdf, pptx) for
          Google-native files like Docs, Sheets, and Slides.
        - `webContentLink` is a pre-authenticated direct download link for non-Google-native
          files (e.g., PDFs, images, ZIPs, etc.).
          
    These links emulate the behavior of Google Drive's "File > Download as..." options
    in the UI and bypass the size limitations of `files.export()`.
    
     **Behavior:**
    - Google-native formats are downloaded using `exportLinks` in appropriate MIME formats.
    - Binary files (non-Google-native) are downloaded using `webContentLink`.
    - All downloads are performed via `requests.get()` using a valid bearer token.


unstructured_ingest/processes/connectors/google_drive.py
```diff
@@ -255,7 +267,9 @@ def precheck(self) -> None:
                         #     that the service account has proper permissions."
                         # )
                     else:
-                        logger.info(f"Found {file_count} files recursively in the folder.")
+                        logger.info(
```
rbiseck3, line 270: Seems like some odd formatting opdates?

unstructured_ingest/processes/connectors/google_drive.py
```diff
@@ -478,61 +496,144 @@ class GoogleDriveDownloaderConfig(DownloaderConfig):
...
+        )
         with open(download_path, "wb") as handler:
             handler.write(file_contents.getbuffer())
-        return self.generate_download_response(file_data=file_data, download_path=download_path)
+        return self.generate_download_response(
+            file_data=file_data, download_path=download_path
+        )
 
-    @requires_dependencies(["googleapiclient"], extras="google-drive")
+    @requires_dependencies(["googleapiclient", "requests"], extras="google-drive")
```
rbiseck3, line 615: This annotation should be wrapping the method that has the `import` statement in it. It looks like the import from this method was actually removed/

unstructured_ingest/processes/connectors/google_drive.py
mateuszkuprowski: Yeah, I think a linter run must have done it, I'll try to manually tuck it in and see what happens

unstructured_ingest/processes/connectors/google_drive.py
mateuszkuprowski: Good catch, you are totally right, I've moved these imports along with the logic into download function and forgot to change it here. Updating

unstructured_ingest/processes/connectors/google_drive.py
```diff
@@ -478,61 +494,144 @@ class GoogleDriveDownloaderConfig(DownloaderConfig):
...
+            return url, ext
+
+        if not web_link:
+            raise SourceConnectionError(
+                f"No webContentLink available for file {file_id}"
+            )
+        return web_link, ""
+
+    @requires_dependencies(["requests", "google.auth"], extras="google-drive")
+    def _download_url(self, url: str) -> io.BytesIO:
```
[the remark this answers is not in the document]
mateuszkuprowski: Good point I'll modify this bit :)

unstructured_ingest/processes/connectors/google_drive.py
```diff
@@ -478,61 +494,144 @@ class GoogleDriveDownloaderConfig(DownloaderConfig):
...
+            metadata = client.get(
+                fileId=file_id, fields="exportLinks,webContentLink"
+            ).execute()
+
+        export_links = metadata.get("exportLinks", {})
+        web_link = metadata.get("webContentLink")
+
+        export_mime = GOOGLE_EXPORT_MIME_MAP.get(mime_type)
+
+        if export_mime:
```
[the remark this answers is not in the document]
mateuszkuprowski: Looks good, will apply :)

pawel-kmiecik (approved):
LGTM!

bryan-unstructured:
well, this is not related to your PR, but I find a previously existing bug :-) Can you please also update it?
between Line 444 - 447
```
def extract_permissions(self, permissions: Optional[list[dict]]) -> list[dict]:
        if not permissions:
            logger.debug("no permissions found")
            return {}
```

Issue #439, opened 2025-03-21
Unrecognized keyword argument in all Notion Database properties


teddysupercuts:
It seems that we are passing in `description` into `SelectOption.__init__` inside of `types/database_properties/select.py`

The documentation is confusing, because it SEEMS like `Select` doesn't support a description field:

<img width="790" alt="Image" src="https://github.com/user-attachments/assets/c870ab30-ca5e-4702-be5a-e0b52cfe9c43" />

But, Notion says this field **may** exist on ALL database properties, so we should expect it:

<img width="431" alt="Image" src="https://github.com/user-attachments/assets/fde56f1a-e353-45e8-8867-a8657828638c" />

Here is an example of a notion `Select` property with a description. It is kind of hidden but its there. 

<img width="327" alt="Image" src="https://github.com/user-attachments/assets/d5fdbc66-3bb9-412a-8c9d-2c5fe3ce34d4" />

Here is the error:

```
TypeError: StatusOption.__init__() got an unexpected keyword argument 'description'


 Traceback (most recent call last):
   File "/root/venv/lib/python3.12/site-packages/unstructured_ingest/pipeline/source.py", line 62, in run
     return self.get_single(doc=doc, ingest_doc_dict=ingest_doc_dict)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/root/venv/lib/python3.12/site-packages/unstructured_ingest/pipeline/source.py", line 36, in get_single
     doc.get_file()
   File "/root/venv/lib/python3.12/site-packages/unstructured_ingest/interfaces.py", line 534, in wrapper
     return func(self, *args, **kwargs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/root/venv/lib/python3.12/site-packages/unstructured_ingest/utils/dep_check.py", line 57, in wrapper
     return func(*args, **kwargs)
            ^^^^^^^^^^^^^^^^^^^^^
   File "/root/venv/lib/python3.12/site-packages/unstructured_ingest/connector/notion/connector.py", line 220, in get_file
     text_extraction = extract_database_html(
                       ^^^^^^^^^^^^^^^^^^^^^^
   File "/root/venv/lib/python3.12/site-packages/unstructured_ingest/connector/notion/helpers.py", line 149, in extract_database_html
[...]

PR #468, opened 2025-04-09
Fix: Add 'description' field to notion database Select property

teddysupercuts:
See https://github.com/Unstructured-IO/unstructured-ingest/issues/439

Tested this by adding this line to `/Users/<user>/Library/Python/3.9/lib/python/site-packages/unstructured_ingest` and the Notion db download error went away.


teddysupercuts:
Closing in favor of https://github.com/Unstructured-IO/unstructured-ingest/pull/471

PR #473, opened 2025-04-15
feat/add precheck support for embedders

rbiseck3:
### Description
When possible, run precheck for embedder to make sure things like model provided matches one that's available.


amadeusz-ds (approved):
LGTM

mpolomdeepsense (approved):
lg

PR #481, opened 2025-04-21
fix Jira connector cloud option not working issue

bryan-unstructured:


mpolomdeepsense (approved):
LGTM, are there any integration/e2e tests that would cover this case? Would be nice if we could add one.

bryan-unstructured:
> LGTM, are there any integration/e2e tests that would cover this case? Would be nice if we could add one.

The original integration test didn't include `cloud` option, so it ran with default value where `cloud` is set to False. I've updated the connection config to have `cloud=True`
```
connection_config = JiraConnectionConfig(
        url=jira_url,
        username=user_email,
        access_config=access_config,
        cloud=True,
    )
```

PR #486, opened 2025-04-23
Support optional 'description' field for all Notion db properties

teddysupercuts:
Closes https://github.com/Unstructured-IO/unstructured-ingest/issues/439

TLDR: All database property objects returned by the Notion API have an optional `description` field that we are handling. Currently, when it is present the database will simply fail to sync.


teddysupercuts:
Closed in favor of https://github.com/Unstructured-IO/unstructured-ingest/pull/490

PR #501, opened 2025-05-02
Fix broken sync when Notion callout has no Icon

teddysupercuts:


mateuszkuprowski (approved):
Nicely done! We'll integrate this solution!

PR #502, opened 2025-05-05
fix: Snowflake Destination schema example and vector type embeddings

mpolomdeepsense:
1. Fixes table schema example
2. Fixes issue with embeddings column when using vector type


amadeusz-ds (approved):

test/integration/connectors/env_setup/sql/snowflake/destination/snowflake-schema.sql
```diff
@@ -1,43 +1,46 @@
 CREATE TABLE elements (
-    id UUID PRIMARY KEY,
+    id VARCHAR(36) PRIMARY KEY NOT NULL DEFAULT UUID_STRING(),
```
mateuszkuprowski, line 2: You sure it's 36? Is that our element ID used here? I think ours was 32?

test/unit/connectors/sql/test_snowflake.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+@pytest.fixture
+def mock_cursor(mocker: MockerFixture):
+    return mocker.MagicMock()
+
+
+@pytest.fixture
+def mock_get_cursor(mocker: MockerFixture, mock_cursor: MagicMock):
+    mock = mocker.patch.context_manager(SnowflakeConnectionConfig, "get_cursor", autospec=True)
```
mateuszkuprowski, line 52: I think we should use:
`mocker.patch("unstructured_ingest.processes.connectors.sql.snowflake.SnowflakeConnectionConfig.get_cursor")`

So that this does not silently fail when get_cursor is not present. _Also check that path not sure if I got it right._

test/integration/connectors/env_setup/sql/snowflake/destination/snowflake-schema.sql
mpolomdeepsense: Nope, that's not `element_id`. There is a separate `element_id` column of general type `VARCHAR`.

test/unit/connectors/sql/test_snowflake.py
mpolomdeepsense: changed, thanks!

PR #504, opened 2025-05-05
Fix callouts with FileIcons

teddysupercuts:


teddysupercuts:
Pehaps @mpolomdeepsense can also help

PR #505, opened 2025-05-06
fix: don't require port and ssl params if URI is used for Redis connection

pawel-kmiecik:


bryan-unstructured (approved):

unstructured_ingest/processes/connectors/redisdb.py
```diff
@@ -41,20 +41,32 @@ class RedisConnectionConfig(ConnectionConfig):
         default=RedisAccessConfig(), validate_default=True
     )
     host: Optional[str] = Field(
-        default=None, description="Hostname or IP address of a Redis instance to connect to."
+        default=None,
+        description="Hostname or IP address of a Redis instance to connect to "
+                    "if uri is not specified.",
```
mpolomdeepsense, line 46: Let's also change `password` description to match the same pattern `Password used to connect to database if uri is not specified.`

PR #507, opened 2025-05-12
[P6M-1327]: Make user_pname optional for Sharepoint

potter-potter:
Due to Sharepoint inheriting Onedrive, we originally had user_pname required. But it isn't really required.

It's actually not even used unless the User user username/password as their form of Auth which is not recommended by Microsoft.

SharepointConnectionConfig inherits from OnedriveConnectionConfig , which is where the use and declaration of this arg is made. With the changes in the PR, we override the arg to be Optional

Also updated the vertexai model which was deprecated.

### DIFF
--- a/.github/actions/setup-python/action.yml
+++ b/.github/actions/setup-python/action.yml
@@ -12,7 +12,6 @@ runs:
     - name: Install uv
       uses: astral-sh/setup-uv@v5
       with:
-        enable-cache: true
         python-version: ${{ matrix.python-version }}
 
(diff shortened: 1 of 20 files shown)


ahmetmeleq:
Further context for posterity

`SharepointConnectionConfig` inherits from `OnedriveConnectionConfig` , which is where the use and declaration of this arg is made. With the changes in the PR, we override the arg to be Optional

ahmetmeleq (approved):

PR #510, opened 2025-05-13
pawel/fix google drive download links

pawel-kmiecik:
According to Google API documentation, the `webContentLink` and `exportLink` are intended to be used in browsers, not by scripts.
This leads to a situation when e.g. `webContentLink` redirects to the Google'a auth login page, which is downloaded and sent to partition.

Instead of that we should use the `googleclient`'s methods, that [call the Google Drive appropriate APIs to perform download/export operations](https://developers.google.com/workspace/drive/api/guides/manage-downloads#python):
- `get_media` to download standalone files
- `export` to export Google Workspace native files (Google Docs, Google Slides, Google Sheets) to corresponding office files (docx, pptx, xlsx, accordingly)
- `download` to export Google Workspace native files for files that result with >10MB size
  - this operation uses LRO (Long Running Operation) mechanism described [here](https://developers.google.com/workspace/drive/api/guides/long-running-operations)


unstructured_ingest/processes/connectors/google_drive.py
```diff
@@ -501,73 +506,227 @@ class GoogleDriveDownloader(Downloader):
...
+                and e.resp.status not in [403, 429]
+            )
+
+        @backoff.on_exception(
+            backoff.expo,
+            (HttpError, OperationNotFinished),
+            max_tries=max_tries,
+            max_time=max_time,
+            giveup=is_fatal_code,
+        )
```
mpolomdeepsense, line 591: Can we use `tenacity` instead? I've been using it in IBM connector. I don't have anything against this lib, I'm just thinking it will be better to stick to one solution.

unstructured_ingest/processes/connectors/google_drive.py
```diff
@@ -501,73 +507,229 @@ class GoogleDriveDownloader(Downloader):
...
+                        status, done = downloader.next_chunk()
+                        logger.debug(f"Download progress: {int(status.progress() * 100)}.")
+
+            except HttpError as error:
+                if error.resp.status == 403 and "too large" in error.reason:
+                    # If we get a 403, try using LRO for large files
+                    logger.debug("File too large for direct export, using LRO")
+                    self._export_gdrive_file_with_lro(file_id, download_path, mime_type)
+                else:
+                    raise SourceConnectionError(f"Failed to export file: {error}")
```
mpolomdeepsense, line 681: I would need some explanation of this part of the code. Specifically on when the 403 error is thrown because that's not clear to me.

Is it thrown right away, when we start downloading the file, or after it reaches 10MB limit (i.e. it already downloaded 10MB but there is more data to be downloaded and it throws an error) ?

If it's the second case, then is it possible to check the file size before download and fallback to LRO in case the file is too large?

unstructured_ingest/processes/connectors/google_drive.py
```diff
@@ -501,73 +507,229 @@ class GoogleDriveDownloader(Downloader):
...
+                request = client.get_media(fileId=file_id)
+
+                with open(download_path, "wb") as file:
+                    downloader = MediaIoBaseDownload(file, request)
+                    done = False
+                    while done is False:
+                        status, done = downloader.next_chunk()
+                        logger.debug(f"Download progress:{int(status.progress() * 100)}.")
+
+        except (HttpError, ValueError) as error:
```
bryan-unstructured, line 538: I see we have the lro download method for large native google drive file types, what about non-native files? It's common for a user to store large pdf, jpg files in google drive. For now we only have MediaIoBaseDownload approach for non-native files, no matter what their sizes are.
But if this is a complex task, maybe we can add the function in later PR.

unstructured_ingest/processes/connectors/google_drive.py
bryan-unstructured: +1 on checking file size in advance instead of trail and error

bryan-unstructured:
now that there's logic to handle large native files, maybe we can add one more test case in integration test, to process a file with size 15mb? it shouldn't slow down CI too much

unstructured_ingest/processes/connectors/google_drive.py
pawel-kmiecik: Only the export operation (from Google Workspace native files to e.g. docx, pptx) has this limitation of 10MB. Standard files (like pdfs, etc) can be downloaded with `get_media` operation - it's like blob file download so doesn't use much Google resources (in contradiction to export - it requires some CPU load to convert from one format to another - that's the reason the API call is limited to avoid timeouts, connection issues on the client side)

`MediaIoBaseDownload` just fetches the file with buffering - exactly what `_raw_download_google_drive_file` method does (had to implement it my own as the LRO API is incompatible with `MediaIoBaseDownload`)

I thought it's better to use official client methods (taken from examples like [here](https://developers.google.com/workspace/drive/api/guides/manage-downloads#download-content) than implementing such things myself - and only do exceptions it if the official API doesn't support something.

unstructured_ingest/processes/connectors/google_drive.py
pawel-kmiecik: AFAIK it's impossible to check the file size of exported file in advance. I didn't check this endpoint exact execution time but it raises this error rather quickly - it looks the API estimates the exported file size and if it's too large - responds with 403.

But it's a good idea to double check that - I'll measure the time this API returns with 403 and if it's not immediate - I'll look to some metadata info about estimated export filesize and do LRO early if that's available.

pawel-kmiecik: Ok - I checked how long it takes to error if export is too large - it was ~5-10 secs (I guess depending on the source Google Workspace native file size) so you guys were right!

As I didn't find an "estimated exported" file size in metadata I introduced a threshold on source file size at 9MB (the exported sizes are slightly bigger from what I've observed).

I left this fallback mechanism for cases where e.g. 8MB native files exports to 11MB and errors out here.

unstructured_ingest/processes/connectors/google_drive.py
bryan-unstructured: Ah I see, there's a deeper reason behind this architecture

bryan-unstructured:
> now that there's logic to handle large native files, maybe we can add one more test case in integration test, to process a file with size 15mb? it shouldn't slow down CI too much

@pawel-kmiecik what about this?

pawel-kmiecik:
> > now that there's logic to handle large native files, maybe we can add one more test case in integration test, to process a file with size 15mb? it shouldn't slow down CI too much
> 
> @pawel-kmiecik what about this?

Oh, sure! I hope I just need to add a large native file somewhere :)

bryan-unstructured (approved):

PR #522, opened 2025-05-29
fix: change default to pinecone serverless_region default

ahmetmeleq:
>  I'm using the "opinionated writes" feature for the Pinecone destination connector and getting this error: Your free plan does not support indexes in the us-west-2 region of aws. To create indexes in this region, upgrade your plan. But I don't want to upgrade my free plan. Is there any way to force this index to be created in us-east-1 instead? 

Per [the docs on Regions available for serverless indexes](https://docs.pinecone.io/troubleshooting/available-cloud-regions#regions-available-for-serverless-indexes), we need to change our default to a starter plan friendly value. This PR addresses that


unstructured_ingest/processes/connectors/pinecone.py
```diff
@@ -240,7 +240,7 @@ def create_destination(
         destination_name: str = "unstructuredautocreated",
         destination_type: Literal["pod", "serverless"] = "serverless",
         serverless_cloud: str = "aws",
-        serverless_region: str = "us-west-2",
+        serverless_region: str = "us-east-1",
```
Paul-Cornell, line 243: OK, but shouldn't the caller be able to change this, in both the UI and with the API (Platform and Ingest)? Otherwise, what about callers who actually want to use something other than `us-east-1` instead? Then this code will similarly fail.

ahmetmeleq: That'd be a great, but would require changes in UI and API to support. We don't have interfaces / designs to include settings on opinionated writes right now. When we implement those, we'd also want to make sure we don't confuse the users on what's a destination connector setting (to connect) versus what's an opinionated writes setting (to create)

Until that gets prioritized, this would quickly introduce support for starter plans

shreyanid (approved):

Paul-Cornell (approved):

PR #523, opened 2025-05-30
feat: check all user specified spaces for confluence source

ryannikolaidis:
Improve `precheck` method of Confluence's Indexer.

Validate that each space provided in configuration can be accessed, raise exception if at least one of them can't.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
-## 1.0.33-dev
+## 1.0.34-dev
 
 * **Improve Confluence Indexer's precheck** - validate access to each space
+
+## 1.0.33
+
+* **Fix google drive not setting the display_name property on the FileData object**
 
 ## 1.0.32
 
--- a/unstructured_ingest/__version__.py
+++ b/unstructured_ingest/__version__.py
@@ -1,1 +1,1 @@
-__version__ = "1.0.33-dev"  # pragma: no cover
+__version__ = "1.0.34-dev"  # pragma: no cover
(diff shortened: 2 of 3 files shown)


test/unit/connectors/test_confluence.py
```diff
@@ -69,3 +84,80 @@ def test_connection_config_pat_auth():
...
+        type(connection_config).get_client.return_value.__enter__.return_value = mock_client
+
+        result = indexer.precheck()
+        mock_client.get_all_spaces.assert_called_once_with(limit=1)
+        mock_client.get_space.assert_not_called()
+        assert result is True
+
+
+def test_precheck_with_spaces_raises(monkeypatch, connection_config):
+    """Test that precheck raises SourceConnectionError if get_space fails."""
```
pawel-kmiecik: I see docstring mentions `SourceConnectionError` while the `precheck` raises `UserError` - is this just a typo-like mistake or something is wrong with the exception hierarchy?

unstructured_ingest/processes/connectors/confluence.py
```diff
@@ -126,15 +127,36 @@ class ConfluenceIndexer(Indexer):
...
+
+            # If specific spaces are provided, check if we can access them
+            errors = []
+
+            if self.index_config.spaces:
+                for space_key in self.index_config.spaces:
+                    try:
+                        client.get_space(space_key)
+                    except Exception as e:
+                        logger.error(f"Failed to connect to Confluence: {e}", exc_info=True)
```
pawel-kmiecik: Can't we use `logger.exception` instead of passing `exc_info` to the `error` method?

test/unit/connectors/test_confluence.py
ds-filipknefel: Looks like a type, the `UserError` is meant to be raised in prechecks

PR #528, opened 2025-06-04
fix output path in fsspec connector

bryan-unstructured:
the output path in destination connector should use relative path, instead of full path. If relative path is not available, use filename instead.
The errors in CI are not related to this change. They were from prior time.


potter-potter (approved):
works as expected

PR #529, opened 2025-06-05
RIVET-795: add library option to sharepoint

potter-potter:
Thanks to @Anjum48 we have a nice addition to the Sharepoint connector.

We can now download from Libraries other than the built in Documents(Shared Documents)


potter-potter:
> looks good to me. Just wonder why _get_drive_item function is in connection config instead of indexer. that function returns DriveItem object, is it that the connection config usually returns client or drive object, indexer or downloader handles how to get DriveItem?

Totally. Was hesitant myself. But we need it in both the Indexer and the Downloader. And connection_config is the only thing imported in both so I figured it could be put there. Open to suggestions. I'm a bit out of touch with connector code.

bryan-unstructured (approved):

PR #530, opened 2025-06-05
feat: notion syncblock fix by teddysupercuts

mateuszkuprowski:
This PR is @teddysupercuts contribution from his fork. 
Original PR can be found [here.](https://github.com/Unstructured-IO/unstructured-ingest/pull/508/)

Re-opening this PR from origin so that our tests work correctly.

### DIFF
--- a/unstructured_ingest/processes/connectors/notion/types/blocks/synced_block.py
+++ b/unstructured_ingest/processes/connectors/notion/types/blocks/synced_block.py
@@ -18,7 +18,10 @@ class OriginalSyncedBlock(BlockBase):
 
     @classmethod
     def from_dict(cls, data: dict):
-        # Original blocks contain children content
+        """Create OriginalSyncedBlock from dictionary data.
+        
+        Original blocks contain children content.
+        """
         if "children" not in data:
              raise ValueError(f"OriginalSyncedBlock data missing 'children': {data}")
         return cls(children=data["children"])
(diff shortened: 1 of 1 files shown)


unstructured_ingest/processes/connectors/notion/types/blocks/synced_block.py
```diff
@@ -31,27 +34,56 @@ class DuplicateSyncedBlock(BlockBase):
 
     @staticmethod
     def can_have_children() -> bool:
+        # Duplicate blocks themselves don't have children directly fetched here,
```
pawel-kmiecik: I think comments like this can go to the docstring (this apply to other similar comments in methods here)

unstructured_ingest/processes/connectors/notion/types/blocks/synced_block.py
```diff
@@ -31,27 +34,56 @@ class DuplicateSyncedBlock(BlockBase):
...
+            # If this assumption is wrong, errors might occur later.
+            # Consider logging a warning here if strictness is needed.
+            return OriginalSyncedBlock(children=[])
+
 
     def get_html(self) -> Optional[HtmlTag]:
+        # The specific instance returned by from_dict (Original or Duplicate)
+        # will handle its own get_html logic.
+        # This method on the base SyncBlock might not be directly called.
         return None
```
pawel-kmiecik, line 110: If this shouldn't be called in this class and it cannot be an ABC class then this method should throw `NotImplementedError`. WDYT?
BTW: 
> The specific instance returned by from_dict (Original or Duplicate)
> will handle its own get_html logic.

I see they all return `None`

unstructured_ingest/processes/connectors/notion/types/blocks/synced_block.py
mateuszkuprowski: Transferred comments into docstrings

unstructured_ingest/processes/connectors/notion/types/blocks/synced_block.py
mateuszkuprowski: I've taken a deeper look into this and found a few things. 
1. Return None behaviour is consistent with other blocks that we handle
2. This is actually expected and handled in the helpers.py:
` if current_block["block"].get_html():
       html.append(current_block["block"].get_html())
   else:
       html.append([])`
       
   WDYT?

pawel-kmiecik: Weird but ok - let's keep consistency.

pawel-kmiecik (approved):
LGTM

PR #532, opened 2025-06-06
feat: add display name to connectors

pawel-kmiecik:
# Add `display_name` attribute for `FileData` in connectors that miss it
Display name might be helpful when debugging failed processing.
This PR:
- adds this attribute to the connectors that miss it
- updates tests to check for that field
- udpates fixtures/expected processed data files (=fixtures update) - __MOST OF THE CHANGES IN THIS PR__


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.39
+
+* **Add `display_name` to FileData in 14 connectors**
```
mateuszkuprowski, line 3: You;ll need to bump it, I just deployed 1.0.40

mateuszkuprowski (approved):
Honestly the changes are rather simple, and the tests work correctly with new fixtures. Must have taken a long time to check all these, but all this looks good to me.

PR #533, opened 2025-06-06
feat: added metadata storage in dynamic fields for Milvus

mateuszkuprowski:
Added tests for dynamic field usage and metadata storage in Milvus connector.
Now all the metadata in files should go into dynamic filed in a collection if it's available. If it's not we omit saving it. 
Previously it wasn't saved at all.

### DIFF
--- a/test/integration/connectors/test_milvus.py
+++ b/test/integration/connectors/test_milvus.py
@@ -177,8 +177,7 @@ async def test_milvus_destination(
     upload_file_with_embeddings = add_fake_embeddings(upload_file, tmp_path)
     file_data = FileData(
         source_identifiers=SourceIdentifiers(
-            fullpath=upload_file_with_embeddings.name,
-            filename=upload_file_with_embeddings.name,
+            fullpath=str(upload_file), filename=upload_file.stem
         ),
         connector_type=CONNECTOR_TYPE,
         identifier="mock file data",
@@ -221,8 +220,7 @@ async def test_milvus_metadata_storage_with_dynamic_fields(
     upload_file_with_embeddings = add_fake_embeddings(upload_file, tmp_path)
     file_data = FileData(
         source_identifiers=SourceIdentifiers(
-            fullpath=upload_file_with_embeddings.name,
-            filename=upload_file_with_embeddings.name,
+            fullpath=str(upload_file), filename=upload_file.stem
         ),
         connector_type=CONNECTOR_TYPE,
         identifier="metadata_test_file",
@@ -307,8 +305,7 @@ async def test_milvus_metadata_filtering_without_dynamic_fields(
     upload_file_with_embeddings = add_fake_embeddings(upload_file, tmp_path)
     file_data = FileData(
         source_identifiers=SourceIdentifiers(
-            fullpath=upload_file_with_embeddings.name,
-            filename=upload_file_with_embeddings.name,
+            fullpath=str(upload_file), filename=upload_file.stem
         ),
         connector_type=CONNECTOR_TYPE,
         identifier="no_dynamic_test_file",
--- a/test/integration/connectors/test_milvus.py
+++ b/test/integration/connectors/test_milvus.py
@@ -115,7 +115,8 @@ def collection():
                 index_params=index_params,
             )
             print(
-                f"Created collection {COLLECTION_WITHOUT_DYNAMIC_FIELDS}: {collection_resp_no_dynamic}"
+                f"Created collection {COLLECTION_WITHOUT_DYNAMIC_FIELDS}: \
+                {collection_resp_no_dynamic}"
             )
 
             yield EXISTENT_COLLECTION_NAME
@@ -236,7 +237,8 @@ async def test_milvus_metadata_storage_with_dynamic_fields(
 
     assert (
         len(metadata_found) > 0
-    ), f"Expected to find metadata fields in staged data. Available keys: {list(sample_element.keys())}"
+    ), f"Expected to find metadata fields in staged data.\
+   Available keys: {list(sample_element.keys())}"
     print(f"Found metadata fields: {metadata_found}")
 
     uploader.run(path=staged_filepath, file_data=file_data)
@@ -261,7 +263,8 @@ async def test_milvus_metadata_storage_with_dynamic_fields(
 
         assert (
             len(stored_metadata) > 0
-        ), f"Expected metadata fields to be stored in Milvus. Available fields: {list(sample_result.keys())}"
+        ), f"Expected metadata fields to be stored in Milvus. \
+        Available fields: {list(sample_result.keys())}"
         print(f"Successfully stored metadata fields: {stored_metadata}")
 
         # Verify filename is specifically stored if present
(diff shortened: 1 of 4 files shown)


unstructured_ingest/processes/connectors/milvus.py
```diff
@@ -97,10 +126,46 @@ def parse_date_string(date_string: str) -> float:
...
+                }
+            )
+
+        # Flatten the extracted metadata
+        flattened_metadata = flatten_dict(
+            extracted_metadata,
+            separator="_",
+            flatten_lists=True,
+            remove_none=True,
+        )
```
mpolomdeepsense: Wouldn't it be safer to use `working_data` instead of `element_dict`? If a user were to add `metadata` to `metadata_fields` then after flattening it could result in duplicated values (under different keys, but still, values would be duplicated)

unstructured_ingest/processes/connectors/milvus.py
```diff
@@ -97,10 +126,46 @@ def parse_date_string(date_string: str) -> float:
...
+                flatten_dict(
+                    metadata,
+                    keys_to_omit=["data_source_record_locator"],
+                    remove_none=True,
+                )
+            )
+
+        # Add the extracted metadata as separate fields
+        # (will be stored in dynamic fields if supported)
+        working_data.update(flattened_metadata)
```
mpolomdeepsense: And here we also flatten all original `metadata`, next, add it to `working_data`, and later, again, add previously `flattened_metadata` to `working_data`. Should it work this way?

unstructured_ingest/processes/connectors/milvus.py
```diff
@@ -200,14 +306,61 @@ def insert_results(self, data: Union[dict, list[dict]]):
...
+            }
+            # Remove metadata fields that are not part of the base schema
+            filtered_data = []
+            for item in data:
+                filtered_item = {
+                    key: value for key, value in item.items() if key in schema_fields
+                }
+                filtered_data.append(filtered_item)
+            data = filtered_data
+
```
mpolomdeepsense: I think this could be moved to a separate function that would only be responsible for modifying data to match the schema.

test/integration/connectors/test_milvus.py
```diff
@@ -133,25 +230,256 @@ async def test_milvus_destination(
...
+        logger.debug(json.dumps(get_printable_sample(staged_elements[0]), indent=2))
+
+    # Verify that metadata fields are present in staged data
+    sample_element = staged_elements[0]
+    metadata_found = []
+    for field in ALLOWED_METADATA_FIELDS:
+        if field in sample_element:
+            metadata_found.append(field)
+
+    assert (
```
pawel-kmiecik: A bit weird formatting, something like that isn't better?
```python
    assert len(metadata_found) > 0, (
        "Expected to find metadata fields in staged data."
        f"Available keys: {list(sample_element.keys())}"
   )
```

test/integration/connectors/test_milvus.py
```diff
@@ -133,25 +230,256 @@ async def test_milvus_destination(
...
+        assert len(results) > 0, "Should have results from the uploaded data"
+
+        # Check that metadata fields are present in the results
+        sample_result = results[0]
+        stored_metadata = []
+        for field in metadata_found:
+            if field in sample_result:
+                stored_metadata.append(field)
+
+        assert (
```
pawel-kmiecik: as above

unstructured_ingest/processes/connectors/milvus.py
```diff
@@ -79,6 +100,14 @@ class MilvusUploadStagerConfig(UploadStagerConfig):
     flatten_metadata: bool = True
     """If set - flatten "metadata" key and put contents directly into data"""
 
+    metadata_fields: list[str] = Field(
+        default=list(ALLOWED_METADATA_FIELDS),
```
pawel-kmiecik: I guess we need a factory here - shouldn't pass a list as a default argument.

unstructured_ingest/processes/connectors/milvus.py
```diff
@@ -154,6 +221,29 @@ class MilvusUploader(Uploader):
     upload_config: MilvusUploaderConfig
     connector_type: str = CONNECTOR_TYPE
 
+    def has_dynamic_fields_enabled(self) -> bool:
+        """Check if the target collection has dynamic fields enabled."""
+        try:
+            with self.get_client() as client:
+                # Get collection schema information
```
pawel-kmiecik: I think these comments are redundant - the code is rather straightforward

unstructured_ingest/processes/connectors/milvus.py
```diff
@@ -182,7 +286,9 @@ def delete_by_record_id(self, file_data: FileData) -> None:
             f"from milvus collection {self.upload_config.collection_name}"
         )
         with self.get_client() as client:
-            delete_filter = f'{self.upload_config.record_id_key} == "{file_data.identifier}"'
+            delete_filter = (
```
pawel-kmiecik, line 217: it's a rather long and complex method right now - can we divide it into more smaller ones?

pawel-kmiecik (commented):
And general comment - I'm not sure if we need so many logger calls in integration tests - the code gets blurred. I know they might help debugging but they can be added if needed. At the same time I don't have a strong opinion here :)

unstructured_ingest/processes/connectors/milvus.py
mateuszkuprowski: You are 100% right, this mess was due to chaning this funciton many times, udpate to much more simple and concise version

unstructured_ingest/processes/connectors/milvus.py
mateuszkuprowski: Same as above I've simplified entire funciton and cleaned repeated methods, mess after changing function many times and then forgetting about it

unstructured_ingest/processes/connectors/milvus.py
mateuszkuprowski: You're right especially considering we have so much debug reports :)

unstructured_ingest/processes/connectors/milvus.py
mateuszkuprowski: Done major cleanup :)

unstructured_ingest/processes/connectors/milvus.py
mateuszkuprowski: done major cleanup of this function

unstructured_ingest/processes/connectors/milvus.py
mateuszkuprowski: Good idea, refactored into it's own method.

unstructured_ingest/processes/connectors/milvus.py
mateuszkuprowski: Done! Good point!

unstructured_ingest/processes/connectors/milvus.py
mateuszkuprowski: Cleaned it up a little bit. Left some logging I think is the most useful and removed other which was excessive and I used it for debugging when things were breaking.

unstructured_ingest/processes/connectors/milvus.py
mateuszkuprowski: You mean delete by record ID? It feels very concise. If you meant next function below it I've refactored it and you are right this was too large.

test/integration/connectors/test_milvus.py
mateuszkuprowski: Outdated already, done major refactor :)

test/integration/connectors/test_milvus.py
mateuszkuprowski: Taht was claude's idea, maybe I need better rules for this

mateuszkuprowski: Fixed it anyway

PR #535, opened 2025-06-10
fix: search only for valid attachment links in Confluence page

ds-filipknefel:
Modify the `HtmlMixin` to allow for customized hyperlink tag search in subclasses using the mixin by overwriting a dedicated method.
Overwrite said method in the `ConfluenceDownloaderConfig` to target only embedded attachment files when running with `extract_files` flag.


unstructured_ingest/utils/html.py
```diff
@@ -158,3 +159,15 @@ def extract_embedded_files(
             )
             for url_to_download in urls_to_download
         ]
+
+    @requires_dependencies(["bs4"])
+    def _find_hyperlink_tags(self, html_soup: "BeautifulSoup") -> list["Tag"]:
+        """Find hyperlink tags in the HTML.
+
+        Overwrite this method to customize the tag search.
```
ds-filipknefel, line 167: Alternatively I'm thinking this method could be extended instead of overwritten, such that in subclass you'd do the following - however this means you can't use a customized query which you might want to.
```py
tags = super()._find_hyperlink_tags(html_soup)
return [tag for tag in tags if subclass_predicate(tag)]
```

mpolomdeepsense (approved):
LGTM

PR #536, opened 2025-06-21
fix: confluence page limits

ryannikolaidis:
The limit argument supported by the confluence client is not limiting the pages as advertised. This adds a limit to the returned results. 

Integration test was updated to a parameterized pattern to reduce duplication and include an additional test that validates the doc limit.

> [!NOTE]
> the current CI failures are unrelated to this PR and will be resolved by: https://github.com/Unstructured-IO/unstructured-ingest/pull/538

### DIFF
--- a/unstructured_ingest/processes/connectors/confluence.py
+++ b/unstructured_ingest/processes/connectors/confluence.py
@@ -186,7 +186,6 @@ class ConfluenceIndexer(Indexer):
             pages = client.get_all_pages_from_space(
                 space=space_key,
                 start=0,
-                limit=self.index_config.max_num_of_docs_from_each_space,
                 expand=None,
                 content_type="page",  # blogpost and comment types not currently supported
                 status=None,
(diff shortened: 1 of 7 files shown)


unstructured_ingest/processes/connectors/confluence.py
```diff
@@ -191,7 +191,11 @@ def _get_docs_ids_within_one_space(self, space_key: str) -> List[dict]:
                 content_type="page",  # blogpost and comment types not currently supported
                 status=None,
             )
-        doc_ids = [{"space_id": space_key, "doc_id": page["id"]} for page in pages]
+        # Limit the number of documents to max_num_of_docs_from_each_space
+        # Note: this is needed because the limit field in client.get_all_pages_from_space does 
+        # not seem to work as expected
+        limited_pages = pages[: self.index_config.max_num_of_docs_from_each_space]
```
potter-potter, line 196: if we are doing this here can we remove self.index_config.max_num_of_docs_from_each_space from line 189. or is it required?

PR #538, opened 2025-06-21
fix: bump together ai model used in testing

ryannikolaidis:
we were using m2-bert-80M-8k-retrieval which is no longer supported. bumped to: m2-bert-80M-32k-retrieval


potter-potter (approved):
LG

PR #543, opened 2025-06-26
feat(delta-table): add record-level upsert

amadeusz-ds:
- `DeltaTableUploadStager`
  - injects `RECORD_ID_LABEL` on every row, enabling per-record tracking.

- `DeltaTableUploader.upload_dataframe`
  - deletes rows whose `record_id` matches the current file before appending new data instead of doing a full table overwrite.
  - handles concurrent writers by wrapping the delete-then-append sequence in a tenacity-based retry (`10` attempts, random 0.2-1 s back-off) that re-runs only on commit-conflict errors (“CommitFailed”, “Metadata changed”).
  - keeps the SIGABRT-work-around: runs the writer in a subprocess unless executing inside a daemon worker; error propagation unified through a single `multiprocessing.Queue`.

- refactor
  - removed duplicated queue/error-handling blocks, preserved detailed explanatory comments.
  - added `@requires_dependencies(["tenacity"], extras="delta-table")` to enforce the new optional dependency.

### DIFF
--- a/unstructured_ingest/processes/connectors/delta_table.py
+++ b/unstructured_ingest/processes/connectors/delta_table.py
@@ -1,7 +1,7 @@
-import os
+import logging
 import traceback
 from dataclasses import dataclass, field
-from multiprocessing import Process, Queue
+from multiprocessing import Process, Queue, current_process
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Optional
 from urllib.parse import urlparse
@@ -20,6 +20,7 @@ from unstructured_ingest.interfaces import (
 )
 from unstructured_ingest.logger import logger
 from unstructured_ingest.processes.connector_registry import DestinationRegistryEntry
+from unstructured_ingest.utils.constants import RECORD_ID_LABEL
 from unstructured_ingest.utils.data_prep import get_data_df, get_json_data
 from unstructured_ingest.utils.dep_check import requires_dependencies
 from unstructured_ingest.utils.table import convert_to_pandas_dataframe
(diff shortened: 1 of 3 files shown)


unstructured_ingest/processes/connectors/delta_table.py
```diff
@@ -138,41 +141,91 @@ def precheck(self):
...
+            from deltalake import DeltaTable  # pylint: disable=import-error
+
+            dt = DeltaTable(upload_path, storage_options=storage_options)
+            # Table exists – remove any previous rows for this record_id so that appending is
+            # effectively an idempotent overwrite for the record.
+            dt.delete(predicate=f"{RECORD_ID_LABEL} = '{file_data.identifier}'")
+            mode = "append"
+        except Exception:
+            # Table does not exist yet (or cannot be opened) – we will create it below with
+            # mode="overwrite". All other failures will be captured later by the writer.
```
ds-filipknefel, line 169: Can we log here? Given how wide the Exception is if in the future we encounter some issue with this mechanism we'll likely benefit from seeing what caused us to chose "overwrite"

unstructured_ingest/processes/connectors/delta_table.py
```diff
@@ -138,41 +141,91 @@ def precheck(self):
...
+            """One optimistic transaction: delete old rows, then append new ones."""
+
+            # NOTE: deltalake writer on Linux sometimes can finish but still trigger a SIGABRT and
+            # cause ingest to fail, even though all tasks are completed normally. Putting the writer
+            # into a process mitigates this issue by ensuring python interpreter waits properly for
+            # deltalake's rust backend to finish
+            queue: Queue[str] = Queue()
+
+            if current_process().daemon:
+                # write_deltalake_with_error_handling will push any traceback to our queue
```
ds-filipknefel, line 212: Can you provide more context on this?

amadeusz-ds: Roman introduced a queue but it doesn't work within a daemon process which is being spawn for this class.

unstructured_ingest/processes/connectors/delta_table.py
amadeusz-ds: Done

PR #547, opened 2025-06-30
fix(delta-table): normalize S3 LocationConstraint values

amadeusz-ds:
The S3 GetBucketLocation operation returns `null` (`None`) for buckets in the legacy `us-east-1` region and `EU` for very old buckets that were created in the historical `EU` region (now `eu-west-1`). For every other region the API already returns the correct AWS region string.
Add a function that normalizes bucket location.

### DIFF
--- a/unstructured_ingest/processes/connectors/delta_table.py
+++ b/unstructured_ingest/processes/connectors/delta_table.py
@@ -280,10 +280,10 @@ class DeltaTableUploader(Uploader):
 def _normalize_location_constraint(location: Optional[str]) -> str:
     """Return canonical AWS region name for a LocationConstraint value.
 
-    The S3 GetBucketLocation operation returns *null* (→ ``None``) for buckets in
-    the legacy **us-east-1** region and ``"EU"`` for very old buckets that were
-    created in the historical **EU** region (now **eu-west-1**).  For every other
-    region the API already returns the correct AWS region string.  This helper
+    The S3 GetBucketLocation operation returns `null` (`None`) for buckets in
+    the legacy `us-east-1` region and `EU` for very old buckets that were
+    created in the historical `EU` region (now `eu-west-1`). For every other
+    region the API already returns the correct AWS region string. This helper
     normalises the legacy values so callers can reliably compare regions.
 
     Args:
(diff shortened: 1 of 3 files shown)


unstructured_ingest/processes/connectors/delta_table.py
```diff
@@ -270,6 +277,29 @@ def run(self, path: Path, file_data: FileData, **kwargs: Any) -> None:  # type:
...
+    Args:
+        location: The LocationConstraint value returned by the S3 GetBucketLocation operation.
+
+    Returns:
+        The canonical AWS region name for the given location constraint.
+    """
+
+    if location is None:
+        return "us-east-1"
+    if location == "EU":
```
mpolomdeepsense, line 298: Should we handle for both upper and lower case?

```suggestion
    if location.lower() == "eu":
```

unstructured_ingest/processes/connectors/delta_table.py
```diff
@@ -134,8 +134,15 @@ def precheck(self):
...
-                    raise ValueError("Wrong AWS Region was provided.")
+                bucket_region = _normalize_location_constraint(response.get("LocationConstraint"))
+
+                if self.connection_config.aws_region != bucket_region:
+                    raise ValueError(
+                        "Wrong AWS region provided: bucket "
+                        f"'{bucket_name}' resides in '{bucket_region}', "
+                        "but configuration specifies "
+                        f"'{self.connection_config.aws_region}'."
+                    )
```
mpolomdeepsense, line 145: Would it be possible here to just pick one of the regions? It might work later, and we would only throw this error when it fails during data upload.

unstructured_ingest/processes/connectors/delta_table.py
amadeusz-ds: It returns uppercase.

unstructured_ingest/processes/connectors/delta_table.py
amadeusz-ds: It fails silently - you don't realize until you check the bucket.

mpolomdeepsense: I see. Resolving

PR #552, opened 2025-07-09
fix: resolve UnboundLocalError in SharePoint connector

CyMule:


potter-potter:
can we get some description above and a link to the ticket so I can see the background for this fix?

unstructured_ingest/processes/connectors/sharepoint.py
```diff
@@ -150,6 +150,7 @@ def _fetch_file(self, file_data: FileData) -> DriveItem:
             site_drive_item = self.connection_config._get_drive_item(client_site)
         except ClientRequestException:
             logger.info("Site not found")
+            raise SourceConnectionError(f"Site not found: {self.connection_config.site}")
```
raiderrobert, line 153: So just eyeballing this. It does catch and raise this fatal bug. But I'm wondering how we got here in the first place?

I would expect the `precheck` to catch this, but I look above and I see no precheck on things related to this being Sharepoint.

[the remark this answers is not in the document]
potter-potter: bro. yes it is. pull your head out.

potter-potter:
> can we get some description above and a link to the ticket so I can see the background for this fix?

aah. i see its from the slack thread.

unstructured_ingest/processes/connectors/sharepoint.py
potter-potter: Totally. At one point i was hoping to override the precheck we inherit from Onedrive.

potter-potter: Should we be using our new error "taxonomy"

https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/errors_v2.py

raiderrobert: @potter-potter yeah, that's a good call out.

PastelStorm: I would encourage people not to use an external dependency just to raise an error. Our external "taxonomy" is yet another layer of abstraction that sits on top of another layer in our FastAPI wrapper. This is not a good architecture.

potter-potter: Can you suggest a better solution? Also why is it an external dependency if it is in the same repo?

PR #557, opened 2025-07-14
pawel/improve connectors logging

pawel-kmiecik:


unstructured_ingest/processes/utils/connector_logging.py
```diff
@@ -0,0 +1,662 @@
...
+        self.log_timings = log_timings
+        self.timing_threshold_seconds = timing_threshold_seconds
+
+
+class OperationTimer:
+    """Utility class for tracking operation timing."""
+
+    def __init__(self):
+        self.start_times: Dict[str, float] = {}
+        self.durations: Dict[str, list[float]] = {}
```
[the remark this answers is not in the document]
potter-potter: good point. this repo is still 3.9+

potter-potter:
@claude please review

pawel-kmiecik:
> This is really cool.
> 
> I didn't see the Linear ticket for this.
> 
> For more clarity and understanding where did you find the logging lacking?

I just didn't see much "debug" logging in the pipeline and thought they would be nice to have with some time performance statistics, but:

I actually configured everything correctly and now **it looks this change is not necessary**...

Here we have an OTEL configuration that logs operations for each span:
https://github.com/Unstructured-IO/unstructured-ingest/blob/main/unstructured_ingest/otel.py#L76
I just had to configure logging on run on the `debug` level and now the operations logged from the Mixin class seem "redundant" as we have the same from the OTEL LogExporterProcessor.

The operations look like:
```
2025-07-17 13:50:13,493 SpawnPoolWorker-3 DEBUG    chunk finished in 0.005469s, attributes: file_id=272b42cbf30c
```
and are logged automatically. The timings logged are taken directly from spans in that approach.

The mechanism is not visible at the first look - this is the reason I started working on the LoggingMixin class / feature.

IMO we should stay with OTEL (somewhat this might look controversial as the mechanism is designed to log spans generally and not just operations with status - but it's less modifications that need to be added) and optionally import sanitizing functionality and review the spans to be complete.

Chmielok:
Timing related mechanisms have been removed from this PR, as they are handled by OTEL automatically.

unstructured_ingest/processes/connectors/fsspec/s3.py
```diff
@@ -104,7 +104,13 @@ def wrap_error(self, e: Exception) -> Exception:
                 return UserError(message)
             if http_code >= 500:
                 return ProviderError(message)
-        logger.error(f"unhandled exception from s3 ({type(e)}): {e}", exc_info=True)
+        logger.error(
```
bryan-unstructured, line 107: why only fsspec and s3 connectors' logger are updated, what about sql connectors, vector db connectors, etc?

Chmielok: These will be done in other PRs

unstructured_ingest/processes/connectors/fsspec/fsspec.py
```diff
@@ -353,14 +424,31 @@ def get_upload_path(self, file_data: FileData) -> Path:
     def run(self, path: Path, file_data: FileData, **kwargs: Any) -> None:
         path_str = str(path.resolve())
         upload_path = self.get_upload_path(file_data=file_data)
-        logger.debug(f"writing local file {path_str} to {upload_path}")
-        with self.connection_config.get_client(protocol=self.upload_config.protocol) as client:
-            client.upload(lpath=path_str, rpath=upload_path.as_posix())
+        self.log_upload_start(file_path=path_str, destination=upload_path.as_posix())
```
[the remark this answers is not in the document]
potter-potter: All this will be cleaned up with the linter. `make tidy` and `make check`

potter-potter (approved):
looks good. just make sure you lint `make check` `make tidy`

unstructured_ingest/processes/utils/logging/connector.py
```diff
@@ -0,0 +1,393 @@
...
+            all_context.update(kwargs)
+
+        if all_context:
+            if self._should_sanitize():
+                sanitized_context = self._sanitizer.sanitize_dict(all_context)
+                logger.debug("%s context: %s", log_type, sanitized_context)
+            else:
+                logger.debug("%s context: %s", log_type, all_context)
+
+    def log_warning_with_context(self, message: str, context: Optional[Dict[str, Any]] = None):
```
bryan-unstructured: what's the difference between this function and log_warning? it seems log_warning_with_context covers a portion of log_warning's functionality, thus can be replaced by log_warning

unstructured_ingest/processes/utils/logging/sanitizer.py
```diff
@@ -0,0 +1,113 @@
...
+        path_obj = Path(path_str)
+
+        if len(path_obj.parts) > 2:
+            return f".../{path_obj.parent.name}/{path_obj.name}"
+        return path_obj.name
+
+    @staticmethod
+    def sanitize_id(identifier: str) -> str:
+        """Sanitize IDs for logging, showing only first/last few characters."""
+        if not identifier or len(identifier) < 8:
```
bryan-unstructured: it's sad I don't get to see any part of the ID in logging if it's shorter than 8. It would be helpful if it could even show first 4 characters.

unstructured_ingest/processes/utils/logging/sanitizer.py
```diff
@@ -0,0 +1,113 @@
...
+            return f"{parsed.scheme}://{parsed.netloc}{parsed.path}"
+        except (ValueError, TypeError):
+            return "<url>"
+
+    @staticmethod
+    def sanitize_token(token: str) -> str:
+        """Sanitize tokens and secrets for logging."""
+        if not token:
+            return "<token>"
+        if len(token) < 8:
```
bryan-unstructured: same comment as above in sanitize_id

unstructured_ingest/processes/utils/logging/sanitizer.py
```diff
@@ -0,0 +1,113 @@
...
+                "refresh_token",
+                "client_secret",
+                "private_key",
+                "credentials",
+            }
+
+        sanitized = {}
+        for k, v in data.items():
+            key_lower = k.lower()
+            if any(sensitive_key in key_lower for sensitive_key in sensitive_keys):
```
bryan-unstructured, line 99: this if statement would return True if `key_lower` is something like "key_counter" or "if_use_credentials", in these cases we probably don't need to sanitize the values.
But I cannot come up with a better solution yet.

unstructured_ingest/processes/utils/logging/connector.py
Chmielok: Good catch!

unstructured_ingest/processes/utils/logging/sanitizer.py
Chmielok: Fixed

unstructured_ingest/processes/utils/logging/sanitizer.py
Chmielok: Fixed

unstructured_ingest/processes/utils/logging/sanitizer.py
Chmielok: Sanitization has to be thought through more deeply.
It's included here as sort of inheritance from Paweł, but I would discuss the approach to it.

PR #558, opened 2025-07-18
o11y: Downgrade OTEL logs to `DEBUG` by default, make it configurable

Chmielok:
This PR can be thought of as a successor to closed #557, with smaller scope and removed duplicate logging (see discussion in #557).

The goal of this PR is primarily to downgrade OTEL logs to `DEBUG`, but at the same time allow for keeping the previous behavior through the use `OTEL_LOG_LEVEL` env var.

This should clean the logs that are currently flooded with time statistics.


unstructured_ingest/otel.py
```diff
@@ -31,13 +32,27 @@ def export(self, spans: Sequence[ReadableSpan]) -> SpanExportResult:
...
+        'ERROR': logging.ERROR,
+        'WARN': logging.WARNING,
+        'WARNING': logging.WARNING,
+        'INFO': logging.INFO,
+        'DEBUG': logging.DEBUG,
+        'NOTSET': logging.NOTSET,
+    }
+    log_level = os.getenv("OTEL_LOG_LEVEL", "DEBUG").upper()
+    log_level_int = level_names_mapping.get(log_level, logging.DEBUG)
+    return lambda message: logger.log(log_level_int, message)
```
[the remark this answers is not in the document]
Chmielok: There is no need for dynamic log level adjustment.

unstructured_ingest/otel.py
```diff
@@ -31,13 +32,27 @@ def export(self, spans: Sequence[ReadableSpan]) -> SpanExportResult:
...
 
 @dataclass
 class OtelHandler:
     otel_endpoint: Optional[str] = None
     service_name: str = "unstructured-ingest"
     trace_provider: TracerProvider = field(init=False)
-    log_out: Callable = field(default=logger.info)
+    log_out: Callable = field(default=get_log_out())
```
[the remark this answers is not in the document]
Chmielok: Not a problem as this should not change without restarting.

unstructured_ingest/otel.py
```diff
@@ -31,13 +32,27 @@ def export(self, spans: Sequence[ReadableSpan]) -> SpanExportResult:
...
+        'CRITICAL': logging.CRITICAL,
+        'FATAL': logging.FATAL,
+        'ERROR': logging.ERROR,
+        'WARN': logging.WARNING,
+        'WARNING': logging.WARNING,
+        'INFO': logging.INFO,
+        'DEBUG': logging.DEBUG,
+        'NOTSET': logging.NOTSET,
+    }
+    log_level = os.getenv("OTEL_LOG_LEVEL", "DEBUG").upper()
```
bryan-unstructured, line 46: if an env var is used to control log level, where is this env var set?
E.g, when a service imports unstructured-ingest, does it require that service to operate in an env where OTEL_LOG_LEVEL is set? If yes, it seems multiple repos need to be updated

Chmielok: > does it require that service to operate in an env where OTEL_LOG_LEVEL is set?

No, the user can manipulate this env var, but if they ignore/forget about it, it defaults to DEBUG.

potter-potter: we might want to default to INFO? At least in Platform we are trying not to fill our logs with debug.

Chmielok: That would be even worse as these logs would then flood INFO level.
There is a notion of TRACE level that's even lower than DEBUG, but unfortunately it's not supported by Python logging.

potter-potter: Yeah. Sorry. Of course. I was thinking wrong. Whoops. Was late night PR reviewing...

potter-potter: Debug is right. :)

potter-potter (approved):
LG

PR #559, opened 2025-07-21
feat: html mixin metadata improvements

ds-filipknefel:
Update the `display_name` metadata field for `FileData` produced by `HtmlMixin` to reflect relation between parent HTML file and downloaded file.

Format of the display name: `{parent-display-name} > {child-filename}`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.56-dev0
+
+- **Feat: improve HtmlMixin produced metadata**
+
 ## 1.0.55
 
 * **Fix: add precheck method to SharePoint connector**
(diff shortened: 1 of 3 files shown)


unstructured_ingest/utils/html.py
```diff
@@ -129,6 +129,7 @@ def get_download_response(
         )
         result_file_data = file_data.model_copy(deep=True)
         result_file_data.metadata.url = url
+        result_file_data.display_name = f"{result_file_data.display_name or ''} > {filename}"
```
ds-filipknefel: Upon consideration - relationship should be modeled elsewhere, the display name should be relevant to the file at hand.
```suggestion
        result_file_data.display_name = filename
```

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.56-dev0
+
+- **Feat: improve HtmlMixin produced metadata**
```
ds-filipknefel:
```suggestion
- **Fix: set correct display_name in HtmlMixin produced FileData**
```

PR #560, opened 2025-07-22
test: Longer interval for pinecone integration tests

Chmielok:
Short interval times result in pinecone tests not finishing properly.
By giving these tests more time to finish, they can properly pass.


potter-potter (approved):
LG

PR #564, opened 2025-07-24
Fix: Update examples to acknowledge actual folder structure; remove incorrect logging

Chmielok:
For some reason all examples point to folders that are too high in the hierarchy, making them always fail.
This PR changes them to point to actual example_data.

Another change is getting rid of logs that didn't even work - due to the current logger approach, they've lacked a handler and never logged anything.


potter-potter (approved):
looks good thank you.

PR #567, opened 2025-07-25
o11y: Reduce redundancies in logging code; split into multiple files

Chmielok:
This PR aims to split the current logging mixin into several smaller, specific ones.
It also reduces the redundancy of the code and makes more natural use of the parameters provided in the methods.


unstructured_ingest/processes/connectors/airtable.py
```diff
@@ -164,11 +164,11 @@ def get_table_metas(self) -> list[AirtableTableMeta]:
             return self.get_all_table_meta()
         return self.get_meta_from_list()
 
-    def precheck(self) -> None:
+    def _precheck(self) -> None:
```
bryan-unstructured, line 167: Note: this function is used in other services. It's gonna affect the code in other repos.

unstructured_ingest/processes/connectors/fsspec/fsspec.py
```diff
@@ -102,41 +102,25 @@ class FsspecIndexer(Indexer):
...
-            raise self.wrap_error(e=e)
+        files = fs.ls(path=self.index_config.path_without_protocol, detail=True)
+        valid_files = [x.get("name") for x in files if x.get("type") == "file"]
+        if not valid_files:
+            self.log_operation_complete("Connection validation", count=0)
+            return
+        file_to_sample = valid_files[0]
+        self.log_debug(f"attempting to make HEAD request for file: {file_to_sample}")
+        with self.connection_config.get_client(protocol=self.index_config.protocol) as client:
+            client.head(path=file_to_sample)
```
bryan-unstructured, line 123: connection validation is no longer needed here?

unstructured_ingest/processes/utils/logging/connectors/base.py
```diff
@@ -0,0 +1,112 @@
...
+        **kwargs,
+    ):
+        """Log an error message with optional exception, context and sanitization."""
+        if error:
+            logger.error(f"{message}: {error}", exc_info=True)
+        else:
+            logger.error(message)
+        self._log_context("Error", **kwargs)
+
+    def _log_context(self, log_type: str, **kwargs):
```
bryan-unstructured, line 65: DEBUG level logs are not presented in staging and production elastic dashboards, so nobody will see any of these context logs (at least in staging and prod). and from my understanding, this is not something user can tweak in the dashboards, which means even if I want to see DEBUG level logs for a job, it's impossible. So, do we still want to log them in DEUBG level?

unstructured_ingest/interfaces/indexer.py
```diff
@@ -15,15 +16,80 @@ class IndexerConfig(BaseModel):
...
+        except Exception as e:
+            self.log_connection_validation_failed(
+                connector_type=self.connector_type, error=e, endpoint=self.endpoint_to_log
+            )
+            raise self.wrap_error(e=e)
+        self.log_connection_validation_success(
+            connector_type=self.connector_type, endpoint=self.endpoint_to_log
+        )
+
+    def _precheck(self) -> None:
```
bryan-unstructured, line 91: what's the difference between precheck and _precheck?

unstructured_ingest/processes/connectors/fsspec/fsspec.py
Chmielok: I did not want to introduce changes to logic here, so I kept it as is. Note that failure in precheck is still handled as it would be before the changes i.e. an exception is generated and passed further.

unstructured_ingest/processes/utils/logging/connectors/base.py
Chmielok: That's a good point.
The best approach would be to use fully structured logging, but that would be an enormous change that we'd have to discuss.

bryan-unstructured: yeah let's discuss it and address it in future PRs if necessary

PR #568, opened 2025-07-29
fix(embedding): embedding batches overwriting one another

ds-filipknefel:
### Bug
If embedding happens in more than one batch the resulting embeddings are doubled embeddings from last batch. When embeddings are then assigned to elements it results in either missing or invalid embeddings.
```py
        embeddings = []
        try:
            for batch in batch_generator(texts, batch_size=self.config.batch_size or len(texts)):
                embeddings = self.embed_batch(client=client, batch=batch) # only last batch will be kept
                embeddings.extend(embeddings) # the last batch will be doubled
```

### Solution
- fix the variables assignment in above snippet
- use `zip(..., strict=True)` option when assigning embeddings to elements to guarantee a missmatch between number of generated embeddings and elements is caught in tests
- update the embeddings test file to have more elements than default batch size


mateuszkuprowski (approved):
Good catch! So wierd that we had this bug for so long and never noticed

PR #569, opened 2025-07-29
fix: prevent S3 path conflicts using hash-based directory isolation

CyMule:


cmscmadd:
Does this fix the file not found errors we sometimes see as an S3 source?

PastelStorm:
@claude sanity check

test/integration/connectors/utils/validation/source.py
```diff
@@ -167,11 +167,26 @@ def run_expected_download_files_validation(
...
+    if s3_keys_file.exists():
+        with s3_keys_file.open("r") as f:
+            s3_keys = json.load(f)["s3_keys"]
+        
+        expected_filenames = [Path(s3_key).name for s3_key in s3_keys]
+        actual_filenames = [Path(download_file).name for download_file in download_files]
+        
+        expected_filenames.sort()
+        actual_filenames.sort()
+        assert expected_filenames == actual_filenames, (
```
PastelStorm: It's not super important here and shouldn't be a blocker but in general I would avoid this pattern in the code.

I did some math and you should get about 10x-12x speedup if you create and compare two sets because TimSort has O(n*log n) complexity. Comparing two sets or two lists is the same O(n).

For 100k files this would be a difference of 3.5kk operations (sorted lists) vs. 300k operations (sets)

```
expected_filenames = {Path(s3_key).name for s3_key in s3_keys}
actual_filenames = {Path(download_file).name for download_file in download_files}

assert expected_filenames == actual_filenames
```

unstructured_ingest/processes/connectors/fsspec/fsspec.py
```diff
@@ -270,6 +270,22 @@ class FsspecDownloader(Downloader):
...
         default_factory=lambda: FsspecDownloaderConfig()
     )
+    
+    def get_download_path(self, file_data: FileData) -> Optional[Path]:
+        if not file_data.source_identifiers:
+            return None
+        
+        filename = file_data.source_identifiers.filename
+        if not filename:
+            return None
```
PastelStorm: define both booleans as variables, join them with an `and` and return `None` once

unstructured_ingest/processes/connectors/fsspec/fsspec.py
```diff
@@ -270,6 +270,22 @@ class FsspecDownloader(Downloader):
...
+        
+        mkdir_concurrent_safe(self.download_dir)
+        
+        temp_dir = tempfile.mkdtemp(
+            prefix="unstructured_",
```
PastelStorm: I'd make this a class-level constant

PastelStorm (approved):
A few nits but otherwise LGTM!

PR #572, opened 2025-07-31
chore: drop support for python 3.9

ds-filipknefel:
In lights of scheduled EOL in October (https://devguide.python.org/versions/) we're dropping support for python `3.9`.


unstructured_ingest/processes/connectors/fsspec/fsspec.py
```diff
@@ -367,7 +367,7 @@ class FsspecUploaderConfig(FileConfig, UploaderConfig):
 @dataclass
 class FsspecUploader(Uploader):
     connector_type: str = CONNECTOR_TYPE
-    upload_config: FsspecUploaderConfigT = field(default=None)
+    upload_config: FsspecUploaderConfigT = field(kw_only=True)
```
mateuszkuprowski: I suspect there are more places where we should make this change, you sure it's the only one?

ds-filipknefel: I'm going to drop this change actually and limit the scope of this PR to version limiting and dependency updates only (and code formatting required by pre-commit)

PR #573, opened 2025-07-31
fix: auto-detect S3 bucket region from error headers

CyMule:


qued:
I'm not that close to this, but it seems like we should add the ability to select the region when setting up the connector, maybe as follow-on work?

CyMule:
> I'm not that close to this, but it seems like we should add the ability to select the region when setting up the connector, maybe as follow-on work?

@qued Yes I would agree on that. Relying on the response headers will work, but is a bit of a hack and it means that if the bucket is new and not the default region we will always have to try twice which isn't ideal

potter-potter:
@claude please review

unstructured_ingest/processes/connectors/fsspec/s3.py
```diff
@@ -57,6 +57,11 @@ class S3AccessConfig(FsspecAccessConfig):
     token: Optional[str] = Field(
         default=None, description="If not anonymous, use this security token, if specified."
     )
+    region: Optional[str] = Field(
```
potter-potter, line 60: what do you think about having this be the AWS default region? what is the default region anyway?

CHANGELOG.md
```diff
@@ -1,10 +1,15 @@
+## 1.2.6
+
+* **Fix**: auto-detect S3 bucket region from error headers to resolve new bucket connection failures
+
 ## 1.2.5
 
-**Fix**: move S3 ambient_credentials out of AccessConfig
+* **Fix**: auto-detect S3 bucket region from error headers to resolve new bucket connection failures
```
potter-potter, line 7: some duplication here with line 3

potter-potter:
I say lets get this merged.

PR #582, opened 2025-08-06
fix: enable s3fs cache_regions for bucket region detection

CyMule:
## Problem

New S3 buckets in different regions fail with `UserAuthError: Access Denied` until AWS DNS/routing converges (2-3 hours) when you try to connect with the wrong region (i.e. use `us-east-1` for a bucket in `us-east-2`).

## Solution

Enable `cache_regions=True` in s3fs configuration to handle AWS region redirects automatically.

**Before:** Bucket access fails immediately, works after hours of waiting  
**After:** Bucket access works immediately regardless of region

### DIFF
--- a/unstructured_ingest/__version__.py
+++ b/unstructured_ingest/__version__.py
@@ -1,1 +1,1 @@
-__version__ = "99999"  # pragma: no cover
+__version__ = "1.2.9"  # pragma: no cover


CyMule:
> This seems cool. I wonder why this is false by default?

s3fs makes an additional head_bucket call for each new bucket to detect its region. If you are processing a large number of buckets and you know the region this would add some latency

We could make it configurable but on my default like:
```python
cache_regions_enabled = os.getenv("DISABLE_S3_REGION_CACHE", "").lower() != "true"
access_configs["cache_regions"] = cache_regions_enabled
```

potter-potter:
> > This seems cool. I wonder why this is false by default?
> 
> s3fs makes an additional head_bucket call for each new bucket to detect its region. If you are processing a large number of buckets and you know the region this would add some latency
> 
> We could make it configurable but on my default like:
> 
> ```python
> cache_regions_enabled = os.getenv("DISABLE_S3_REGION_CACHE", "").lower() != "true"
> access_configs["cache_regions"] = cache_regions_enabled
> ```

no. I think its good as is. Users will not have a large number of buckets most likely.

PR #583, opened 2025-08-08
Eng 319 standardize exception classes across the repo

bryan-unstructured:
the purpose is to standardize exception classes across the repo and make downstream services easier to handle exceptions. downstream services only need to catch unstructured-ingest.error.UnstructuredIngestError, because any exception thrown from ingest is a subclass of UnstructuredIngestError. downstream services no longer need to wrap_error


unstructured_ingest/errors_v2.py
```diff
@@ -1,24 +1,140 @@
-class UserError(Exception):
-    pass
+from abc import ABC
+from functools import wraps
+
+
+class APIError(Exception, ABC):
```
yuming-long: just a quick question why the based error named API error not ingest error?

unstructured_ingest/errors_v2.py
```diff
@@ -1,24 +1,140 @@
...
+    error_string: str
+
+    @classmethod
+    def wrap(cls, f):
```
yuming-long: i would try to avoid adding any wrapper logic possible

unstructured_ingest/processes/connectors/github.py
```diff
@@ -85,7 +85,7 @@ def wrap_http_error(self, e: "HTTPError") -> Exception:
         if status_code > 500:
             return ProviderError(e.response.text)
         logger.debug(f"unhandled http error: {e}")
-        return e
```
yuming-long, line 88: clearly we have a status code here
is there a way we put a status code in the base class?

unstructured_ingest/errors_v2.py
```diff
@@ -1,24 +1,140 @@
-class UserError(Exception):
```
yuming-long, line 1: also do we have to stick with errors_v2.py? why not just errors.py

unstructured_ingest/errors_v2.py
bryan-unstructured: not any particular reason, just I see some other libraries use this name, but I've changed it to IngestError

unstructured_ingest/errors_v2.py
bryan-unstructured: can you please elaborate, what's the concern with wrapper logic?

unstructured_ingest/errors_v2.py
yuming-long: or `UnstructuredIngestError` :)

unstructured_ingest/errors_v2.py
yuming-long: just an extra layer of processing :) if we can keep them simple and plain would be better

unstructured_ingest/processes/connectors/github.py
bryan-unstructured: added status code in base exception

unstructured_ingest/errors_v2.py
bryan-unstructured: replace errors_v2 by error

unstructured_ingest/errors_v2.py
bryan-unstructured: this might be another relatively big change, currently many connectors use @DestinationConnectionError.wrap or @SourceConnectionError.wrap decorator. we might want to leave it for another PR

unstructured_ingest/error.py
```diff
@@ -25,25 +27,130 @@ def wrapper(*args, **kwargs):
...
+    status_code: Optional[int] = 400
 
 
-class EmbeddingEncoderConnectionError(CustomError):
+class EmbeddingEncoderConnectionError(ConnectionError):
     error_string = "Error in connecting to the embedding model provider: {}"
+    status_code: Optional[int] = 400
+
+
+class UserError(UnstructuredIngestError):
```
yuming-long, line 55: thank you for these new error class! this is the pattern i wanted and we can iterate from here!

yuming-long:
@claude looks good right

yuming-long (approved):
LGTM 🎉

unstructured_ingest/processes/connectors/fsspec/gcs.py
```diff
@@ -125,7 +125,7 @@ def wrap_error(self, e: Exception) -> Exception:
                 raise UserError(message)
             if http_error_code >= 500:
                 raise ProviderError(message)
-        logger.error(f"unhandled exception from gcs ({type(e)}): {e}", exc_info=True)
+        logger.error(f"({type(e)} from gcs): {e}", exc_info=True)
```
bryan-unstructured, line 128: @yuming-long updated the error message so that it's less confusing

yuming-long: hmm they are still unhandled right? imo we should still keep this but maybe move to logger.debug

PR #584, opened 2025-08-09
improve astra db connector

potter-potter:


potter-potter:
@claude please review and suggest any improvements in the async await

PR #585, opened 2025-08-14
temporarily restore errors_v2.py

bryan-unstructured:
temporarily restore errors_v2.py. will remove it once all platform plugins are updated to import from error.py instead of errors_v2.py


potter-potter (approved):
lg

PR #587, opened 2025-08-15
fix Milvus stager to use correct exception

bryan-unstructured:
the correct exception to catch here is Python built in ValueError, not UnstructuredIngestError.ValueError

### DIFF
--- a/unstructured_ingest/processes/connectors/milvus.py
+++ b/unstructured_ingest/processes/connectors/milvus.py
@@ -10,7 +10,6 @@ from unstructured_ingest.data_types.file_data import FileData
 from unstructured_ingest.error import (
     DestinationConnectionError,
     KeyError,
-    ValueError,
     WriteError,
 )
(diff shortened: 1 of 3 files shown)


unstructured_ingest/processes/connectors/milvus.py
```diff
@@ -10,7 +10,6 @@
 from unstructured_ingest.error import (
     DestinationConnectionError,
     KeyError,
-    ValueError,
```
mateuszkuprowski, line 13: I see only exception type removed, but changelog says stager uses a correct type now which implies there should be a type switch somewhere in the code which I'm not seeing?

bryan-unstructured: So there are two ValueError, (1) Python built in, (2) unstructured_ingest.error.ValueError.
the error handling here is supposed to catch Python built in one. It shouldn't be the ValueError from unstructured_ingest.error

PR #588, opened 2025-08-18
retry with wait when throttling error happen in sharepoint

bryan-unstructured:


unstructured_ingest/processes/connectors/sharepoint.py
```diff
@@ -233,13 +251,27 @@ def _fetch_file(self, file_data: FileData) -> DriveItem:
...
+            reraise=True,
+        )
+        def _get_item_by_path() -> DriveItem:
+            try:
+                client_site = (
+                    client.sites.get_by_url(self.connection_config.site).get().execute_query()
+                )
+                site_drive_item = self.connection_config._get_drive_item(client_site)
+            except ClientRequestException:
+                logger.info("Site not found")
```
PastelStorm: I would use the same string here:
`(f"Site not found: {self.connection_config.site}")`
`loguru` supports f-strings, otherwise you can use standard `logging` string substitution

unstructured_ingest/processes/connectors/sharepoint.py
```diff
@@ -219,10 +225,22 @@ class SharepointDownloader(OnedriveDownloader):
...
     @requires_dependencies(["office365"], extras="sharepoint")
     def _fetch_file(self, file_data: FileData) -> DriveItem:
         from office365.runtime.client_request_exception import ClientRequestException
+        from tenacity import (
+            before_log,
+            retry,
+            retry_if_exception,
+            stop_after_attempt,
+            wait_exponential,
+        )
```
PastelStorm, line 238: the import should be at the top of the module

unstructured_ingest/processes/connectors/sharepoint.py
```diff
@@ -210,7 +211,12 @@ async def run_async(self, **kwargs: Any) -> AsyncIterator[FileData]:
 
 
 class SharepointDownloaderConfig(OnedriveDownloaderConfig):
-    pass
+    max_retries: int = Field(
+        default=10,
+        description="Maximum number of retries to download data",
+        ge=2,
+        le=50,
+    )
```
PastelStorm: This feels like an overkill, given that tenacity already has everything you might need for retries

test/unit/connectors/test_sharepoint.py
```diff
@@ -0,0 +1,119 @@
+from unittest.mock import Mock
```
PastelStorm, line 1: I don't see any happy path tests here

bryan-unstructured: added test_without_error

unstructured_ingest/processes/connectors/sharepoint.py
bryan-unstructured: changed to simpler form

unstructured_ingest/processes/connectors/sharepoint.py
bryan-unstructured: updated

unstructured_ingest/processes/connectors/sharepoint.py
bryan-unstructured: I thought loading the module within the function would help avoid loading too many things in the beginning, also avoid loading unnecessary modules. 
I saw other connector did similar thing: https://github.com/Unstructured-IO/unstructured-ingest/blob/abc27a2ae00d1d3119da410abb0c2fa25fc656e1/unstructured_ingest/processes/connectors/ibm_watsonx/ibm_watsonx_s3.py#L157

test/unit/connectors/test_sharepoint.py
```diff
@@ -0,0 +1,135 @@
...
+        Exception("Site not found")
+    )
+
+    with pytest.raises(SourceConnectionError, match="Site not found"):
+        sharepoint_downloader._fetch_file(file_data)
+
+    assert mock_client.sites.get_by_url.return_value.get.return_value.execute_query.call_count == 1
+
+
+def test_fetch_file_without_errors(
```
PastelStorm: Nit: positive tests should be named just `test_<function>`
Nit: positive tests should be at the top. Irrelevant when running test suites but makes it easier for manually running tests. As there's no reason to test edge cases if the sanity checks fail.

unstructured_ingest/processes/connectors/sharepoint.py
PastelStorm: probably makes sense in this case, I think it shows we're not structuring our connector logic correctly if we have to resort to this.

PastelStorm (approved):

PR #598, opened 2025-09-15
fix: only send fields in schema

qued:
Change to fix notion CI, but also make connectors a bit more robust.

Filters fields to pass only those in the data model to the `Page` constructor.


potter-potter (approved):
looks good

PR #649, opened 2026-02-20
fix: reinstate the confluence page limit

camilorequena:
Fixes Issue: Confluence page limits not working #645

Otherwise one can only get a maximum of 50 pages as this is the default of the atlassian confluence package.
The limit now seems to work as expected for client.get_all_pages_from_space.


jungep-0:
We also ran into this problem when trying to partition more than 50 docs from a confluence space.

pedro-correia-tech:
what is the status of this issue? I have found the same problem

PR #651, opened 2026-02-25
[PLU-121]: fix Teradata connector case sensitivity for Enterprise databases

simoncoombes:
## Summary
- Enterprise Teradata uppercases all column names in `cursor.description`, while ClearScape Analytics preserves the original case
- This causes a `KeyError` in the downloader when it tries to look up the user-configured `id_column` (lowercase) against uppercase DataFrame columns
- Normalize column names to lowercase in `TeradataDownloader.query_db` (line 157) and `TeradataUploader.get_table_columns` (line 205)

## Root Cause
Standard Enterprise Teradata folds unquoted identifiers to uppercase per the ANSI SQL spec. The existing connector was only tested against ClearScape Analytics (Teradata's developer edition), which preserves case as-typed. When a customer pointed the connector at Enterprise Teradata, `cursor.description` returned `ID` instead of `id`, causing the downstream `result.iloc[0][id_column]` lookup to fail with `KeyError: 'id'`.

## Changes
- `TeradataDownloader.query_db`: `columns = [col[0].lower() for col in cursor.description]`
- `TeradataUploader.get_table_columns`: `self._columns = [desc[0].lower() for desc in cursor.description]`

## Test plan
- [x] Verified against patched image (`teradata-case-fix`) on SND with ClearScape source — no regression
- [ ] Customer to verify against Enterprise Teradata once deployed

### DIFF
--- a/unstructured_ingest/processes/connectors/sql/teradata.py
+++ b/unstructured_ingest/processes/connectors/sql/teradata.py
@@ -154,7 +154,7 @@ class TeradataDownloader(SQLDownloader):
             logger.debug(f"running query: {query}\nwith values: {ids}")
             cursor.execute(query, ids)
             rows = cursor.fetchall()
-            columns = [col[0] for col in cursor.description]
+            columns = [col[0].lower() for col in cursor.description]
             return rows, columns
 
 
@@ -202,7 +202,7 @@ class TeradataUploader(SQLUploader):
         if self._columns is None:
             with self.get_cursor() as cursor:
                 cursor.execute(f'SELECT TOP 1 * FROM "{self.upload_config.table_name}"')
-                self._columns = [desc[0] for desc in cursor.description]
+                self._columns = [desc[0].lower() for desc in cursor.description]
         return self._columns
 
     def delete_by_record_id(self, file_data: FileData) -> None:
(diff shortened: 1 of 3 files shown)


tabossert:
@simoncoombes ok to close this?

unstructured_ingest/processes/connectors/sql/teradata.py
```diff
@@ -154,7 +154,7 @@ def query_db(self, file_data: SqlBatchFileData) -> tuple[list[tuple], list[str]]
             logger.debug(f"running query: {query}\nwith values: {ids}")
```
brndnblck, line 154: No tests verify the `.lower()` behavior - consider adding a unit test that mocks `cursor.description` returning uppercase column names and asserts the result is lowercase. Important for regression prevention since this fixes a production bug.

brndnblck, line 154: Snowflake connector handles the same ANSI SQL identifier uppercasing via `_fit_to_schema(case_sensitive=False)` in the base class. Worth considering that pattern here for consistency, though the immediate fix is correct.

simoncoombes:
Closing this as stale. It's several months behind main and CI has drifted; if the Teradata case-sensitivity fix is still needed I'll revive it on a fresh branch.

PR #677, opened 2026-03-30
fix(mongodb): add pymongo error handling to classify user vs platform errors

aballman:
## Summary

- Add explicit pymongo exception handling to `run_data()` and `delete_by_record_id()` in the MongoDB uploader
- Pymongo exceptions were falling through to the generic `Exception` catch in `api_generator.py`, defaulting to HTTP 500 and being classified as "platform" errors — this was the largest source of SLI misattribution for the DAG Job Completions SLO (933 errors in 7 days)
- Map pymongo exceptions to appropriate error types: `OperationFailure` with quota → `QuotaError` (401), `ServerSelectionTimeoutError` → `TimeoutError` (408), `BulkWriteError` → `WriteError` (400), `AutoReconnect` → `DestinationConnectionError` (400)

## Test plan

- [x] 8 new unit tests covering all exception mappings for both `run_data()` and `delete_by_record_id()`
- [x] Ruff lint and format pass
- [ ] After deploy: query Elasticsearch to confirm MongoDB uploader errors shift from `error_type: "platform"` to `error_type: "user"`

---

> [!NOTE]
> **Medium Risk**
> Changes how MongoDB write/delete failures are surfaced (error type/status code), which may affect downstream error handling and retries. Logic is straightforward and covered by new unit tests, but impacts runtime classification of production failures.
> 
> **Overview**
> **MongoDB uploader errors are now explicitly classified.** `MongoDBUploader.run_data()` and `delete_by_record_id()` catch key `pymongo` exceptions and re-raise them as typed ingest errors (`QuotaError`, `TimeoutError`, `WriteError`, `DestinationConnectionError`) instead of falling through as generic failures.
> 
> Adds a focused unit test suite (`test_mongodb_errors.py`) to assert the exception-to-error mapping for both insert and delete paths.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 75eb74098bd118d7c239fb5d827f89fe69bf54ae. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.6.3]
+
+### Fixes
+
+- **fix(mongodb): map pymongo exceptions to correct error types.** `MongoDBUploader` now classifies `OperationFailure` (quota / authorization) as `QuotaError`, `ServerSelectionTimeoutError` as `TimeoutError`, `BulkWriteError` as `WriteError`, and `AutoReconnect` as `DestinationConnectionError`. Previously these all surfaced as generic `DestinationConnectionError`, which the platform controller classified as `error_type=platform` and counted against the platform-error SLO denominator. Exception ordering is preserved so subclasses are caught before their parents (`BulkWriteError` before `OperationFailure`; `ServerSelectionTimeoutError` before `AutoReconnect`).
+
 ## [1.6.2]
 
 ### Fixes
--- a/test/unit/connectors/test_mongodb_errors.py
+++ b/test/unit/connectors/test_mongodb_errors.py
@@ -1,0 +1,128 @@
+from unittest.mock import MagicMock
+
+import pytest
+from pymongo.errors import (
+    AutoReconnect,
+    BulkWriteError,
+    OperationFailure,
+    ServerSelectionTimeoutError,
+)
+
+from unstructured_ingest.error import (
+    DestinationConnectionError,
+    QuotaError,
+    TimeoutError,
+    WriteError,
+)
+from unstructured_ingest.processes.connectors.mongodb import (
+    MongoDBConnectionConfig,
+    MongoDBUploader,
+    MongoDBUploaderConfig,
+)
+
+
+def _make_uploader():
+    connection_config = MagicMock(spec=MongoDBConnectionConfig)
+    connection_config.host = "test_host"
+    upload_config = MagicMock(spec=MongoDBUploaderConfig)
+    upload_config.record_id_key = "record_id"
+    upload_config.database = "test_db"
+    upload_config.collection = "test_collection"
+    upload_config.batch_size = 100
+    return MongoDBUploader(
+        connection_config=connection_config,
+        upload_config=upload_config,
+    )
+
+
+def _mock_client(uploader, collection_side_effect=None):
+    mock_client = MagicMock()
+    uploader.connection_config.get_client.return_value.__enter__ = MagicMock(
+        return_value=mock_client
+    )
+    uploader.connection_config.get_client.return_value.__exit__ = MagicMock(return_value=False)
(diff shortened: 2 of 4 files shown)


brndnblck (changes requested):
Bugbot flagged a real issue that needs addressing:

**`except Exception` swallows correctly classified errors**

When `delete_by_record_id()` raises `QuotaError` or `TimeoutError`, those exceptions propagate into `run_data()`'s try block, skip the pymongo-specific handlers (since they're no longer pymongo exceptions), and get caught by the generic `except Exception` - converting them back to `DestinationConnectionError`. This directly undermines the PR's goal.

Fix by re-raising already-classified errors before the catch-all:

```python
except (QuotaError, TimeoutError, WriteError, DestinationConnectionError):
    raise  # Re-raise already-classified errors
except Exception as e:
    raise DestinationConnectionError(f"MongoDB error: {e}") from e
```

Or if `UnstructuredIngestError` is the base class:
```python
except UnstructuredIngestError:
    raise
except Exception as e:
    raise DestinationConnectionError(f"MongoDB error: {e}") from e
```

Also worth adding a test that exercises the delete -> error -> propagation path to catch this regression.

aballman:
Thanks for the review. I want to push back on the main finding before applying the rest.

The "critical bug" claim that `except Exception` re-wraps the `QuotaError` from `delete_by_record_id` is not correct. Python evaluates `except` clauses top-to-bottom and the first matching clause wins. The `except (QuotaError, TimeoutError, WriteError, DestinationConnectionError): raise` is placed first, so when `delete_by_record_id` raises `QuotaError`, that clause matches and re-raises before any other handler (including `except Exception`) is checked.

Standalone repro to confirm the semantics:

```python
class A(Exception): pass
class B(Exception): pass

def inner():
    raise A("mapped")

def outer():
    try:
        inner()
    except (A, B):
        raise
    except Exception as e:
        raise RuntimeError(f"wrapped: {e}") from e

try:
    outer()
except A as e:
    print(f"GOT A (correct): {e}")
except Exception as e:
    print(f"GOT WRONG: {type(e).__name__}: {e}")
# Prints: GOT A (correct): mapped
```

The end-to-end test `test_delete_quota_error_not_relabeled_by_run_data_catch_all` (added in the prior commit) already validates this exact path: it mocks `collection.delete_many` to raise a quota `OperationFailure`, which `delete_by_record_id` maps to `QuotaError`, which then propagates up through `run_data`'s try block. The test asserts `pytest.raises(QuotaError)` and passes. If the catch-all were re-wrapping it, the test would fail with `DestinationConnectionError`. It does not.

So the ordering is correct as written. That said, your suggestion to use the base class is a strict improvement, so I applied that in 38c525ea: replaced the explicit tuple with `except UnstructuredIngestError: raise`. This catches the same types we already handle plus anything else inheriting from the ingest base (`SourceConnectionError`, `ConnectionError`, `ValueError`, etc.) if a future helper raises one. All 14 tests still pass.

On your other points:

- **Quota detection**: agreed, the structured `errmsg` approach is sufficient for now without baking in Atlas error codes I haven't verified.
- **NetworkTimeout ordering**: agreed, current ordering is correct.
- **Credential leak**: real risk worth tracking. pymongo's `OperationFailure.__str__` doesn't typically include the connection string, but the wrapped messages do interpolate the full exception. I've added a queue item to sanitize the wrapper output across all our connectors as a follow-up rather than scope-creep this PR.
- **WriteConcernError**: subclass of `OperationFailure`, current ordering routes it through the `except OperationFailure` handler correctly. Documenting in the code comment isn't load-bearing.
- **Catch-all `except Exception`**: agreed it's misleading shape-wise. Same follow-up as the credential leak.
- **Constructor stability**: confirmed for pymongo 4.x.

Ready for human review when one of the connector folks has a moment.

PR #686, opened 2026-04-09
fix(teradata): validate destination table schema in precheck and upload

tabossert:
## Summary
- Adds schema validation to `TeradataUploader.precheck()` — queries `DBC.ColumnsV` to verify an existing table has all required columns (`id`, `record_id`, `element_id`, `text`, `type`) before the workflow starts
- Adds the same validation at the top of `upload_dataframe()` as a runtime guard, so even if precheck is bypassed or the table is altered mid-run, the error is caught and surfaced cleanly
- Both raise `DestinationConnectionError` (a recognized error type) with a clear message naming the missing columns and suggesting remediation
- Previously, an incomplete table schema (e.g., only `id` and `text`) would fail deep in the SQL INSERT and surface as the generic "Unexpected job failure after 3 retries" in the UI

## Test plan
- [x] Added `test_teradata_uploader_precheck_rejects_missing_columns` — precheck raises on incomplete schema
- [x] Added `test_teradata_uploader_precheck_schema_check_skipped_when_table_name_none` — auto-create mode skips validation
- [x] Added `test_teradata_uploader_precheck_schema_check_tolerates_extra_columns` — extra columns are fine
- [x] Added `test_teradata_uploader_precheck_schema_check_case_insensitive` — handles Teradata uppercase
- [x] Added `test_teradata_uploader_precheck_skips_schema_check_when_table_missing` — nonexistent table defers to create_destination
- [x] Added `test_teradata_uploader_upload_dataframe_rejects_missing_columns` — runtime upload guard
- [x] All 66 existing tests pass

---

> [!NOTE]
> **Medium Risk**
> Adds new fail-fast schema validation for Teradata uploads, which can newly block workflows that previously proceeded (and failed later) when destination tables are missing required columns. Low code-surface change, but it affects runtime behavior and user-facing errors for existing Teradata destinations.
> 
> **Overview**
> **Teradata destination uploads now validate the destination table schema** before a workflow runs and again right before inserting rows.
> 
> `TeradataUploader.precheck()` queries `DBC.ColumnsV` for an existing table and raises a new `DestinationSchemaError` if required columns (`id`, `record_id`, `element_id`, `text`, `type`) are missing; it skips validation when the table is missing (defer to `create_destination`) or when `table_name` is `None` (auto-create mode). `upload_dataframe()` adds the same guard using discovered table columns to fail with a clear, actionable message instead of a late SQL failure.
> 
> Also bumps version to `1.4.21` and expands unit tests to cover missing/extra columns, case-insensitivity, and skip conditions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4dfe433d51a2af5aab280145a0c7b427073bcee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/unstructured_ingest/processes/connectors/sql/teradata.py
+++ b/unstructured_ingest/processes/connectors/sql/teradata.py
@@ -392,20 +392,15 @@ class TeradataUploader(SQLUploader):
                 cursor.execute("SELECT DATABASE")
                 current_db = cursor.fetchone()[0].strip()
                 cursor.execute(
-                    "SELECT TOP 1 1 FROM DBC.ColumnsV "
-                    "WHERE TableName = ? AND DatabaseName = ?",
-                    [table_name, current_db],
-                )
-                if not cursor.fetchone():
-                    # Table doesn't exist yet — create_destination will handle it.
-                    return
-
-                cursor.execute(
                     "SELECT ColumnName FROM DBC.ColumnsV "
                     "WHERE TableName = ? AND DatabaseName = ?",
                     [table_name, current_db],
                 )
-                existing = {row[0].strip().lower() for row in cursor.fetchall()}
+                rows = cursor.fetchall()
+                if not rows:
+                    # Table doesn't exist yet — create_destination will handle it.
+                    return
+                existing = {row[0].strip().lower() for row in rows}
         except Exception:
             # If we can't query DBC metadata, skip validation — the upload
             # will fail with its own error if the schema is truly wrong.
--- a/unstructured_ingest/processes/connectors/sql/teradata.py
+++ b/unstructured_ingest/processes/connectors/sql/teradata.py
@@ -389,17 +389,21 @@ class TeradataUploader(SQLUploader):
         """Check that an existing table has the required columns for uploads."""
         try:
             with self.get_cursor() as cursor:
+                cursor.execute("SELECT DATABASE")
+                current_db = cursor.fetchone()[0].strip()
                 cursor.execute(
-                    "SELECT 1 FROM DBC.ColumnsV WHERE TableName = ? LIMIT 1",
-                    [table_name],
+                    "SELECT TOP 1 1 FROM DBC.ColumnsV "
+                    "WHERE TableName = ? AND DatabaseName = ?",
+                    [table_name, current_db],
                 )
                 if not cursor.fetchone():
                     # Table doesn't exist yet — create_destination will handle it.
                     return
 
                 cursor.execute(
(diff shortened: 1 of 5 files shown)


tabossert:
failed tests unrelated to changes

potter-potter:
I don't think we usually go this deep on precheck. What was the motivation behind this.

potter-potter:
Users could set a custom record id key. I am not real psyched on this thing in general.

unstructured_ingest/processes/connectors/sql/teradata.py
```diff
@@ -378,6 +382,40 @@ def precheck(self) -> None:
...
+                    return
+                existing = {row[0].strip().lower() for row in rows}
+        except Exception:
+            # If we can't query DBC metadata, skip validation — the upload
+            # will fail with its own error if the schema is truly wrong.
+            return
+
+        missing = REQUIRED_DESTINATION_COLUMNS - existing
+        if missing:
+            raise DestinationConnectionError(
```
ctrahey: Recommend swapping out the name of this error to avoid the word "Connection" which could lead to a network-related hypothesis of the issue. 

Perhaps "DestinationConfigurationMismatchError" or something like that

unstructured_ingest/processes/connectors/sql/teradata.py
```diff
@@ -378,6 +382,40 @@ def precheck(self) -> None:
...
+        if table_name:
+            self._validate_table_schema(table_name)
+
+    def _validate_table_schema(self, table_name: str) -> None:
+        """Check that an existing table has the required columns for uploads."""
+        try:
+            with self.get_cursor() as cursor:
+                cursor.execute("SELECT DATABASE")
+                current_db = cursor.fetchone()[0].strip()
+                cursor.execute(
```
ctrahey, line 398: Do we want to also check for columns in the schema which are `NOT NULL` if we don't have a way to fill them? It's not a perfectly easy check because there could be `NOT NULL` with a default specified also, in which case we don't want to fail (this pattern is common with last_updated columns, for example).

ctrahey:
> I don't think we usually go this deep on precheck. What was the motivation behind this.

We want to move toward richer errors, especially if they are able to deterministically predict failure of the job. In fact, I'd say that's the entire purpose of the precheck. Also note that precheck-on-save is just one case. We also want to (soon? Eventually?) support a preflight of jobs when they startup in case some condition has changed (think column added or removed).

PR #692, opened 2026-04-12
Modifies sqlconnector base class with a function to sanitize table names for teradata

cmscmadd:
Teradata doesn't like - or leading numbers or periods or spaces in their table names. You have to wrap it in "" at query time and its a user experience they don't like. Teradata connector appears to be the only sql connector we have in ingest that supports opinionated writes. when it tries to create the sql table it appends the embedding models name that contains -. This change implements a sanitize table name function on the base class and implements it for the teradata connector to ensure friendly names in td. Teradata is ansi sql 2011 compliant, so if it any point any of the other sql connectors gain opinionated writes then we can enable this sanitizer for them also.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes Teradata destination table-name handling during `create_destination`, which can affect table creation and downstream writes if callers relied on the previous hard-error behavior.
> 
> **Overview**
> Teradata opinionated table creation now **sanitizes `destination_name` into a Teradata-safe table name** (replacing non-alphanumeric/underscore characters with underscores and prefixing leading digits) via a new `TeradataUploader.format_destination_name()` helper used by `create_destination()`.
> 
> Validation that previously rejected dashes in `TeradataUploaderConfig.table_name` and `TeradataUploader.precheck()` is removed, so *user-provided* table names are preserved as-is while only auto-generated names are normalized. Unit tests and the package version/changelog are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d11577bb72e8167405dec6799b94b905f4ab6147. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


unstructured_ingest/processes/connectors/sql/sql.py
```diff
@@ -328,11 +329,20 @@ class SQLUploaderConfig(UploaderConfig):
...
+        sanitized = f"_{sanitized}"
+    return sanitized
+
+
 @dataclass
 class SQLUploader(Uploader):
     upload_config: SQLUploaderConfig
     connection_config: SQLConnectionConfig
     values_delimiter: str = "?"
+    sanitize_destination_name: ClassVar[bool] = False
```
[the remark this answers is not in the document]
awalker4: Yeah, it looks like the convention has been to use `format_destination_name` on the child class. That feels like a cleaner approach that will only target teradata.

awalker4 (approved):
Pending those cursorbot comments, lgtm mr. haddock! FYI unrelated CI fix here: https://github.com/Unstructured-IO/unstructured-ingest/pull/693

awalker4:
Okay now you can rebase, grab my test fixes, remove beta on the changelog. And then hopefully a make tidy will deal with that long line.

potter-potter:
So if a user put in 22super-table. Will it pass precheck? (I'm gonna do a deeper dive, but just asking)

If it only sanitizes at upload, then the user will have a bad experience.

potter-potter:
For this to work correctly, it really should fail win the precheck with some sort of warning that you can't have table names with numbers or dashes.

We should never rename a table without them knowing. They may not be able to find the table.

And thats why opinionated writes is something we unwillingly accepted during the scrappy days.

And even worse is the empty table name, which we try not to include in the documentation.

unstructured_ingest/processes/connectors/sql/teradata.py
```diff
@@ -369,12 +371,8 @@ def create_destination(
         that stores metadata as a single JSON column instead of flattening into 20+ columns,
         keeping the schema stable as upstream element fields evolve. Requires the stager to
         have metadata_as_json=True so that element metadata is serialized before insert."""
+        destination_name = self.format_destination_name(destination_name)
         table_name = self.upload_config.table_name or destination_name
-        if "-" in table_name:
```
potter-potter, line 373: I think this is where the sanitize should be used as a check to return an error instead of creating a table that the user woldn't know the name of.

cmscmadd: it only sanitizes when its creating the table itself through opinionated writes. if the user passes a table name it accepts that.

cmscmadd:
Follow-up on the outstanding review comments:

**Addressing cursor[bot] findings (now resolved by recent commits):**

1. *"Base class flag misleads subclasses"* — the `sanitize_destination_name` ClassVar has been removed entirely. Sanitization is now done via `format_destination_name()` on `TeradataUploader` only, following the established child-class convention (as @awalker4 suggested).

2. *"Sanitization silently modifies user-provided config table names"* — in the current code, `self.format_destination_name(destination_name)` runs on the auto-generated `destination_name` *parameter*, then `table_name = self.upload_config.table_name or destination_name` picks the user's explicit value first. User-provided names are never touched.

3. *"Validator still rejects dashes that create_destination now sanitizes"* — the `table_name_must_not_contain_dashes` field validator has been removed.

**On @potter-potter's concern about silent renaming:**

Only the *system-generated* `destination_name` (e.g. derived from workflow/model IDs) gets sanitized — never a `table_name` the user explicitly set. The logger emits `creating table {table_name} for user` so the actual name is visible in logs. If a user provides `table_name="my-table"`, that value passes through unchanged (and would fail in Teradata at query time, which is intentional — user-owned names are the user's responsibility).

Also pushed: removed the `b1` beta suffix from `[1.4.23b1]` in CHANGELOG per @awalker4's request, and ran `ruff format` to tidy the function signature.

PR #694, opened 2026-04-17
fix(teradata): fix off-by-one in split_dataframe that sent empty exec…

cmscmadd:
When a document had exactly N * batch_size elements the last chunk was empty - this causes teradatasql to send a zero-row parameterized batch to the Teradata server which returned Error 3939 (parameter count mismatch).

Replace with range(0, len(df), chunk_size) which never produces an empty slice.

---

> [!NOTE]
> **Low Risk**
> Small, well-scoped change to a shared batching utility plus targeted unit tests; primary risk is subtle behavior change for callers that (incorrectly) relied on receiving empty chunks.
> 
> **Overview**
> Fixes an off-by-one bug in `split_dataframe` so it no longer yields an empty trailing DataFrame slice when `len(df)` is an exact multiple of `chunk_size`, preventing zero-row `executemany` batches that triggered Teradata Error 3939 (and benefiting all connectors that share this helper).
> 
> Adds Teradata-focused regression tests covering exact-multiple chunking cases, and bumps the package version/changelog entry to `1.4.26b1`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4e839f6b7f695941b0ce04460607fe1e81a1763. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/test/unit/connectors/sql/test_teradata.py
+++ b/test/unit/connectors/sql/test_teradata.py
@@ -1181,3 +1181,33 @@ def test_teradata_uploader_create_destination_preserves_user_table_name_with_das
     uploader.create_destination()
 
     assert uploader.upload_config.table_name == "my-table"
+
+
+# --- Regression tests for Error 3939 (parameter count mismatch) ---
+
+
+def test_split_dataframe_no_empty_chunk_when_length_is_multiple_of_chunk_size():
+    """split_dataframe must not yield an empty trailing chunk when len(df) % chunk_size == 0.
+
+    The old formula (num_chunks = len(df)//chunk_size + 1) always appended one
+    extra iteration, producing an empty df slice that caused teradatasql to send
+    a zero-row parameterized batch to the server (Error 3939).
+    """
+    from unstructured_ingest.utils.data_prep import split_dataframe
+
+    df = pd.DataFrame({"a": range(100)})
+    chunks = list(split_dataframe(df, chunk_size=100))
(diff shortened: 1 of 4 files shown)


potter-potter (approved):
LG

test/unit/connectors/sql/test_teradata.py
```diff
@@ -1181,3 +1181,33 @@ def test_teradata_uploader_create_destination_preserves_user_table_name_with_das
     uploader.create_destination()
 
     assert uploader.upload_config.table_name == "my-table"
+
+
+# --- Regression tests for Error 3939 (parameter count mismatch) ---
```
awalker4, line 1186: Is this a hallucinated bug ID? Might be better to just explain the failure mode being tested

awalker4: Sorry, it's already doing that below haha. Maybe just drop this comment

awalker4 (approved):

test/unit/connectors/sql/test_teradata.py
cmscmadd: 3939 is the exception number from tera

cmscmadd: parameter count mismatch is the error

PR #696, opened 2026-04-21
bug(PLU-332): quote fixer for databricks

potter-potter:
> [!NOTE]
> **Low Risk**
> Low risk bugfix limited to Databricks connectors’ session setup SQL; main risk is any behavioral change for unusual identifier names or quoting expectations.
> 
> **Overview**
> Fixes Databricks Delta Tables and Databricks Volumes Delta Tables connectors to **backtick-quote** `catalog` and `database` when issuing `USE CATALOG` / `USE DATABASE`, preventing `INVALID_IDENTIFIER` errors for names containing hyphens, spaces, reserved words, or backticks.
> 
> Introduces a shared `quote_identifier()` helper (with escaping and empty-value validation), adds unit tests covering quoting/escaping and emitted SQL, and bumps the package version + changelog to `1.4.28`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 21c1b7e60bf94c49f2786d6df6a2de76e30c0457. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/test/unit/connectors/sql/test_databricks_delta_tables.py
+++ b/test/unit/connectors/sql/test_databricks_delta_tables.py
@@ -43,9 +43,6 @@ def test_quote_identifier_rejects_empty_and_whitespace(bad):
 
 
 def test_quote_identifier_never_uses_single_quotes():
-    # If someone "simplifies" this back to f"'{name}'" the Databricks
-    # parser will reject USE CATALOG 'name-with-dashes' at runtime
-    # (SQLSTATE 42602 / INVALID_IDENTIFIER). This is the original regression.
     out = quote_identifier("utic-dev-tech-fixtures")
     assert not out.startswith("'")
     assert not out.endswith("'")
--- a/test/unit/connectors/sql/test_databricks_delta_tables.py
+++ b/test/unit/connectors/sql/test_databricks_delta_tables.py
@@ -1,9 +1,21 @@
-"""Unit tests for databricks quote_identifier helper."""
+from unittest.mock import MagicMock
 
 import pytest
+from pydantic import Secret
+from pytest_mock import MockerFixture
 
 from unstructured_ingest.error import ValueError as IngestValueError
+from unstructured_ingest.processes.connectors.sql.databricks_delta_tables import (
+    DatabricksDeltaTablesAccessConfig,
+    DatabricksDeltaTablesConnectionConfig,
+    DatabricksDeltaTablesUploader,
+    DatabricksDeltaTablesUploaderConfig,
+)
 from unstructured_ingest.utils.databricks import quote_identifier
+
+HYPHENATED_CATALOG = "utic-dev-tech-fixtures"
+HYPHENATED_DATABASE = "my-db"
+TABLE_NAME = "elements"
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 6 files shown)


potter-potter:
@claude please review

potter-potter:
@claude please review this last fix

unstructured_ingest/processes/connectors/sql/databricks_delta_tables.py
```diff
@@ -33,6 +33,14 @@
 CONNECTOR_TYPE = "databricks_delta_tables"
```
brndnblck, line 34: Worth extracting `_quote_identifier` to a shared utility module? It's defined here but imported in `volumes_table.py` as a private function (underscore prefix), which creates coupling and divergence risk.

Something like `unstructured_ingest/utils/databricks.py` would give both connectors a single source of truth.

unstructured_ingest/processes/connectors/sql/databricks_delta_tables.py
```diff
@@ -33,6 +33,14 @@
...
 
 
+def _quote_identifier(name: str) -> str:
+    """Wrap ``name`` as a Databricks backtick-quoted identifier so hyphens,
+    reserved words, etc. are legal; embedded backticks are doubled."""
```
brndnblck:
```suggestion
def _quote_identifier(name: str | None) -> str:
    """Wrap ``name`` as a Databricks backtick-quoted identifier so hyphens,
    reserved words, etc. are legal; embedded backticks are doubled."""
    if not name or not name.strip():
        raise ValueError("identifier is required and cannot be empty")
    return "`" + name.replace("`", "``") + "`"
```
The type hint says `str` but the body checks for `None` - aligning the signature and also rejecting empty strings would catch config issues earlier.

unstructured_ingest/processes/connectors/sql/databricks_delta_tables.py
```diff
@@ -137,7 +145,8 @@ def run(self, path: Path, file_data: FileData, **kwargs: Any) -> None:
     @contextmanager
     def get_cursor(self) -> Generator[Any, None, None]:
```
brndnblck, line 139: This adds `USE DATABASE` that wasn't in the original `_connection` context manager - is that intentional? Just want to confirm since `volumes_table.py` already had it but this file didn't.

unstructured_ingest/processes/connectors/databricks/volumes_table.py
```diff
@@ -23,6 +23,7 @@
 from unstructured_ingest.processes.connectors.sql.databricks_delta_tables import (
     DatabricksDeltaTablesConnectionConfig,
```
brndnblck, line 24: Nit: importing a private function (`_` prefix) cross-module. Not blocking, but if you do extract to a shared util, renaming to `quote_identifier` (public) would be cleaner.

unstructured_ingest/processes/connectors/sql/databricks_delta_tables.py
potter-potter: Definitely considered. Dont really want to add more files.

unstructured_ingest/processes/connectors/databricks/volumes_table.py
potter-potter: yeah. debated this.

I will remove `_`

unstructured_ingest/processes/connectors/sql/databricks_delta_tables.py
potter-potter: good question. definitely intentional. 

this was a bug. just never revealed itself since most people just use database='default' i guess

The old code only ran USE CATALOG, so even if you set database="my_schema" in the config, Databricks was still writing to the default schema because nothing ever told it to switch. Adding USE DATABASE makes the connector actually honor the database setting

volumes_table.py was already doing this, this file just wasn't.

unstructured_ingest/processes/connectors/databricks/volumes_table.py
potter-potter: whoops. alan did some of this.

unstructured_ingest/processes/connectors/sql/databricks_delta_tables.py
potter-potter: whoops alan already did this.

unstructured_ingest/utils/databricks.py
```diff
@@ -0,0 +1,11 @@
...
+
+from unstructured_ingest.error import ValueError
+
+
+def quote_identifier(name: Optional[str]) -> str:
+    """Wrap ``name`` as a Databricks backtick-quoted identifier so hyphens,
+    reserved words, etc. are legal; embedded backticks are doubled."""
+    if not name or not name.strip():
+        raise ValueError("identifier is required and cannot be empty")
+    return "`" + name.replace("`", "``") + "`"
```
brndnblck, line 11: Optional hardening: consider rejecting control chars and edge whitespace? Current logic handles the real problem, but these values could still produce confusing SQL/log behavior.

```suggestion
    if not name or not name.strip():
        raise ValueError("identifier is required and cannot be empty")
    if name != name.strip() or any(ch in name for ch in ("\n", "\r", "\t")):
        raise ValueError("identifier cannot contain leading/trailing whitespace or control characters")
    return "`" + name.replace("`", "``") + "`"
```

test/unit/connectors/sql/test_databricks_delta_tables.py
```diff
@@ -0,0 +1,44 @@
+"""Unit tests for databricks quote_identifier helper."""
```
brndnblck: Nice helper coverage. Worth adding a connector-level regression test (mock cursor) to assert both `USE CATALOG` and `USE DATABASE` execute with quoted identifiers? Would directly protect the PLU-332 fix.

brndnblck (approved):

brndnblck:
Also test failure here that needs to be resolved before merge:
https://github.com/Unstructured-IO/unstructured-ingest/actions/runs/24747612424/job/72403174146?pr=696

PR #698, opened 2026-04-27
feat: add additional_metadata_fields to Pinecone upload stager config

micmarty-deepsense:
## Summary

Adds `additional_metadata_fields` to `PineconeUploadStagerConfig`, allowing users to include custom enricher-generated metadata keys in Pinecone vector payloads beyond the hardcoded `ALLOWED_FIELDS` allowlist.

## Problem

The Pinecone stager uses a hardcoded `ALLOWED_FIELDS` tuple to filter which metadata keys are written to each vector. Any key not in this list is silently dropped. Users running enricher nodes that produce custom metadata keys have no way to surface those keys in Pinecone.

## Solution

- Add `additional_metadata_fields: list[str] = []` to `PineconeUploadStagerConfig`
- Keys in this list are unioned with `metadata_fields` in the allowlist filter
- Keys are also passed as `keys_to_omit` to `flatten_dict`, preserving `list[str]` values for native Pinecone `$in` filtering (without this, `flatten_lists=True` converts `["a", "b"]` to indexed keys `field-0`, `field-1`)
- Add 40KB warning at 75% threshold; upgrade metadata drop log from INFO to ERROR

## Precedent

Similar to Milvus connector's `fields_to_include` added in platform 3.5.0.

### DIFF
--- a/unstructured_ingest/processes/connectors/pinecone.py
+++ b/unstructured_ingest/processes/connectors/pinecone.py
@@ -108,6 +108,14 @@ class PineconeUploadStagerConfig(UploadStagerConfig):
             "Pinecone."
         ),
     )
+    additional_metadata_fields: list[str] = Field(
+        default=[],
+        description=(
+            "Additional metadata fields to include beyond the default allowed list. "
+            "Useful for custom enricher-generated keys. "
+            "List values are preserved as-is for native Pinecone $in filtering."
+        ),
+    )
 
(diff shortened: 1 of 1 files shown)


micmarty-deepsense:
Closing per team feedback — needs proper spike and backward-compatibility review before any solution is shipped. Tracking in UNI-64.

PR #701, opened 2026-05-03
fix(PLU-348): azure ai search improvement dropping fields

potter-potter:
## Summary
Azure AI Search uploads were failing with `400: The property 'table_extraction_method' does not exist on type 'search.complex.metadata'` after `unstructured` added `metadata.table_extraction_method` (PR Unstructured-IO/unstructured#4346). The destination Edm.ComplexType validates strictly, so any field not declared in the index schema is rejected.

The connector already had a "drop unknown fields" filter, but it only inspected top-level keys (`set(doc) - set(index.fields)`), so anything new under `metadata.*` (or any other complex type) sailed past it and 400'd at the service.

## Fix
Replace the depth-0 filter with a schema-driven recursive one:

- `get_index_field_tree` walks `index.fields` recursively into a `FieldTree = dict[str, FieldTree | None]` (leaves → `None`, complex / collection-of-complex → nested dict).
- `filter_doc_against_tree` walks the document against that tree and drops any key not declared at the matching level, recursing into both `dict` values and `list[dict]` (collection-of-complex) values.
- `collect_dropped_paths` produces sorted dotted paths (e.g. `metadata.table_extraction_method`) for the existing `"Following fields will be dropped..."` INFO log so nested drops are now visible.
- Defensive cap at `_MAX_INDEX_FIELD_DEPTH = 10`, matching Azure's [documented complex-type nesting limit](https://learn.microsoft.com/en-us/azure/search/search-limits-quotas-capacity).

The legacy `filter_doc` and `get_index_field_names` are kept (with their original semantics) so any subclass override or external caller keeps working.

## Backward compatibility
- Customers whose docs already matched the schema: no behavior change (filter is a no-op when nothing is unknown).
- Customers who were silently shipping unknown nested keys: those keys are now stripped pre-upload instead of returning 400 — strictly an improvement and consistent with how the connector has always treated unknown top-level keys.
- Customers with `metadata` modeled as `Edm.String` (JSON blob): the value passes through verbatim regardless of shape (covered by a dedicated test).

---

> [!NOTE]
> **Medium Risk**
> Changes Azure AI Search upload behavior by recursively filtering documents against the index schema; mistakes in schema-walking or depth limits could drop fields unexpectedly or mask type issues, but scope is isolated to this connector and covered by new unit/integration tests.
> 
> **Overview**
> Fixes Azure AI Search uploads failing with HTTP 400s when *nested* fields not declared in the index schema (e.g. `metadata.table_extraction_method`) are present.
> 
> `AzureAISearchUploader.run_data` now builds a recursive `FieldTree` from `index.fields`, prunes unknown keys at any depth (including collections of complex types), and logs dropped *dotted paths* via `collect_dropped_paths` (with a defensive max-depth cap). Extensive new unit tests plus an integration regression test cover nested-field dropping and logging; version is bumped to `1.5.1` and noted in `CHANGELOG.md`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 92222742bf4cc6bda73b60e62befe53240004b75. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.5.1]
+
+### Fixes
+
+- **fix(azure_ai_search): drop unknown fields at every nesting level of the index schema.** The existing "drop unknown fields" filter only inspected top-level keys, so new nested metadata fields from `unstructured` (e.g. `metadata.table_extraction_method`) reached the service and were rejected with HTTP 400. The filter now recursively walks `index.fields` and prunes any sub-field not declared by the destination schema.
+
 ## [1.5.0]
 
 ### Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Fixes
 
-- **fix(azure_ai_search): drop unknown fields at every nesting level of the index schema.** The existing "drop unknown fields" filter only inspected top-level keys, so new nested metadata fields from `unstructured` (e.g. `metadata.table_extraction_method`) reached the service and were rejected with HTTP 400. The filter now recursively walks `index.fields` and prunes any sub-field not declared by the destination schema.
+- **fix(azure_ai_search): recursively drop unknown fields against the index schema.** New nested fields from `unstructured` (e.g. `metadata.table_extraction_method`) were reaching the service and causing HTTP 400s; the filter now recurses through `index.fields`.
 
 ## [1.5.0]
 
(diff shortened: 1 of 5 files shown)


potter-potter:
@claude please review

awalker4 (approved):

PR #705, opened 2026-05-07
feat(embed): add CustomOpenAICompatibleEmbeddingConfig for OpenAI-compatible gateways

micmarty-deepsense:
### Problem

OpenAI-compatible gateways like vLLM, NIM, and similar proxies often authenticate via a custom HTTP header rather than a Bearer token. The platform needs to support this: send custom headers, and don't send an `Authorization` header at all when no upstream API key is configured.

The previous attempt added `default_headers` and made `api_key` optional on the base `OpenAIEmbeddingConfig`. That's a silent regression: `AzureOpenAIEmbeddingConfig` inherits from the base and calls `self.api_key.get_secret_value()` unconditionally — making `api_key` optional there causes an `AttributeError` at runtime.

### Solution

New subclass `CustomOpenAICompatibleEmbeddingConfig(OpenAIEmbeddingConfig)` carries the optional `api_key` and `default_headers` fields. The base class is untouched.

The key behavioral guarantee: when `api_key` is unset, **no `Authorization` header is sent to the gateway**. This matters because the OpenAI SDK always adds `Authorization: Bearer <key>` — a placeholder like `"ignored"` gets forwarded upstream and causes real 401s from any OpenAI-routed model. The suppression works by passing `api_key=lambda: ""` to the SDK constructor; the SDK stores `""` and its `auth_headers` property returns `{}` on a falsy key. Verified stable across `openai==2.20.0` through `2.36.0`, pinned `<3` defensively.

An `Authorization` key in `default_headers` is rejected at validation time (the SDK owns that header; callers who need a Bearer should use `api_key` instead).

### What to review

- `embed/openai.py`: the new subclass only. The base class diff should be zero.
- The `api_key=lambda: ""` mechanism is non-obvious — the comment in the source and the behavioral test `test_without_api_key_omits_authorization` are the proof.

### Tests

Unit tests verify the behavioral contract directly: no `Authorization` header on outbound requests when `api_key` is unset (sync and async), correct Bearer when set, plaintext custom headers on the wire, `Authorization` rejected in `default_headers`, JSON round-trip. All pass locally and in CI across Python 3.11/3.12/3.13.


awalker4 (approved):

awalker4:
LGTM! Although I guess I jumped on this early if you're still drafting it

micmarty-deepsense:
## Validation evidence

The code path added in this PR (`OpenAIEmbeddingConfig.get_client` / `get_async_client` with optional `api_key` and `default_headers` threading on top of the OpenAI SDK) was exercised end-to-end via a downstream platform integration test on 2026-05-22:

- **Positive smoke** against an OpenAI-compatible gateway protected by a custom auth header: 10/10 files processed, 87 elements per file produced 1024-dim vectors via a `Snowflake/snowflake-arctic-embed-l-v2.0` model fronted by the gateway.
- **Negative smoke** (intentionally bogus value for the custom auth header): the embed plugin pod received HTTP 401 from the gateway with body `{"detail":"Invalid or expired API key"}` (not `"Missing authentication"`), confirming the header value is threaded all the way through `default_headers` to the SDK and onto the wire rather than being dropped, replaced, or short-circuited.

A downstream consumer is currently carrying a local subclass override of `get_client` / `get_async_client` with identical logic because this PR has not yet shipped in a release. Once 1.6.3 lands and the downstream lock catches up, that override can be deleted.

Marking ready for review; already approved by @awalker4.

micmarty-deepsense:
@awalker4 thanks for the prior approval. The approach changed significantly since your review: three independent reviews (including the cubic bot above) flagged that making `api_key` Optional on the base class breaks `AzureOpenAIEmbeddingConfig` silently.

The rewrite introduces `CustomOpenAICompatibleEmbeddingConfig` as a **subclass** instead. The base class is unchanged. The key change is that when `api_key` is unset, we pass `api_key=lambda: ""` to the SDK, which causes `auth_headers` to return `{}` so no `Authorization` header reaches the gateway. Verified stable across `openai==2.20.0` through `2.36.0`.

Quick re-look would be appreciated - the diff on `openai.py` is a net addition with zero base-class changes, and the 10 unit tests cover the behavioral guarantee directly.

awalker4 (approved):

awalker4:
Feel free to merge over the flaky partitioner tests

PR #709, opened 2026-05-15
Fix google native files failures

mateuszkuprowski:
The failure was caused by Google Docs Editors files being treated like binary Drive files in some paths. Google rejects files.get(...?alt=media) for native Docs/Sheets/Slides/Drawings with a 403 and requires files.export(...) instead.

The mitigation makes export handling explicit: supported native formats are exported to processable formats, extension-filtered indexing now includes exportable native MIME types, and unsupported Google-native types fail clearly instead of surfacing as misleading direct-download 403s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.6.2]
+
+### Fixes
+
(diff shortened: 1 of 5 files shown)


jordan-homan:
are the test failures related?

mateuszkuprowski:
> are the test failures related?

Unfortunately not, I noticed CI has been brekaing here for a long time before. connectors break on Notion connector and api based tests are just flaky, I've been rerunning them couple of times and sometimes they work. Like on previous commit they did work, but then I only bumped version after that and they break.

jordan-homan (approved):

potter-potter:
@claude please review

potter-potter (approved):
LG

PR #712, opened 2026-05-20
feat(PLU-161): flatten_metadata option for databricks_volume_delta_tables

awalker4:
## Summary
- Adds opt-in `flatten_metadata: bool` (default `False`) to `DatabricksVolumeDeltaTableUploaderConfig` and `DatabricksVolumeDeltaTableStagerConfig`.
- When set, the stager recursively flattens nested metadata into top-level columns (no `metadata_` prefix) using the existing `flatten_dict` helper (`utils/data_prep.py`) — stops at lists; matches Milvus's flat naming.
- When set, the uploader skips auto-create and intersects incoming columns with the user-managed table schema, dropping unknowns with a log line (mirrors the existing `_fit_to_schema` message in `sql/sql.py`).
- Switches the destination-registry's `upload_stager_config` from the empty SQL pass-class to the now-non-empty Volumes-specific stager config so the new field surfaces in the dashboard schema. Safe — verified no external imports of the swapped class and no field loss.
- Backward compatibility is preserved: with the flag absent or `False`, behavior is byte-identical to today (single `metadata VARIANT` blob, default auto-create). Workflow DB-serialized configs without the field deserialize cleanly with `flatten_metadata=False`.
- No type coercion on the flatten path — values flow into Delta as their JSON-native types. Customers designing the user-managed table should declare datetime fields (`date_created`, `date_modified`, `date_processed`, and the `data_source_*` variants) as `STRING`; they arrive as stringified or float unix epochs.

## Why
Linear: [PLU-161](https://linear.app/unstructured/issue/PLU-161/metadata-flattening-databricks-delta-tables). Customers want to query Delta tables via normal SQL columns (\`SELECT filename, data_source_url ...\`) rather than dotting into a \`VARIANT\` blob. Per Potter, this is also intended as the reusable pattern for other table-like connectors.

## Example job run

All metadata fields are flattened. You can see `languages` comes in as a list,
Screenshot 2026-05-22 at 1 41 30 PM

## Test plan
- [x] New unit tests: stager blob-mode default, flatten drops the `metadata_` prefix, flatten stops at lists, both configs deserialize with default `False` for Workflow DB backcompat, uploader missing-table hard-fail, uploader PARSE_JSON wrapping on the non-flatten path, uploader unknown-column drop + log.
- [x] Existing `test_databricks_delta_tables.py` (SQL connector) and the destination-config JSON-schema test still pass — confirms the registry switch doesn't ripple.
- [x] SND manual matrix: cases include backcompat, default schema auto-create, flatten with custom table, missing-table hard-fail, nullable columns, nested keys, list-valued columns. Pending plugin-image build off this branch.


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## [1.7.0]
+
+### Enhancements
+
```
awalker4, line 4: Way too verbose

test/unit/connectors/databricks/test_volumes_table.py
```diff
@@ -0,0 +1,173 @@
+import json
```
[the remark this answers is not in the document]
awalker4: In the future please be specific about missing coverage. Propose new cases that need to be tested.

awalker4:
Finally address unrelated flaky test: https://github.com/Unstructured-IO/unstructured-ingest/pull/714

unstructured_ingest/processes/connectors/databricks/volumes_table.py
```diff
@@ -93,6 +145,12 @@ def create_destination(
             table_names = [r[1] for r in cursor.fetchall()]
             if table_name in table_names:
                 return False
+            if self.upload_config.flatten_metadata:
+                raise ValueError(
+                    f"Table {table_name!r} does not exist. With flatten_metadata=true, "
```
awalker4: I need to look a bit closer at this. When I tested flatten=True and an invalid table name, we hit this error per document when the job ran. We want this to raise at precheck. My test setup might have been weird though.

potter-potter:
@claude review

potter-potter:
@claude please give one more review. we dropped the datetime and timestamp conversions because we dont want to do that anymore

potter-potter (approved):
LG

PR #713, opened 2026-05-21
fix(notion): make database test row-order insensitive

awalker4:
## Summary
- Notion's database query endpoint doesn't guarantee a stable row order, but the `test_notion_source_database` integration test compared the downloaded HTML byte-for-byte against a checked-in fixture. Every prior "fix" (#693, #605) was just reshuffling the same four rows.
- Add `unordered_table_html_equality_check` that compares table header positionally and data rows as a multiset of their text content, and wire it into the database test only (the page test isn't a table).
- On mismatch the new check prints the symmetric difference so the next real regression won't be a black box like https://github.com/Unstructured-IO/unstructured-ingest/actions/runs/26238030726/job/77216703553 was.

## Test plan
- [ ] `uncategorized_connectors_int_test` Notion tests pass (real verification — needs the CI's `NOTION_API_KEY`)
- [x] Smoke-tested locally against the real fixture: reordered rows compare equal, a tampered row fails with a readable diff

---
## Summary by cubic
Make the Notion database integration test resilient to unstable row ordering by comparing table rows without relying on their sequence. This prevents flaky CI from nondeterministic Notion database query results.

- **Bug Fixes**
  - Added `unordered_table_html_equality_check` that compares the first table’s header positionally and treats data rows as a multiset of their text.
  - Wired this check into `test_notion_source_database` only; the page test is unchanged.
  - On mismatch, prints a readable diff showing header differences and row counts.

- **Dependencies**
  - Bumped `unstructured_ingest` to `1.6.3` and added a changelog entry.

<sup>Written for commit 4590c97db31a0f2e1ee2750d702e162b07bdde1f. Summary will update on new commits. <a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/713?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.6.3]
+
+### Fixes
+
+- **fix(notion): make `test_notion_source_database` row-order insensitive.** The integration test compared the downloaded HTML byte-for-byte, but Notion's database query endpoint doesn't guarantee a stable row order — every prior fixture refresh was just reshuffling the same rows. Compare table rows as a multiset of their text content and print the symmetric difference on mismatch so real regressions aren't silent.
+
 ## [1.6.2]
 
 ### Fixes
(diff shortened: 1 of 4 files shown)


awalker4:
Folding the test fix into #712 (PLU-161) — no need for a separate PR with its own version bump for a test-only change.

PR #714, opened 2026-05-21
fix(PLU-373): make notion database test row-order insensitive

awalker4:
## Summary
- Notion's database query endpoint doesn't guarantee stable row order, but `test_notion_source_database` compared the downloaded HTML byte-for-byte. Every prior "fix" (#693, #605) was just reshuffling the same four rows.
- Add `unordered_table_html_equality_check` that compares the header positionally and data rows as a multiset of their text content, and wire it into the database test only (the page test isn't a table).
- On mismatch the new check prints the symmetric difference so the next real regression won't be a black box like https://github.com/Unstructured-IO/unstructured-ingest/actions/runs/26238030726/job/77216703553 was.
- Version bumped to `1.6.2-dev` to satisfy `check-version` without claiming a release slot for a test-only change.

## Test plan
- [x] `uncategorized_connectors_int_test` Notion tests pass (real verification — needs the CI's `NOTION_API_KEY`)
- [x] Smoke-tested locally against the real fixture: reordered rows compare equal, a tampered row fails with a readable diff

---
## Summary by cubic
Make the Notion database integration test row-order insensitive and correct a misleading helper comment, eliminating flaky failures when Notion returns rows in different orders. Addresses PLU-373.

- **Bug Fixes**
  - Added `unordered_table_html_equality_check` to compare the header positionally and treat table rows as a multiset; prints a readable symmetric diff on mismatch.
  - Used in `test_notion_source_database` only; the page test remains unchanged.
  - Fixed the helper comment to state it scans all `<tr>` elements, not just the first table.
  - Bumped version to `1.6.6-dev` and updated `CHANGELOG.md`; test-only change with no runtime impact.

<sup>Written for commit 87d92cc9f28f5ffa0892328d3e0cfcf8b701c7e7. Summary will update on new commits. <a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/714?utm_source=github">Review in cubic</a></sup>


potter-potter:
@claude please review

potter-potter (approved):
LG

PR #733, opened 2026-06-15
test(connectors): exclude env-specific additional_metadata.url from notion/onedrive/sharepoint diffs

fxdgear:
## Problem

The connector integration suites (`blob_storage_connectors_int_test`, `uncategorized_connectors_int_test`) are failing on **`main` and every open PR** — e.g. the run that produced 1.6.13 (#728) merged with these red. Failures are all `AssertionError: Diffs found between files` in:
- `test_notion.py::test_notion_source_database`, `::test_notion_source_page`
- `test_onedrive.py::test_onedrive_source`
- `test_sharepoint.py::*` (6 tests)

## Root cause

The only diff is `root['additional_metadata']['url']`, and the committed fixture values show why:
- Notion: `https://www.notion.so/test-doc1-1572c376…` — embeds the **page id**
- OneDrive: `/drives/b!w-wVyIs…/root:/eml/fake-email.txt` — embeds the **drive id**
- SharePoint: `/drive/root:/…`

`additional_metadata.url` embeds **environment-specific identifiers** (Notion page id, MS Graph drive id) that change when the test workspace/tenant is re-provisioned. There has been **no connector code change** to these sources, so the URL *format* is intact — only the embedded ids drifted, which is why three independent connectors began failing on the same field at the same time.

## Fix

Exclude `additional_metadata.url` from the fixture comparison for the affected source tests. This is consistent with the harness already excluding other dynamic/environment-specific fields — `metadata.date_created/modified`, `additional_metadata.LastModified`, `additional_metadata.@microsoft.graph.downloadUrl`, and astradb's `display_name` ("includes dynamic ids, might change"). It keeps the suite robust to tenant re-provisioning instead of re-pinning ids that will drift again on the next migration.

## Test plan
- [x] `make check` (ruff) — clean
- [ ] CI: `blob_storage_connectors_int_test` + `uncategorized_connectors_int_test` go green (cannot run integration suites locally — they require the live Notion / MS Graph test tenants)

Note: unblocks #732 (S3 download integrity), which is otherwise green but sits behind this pre-existing red.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/733?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.6.14]
+
+### Fixes
+
+- **fix(connectors): stop asserting environment-specific URLs in Notion/OneDrive/SharePoint source integration tests.** The expected-results diff compared `additional_metadata.url` (Notion page id, MS Graph drive id) and `@microsoft.graph.downloadUrlNoAuth` (a tokenized no-auth download link), both of which vary by test tenant/workspace and per request. With no connector code change, these drifted and reddened `blob_storage_connectors_int_test` and `uncategorized_connectors_int_test` on main and every PR. Both fields are now excluded from the comparison, consistent with the existing exclusions of `@microsoft.graph.downloadUrl`, `LastModified`, and `date_*`.
+
 ## [1.6.13]
 
 ### Fixes
--- a/test/integration/connectors/test_notion.py
+++ b/test/integration/connectors/test_notion.py
@@ -59,7 +59,12 @@ def test_notion_source_database(temp_dir):
             test_id="notion_database",
             expected_num_files=1,
             validate_downloaded_files=True,
-            exclude_fields_extend=["metadata.date_created", "metadata.date_modified"],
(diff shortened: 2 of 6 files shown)


potter-potter:
https://github.com/Unstructured-IO/unstructured-ingest/pull/727/
^
I'm pretty sure this is being fixed in this PR. Just gotta get it merged.

potter-potter:
Well that onedrive and sharepoint issue is new I think. 

Strange

awalker4 (approved):

PR #734, opened 2026-06-16
fix(google_drive): refresh OAuth access tokens via authorized_user credentials

ryannikolaidis:
## Why It Matters

Google Drive ingestion authenticated with a raw OAuth access token that `google-auth` cannot refresh. Access tokens expire after ~1 hour; on the next request after expiry `google-auth` looks for refresh material (`refresh_token`, `token_uri`, `client_id`, `client_secret`), finds none, and raises:

```
[RefreshError] The credentials do not contain the necessary fields need to refresh the access token.
You must specify refresh_token, token_uri, client_id, and client_secret.
```

The indexer then lists zero files and the job "completes" with nothing processed. In practice a Drive source works right after it is connected and silently stops ingesting about an hour later. Google Cloud Storage already avoids this by accepting a self-refreshing `authorized_user` credential; Google Drive shares the same OAuth flow but never gained the consumption path.

## Summary

The Google Drive connector now consumes an `authorized_user` credential. When `service_account_key` holds a credential whose `type` is `authorized_user` (carrying `refresh_token`, `client_id`, `client_secret`, and `token_uri`), the connector builds it with `Credentials.from_authorized_user_info(...)` instead of treating the key as a service account. `google-auth` then mints a fresh access token automatically whenever the current one expires, so long-running and repeated jobs no longer fail with `RefreshError`.

The two existing authentication paths are unchanged: a raw `oauth_token` is still used directly as a bare access token, and a service-account key still goes through `from_service_account_info`. The connector decides purely on the `type` field of the parsed key, so callers that already work keep working.

This is the connector half of the fix. For the path to engage in the hosted platform, the control plane must supply the `authorized_user` credential JSON (the same translation already performed for Google Cloud Storage); that producer-side change ships separately.

## Changes

- `unstructured_ingest/processes/connectors/google_drive.py`:
  - `get_client()` and `_get_credentials()` build a self-refreshing `Credentials.from_authorized_user_info(key, scopes=[DRIVE_READONLY_SCOPE])` when the parsed `service_account_key` has `type == "authorized_user"`; otherwise the service-account path is taken as before.
  - The streaming downloader needs no new branch: an `authorized_user` key leaves `oauth_token` unset, so the existing `else` branch already calls `_get_credentials()` and `creds.refresh(Request())`, which now refreshes the authorized-user credential and uses the freshly minted token.
  - Added a `DRIVE_READONLY_SCOPE` module constant for the Drive read-only scope.
- `test/unit/connectors/test_google_drive.py`: added `TestGoogleDriveAuthorizedUserCredentials` covering that `_get_credentials()` and `get_client()` produce a refreshable credential from an `authorized_user` key (dict and JSON-string forms), that the streaming downloader refreshes and uses a freshly minted token, that the raw `oauth_token` path stays non-refreshable and never builds/refreshes a credential, and that a service-account key still dispatches to `from_service_account_info`.
- Version bump to `1.6.16` with the matching `CHANGELOG.md` entry.
</content>

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/734?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## E2E Validation

Validated live on a personal SND against real Google OAuth and a real connected Google Drive corpus. The `authorized_user` credential this change produces refreshes against live Google (fresh access token, `expires_in 3599`, `drive.readonly` scope) and lists the real corpus (10 files); the pre-fix raw-token credential returns HTTP 401 once expired with no way to refresh — the production failure, reproduced live. Full before/after evidence: https://gist.github.com/ryannikolaidis/b8fbe9cb309140561776b29eb6efc607

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.6.16]
+
+### Fixes
+
+- **fix(google_drive): refresh OAuth access tokens via authorized_user credentials.** The Google Drive connector authenticated with a raw OAuth access token that google-auth cannot refresh, so once the token expired (~1 hour) the indexer raised `RefreshError` and listed zero files. When `service_account_key` holds an `authorized_user` credential (carrying `refresh_token`, `client_id`, `client_secret`, and `token_uri`), the connector now builds it via `Credentials.from_authorized_user_info`, so google-auth mints a fresh access token automatically on expiry. The service-account and raw `oauth_token` paths are unchanged.
+
 ## [1.6.15]
 
 ### Fixes
--- a/test/unit/connectors/test_google_drive.py
+++ b/test/unit/connectors/test_google_drive.py
@@ -4,7 +4,7 @@ from unittest.mock import MagicMock
 import pytest
 
 from unstructured_ingest.data_types.file_data import FileData, SourceIdentifiers
-from unstructured_ingest.error import SourceConnectionError, ValueError
+from unstructured_ingest.error import SourceConnectionError, UserAuthError, ValueError
 from unstructured_ingest.processes.connectors.google_drive import (
     GOOGLE_DRIVE_SKIP_MIME_TYPES,
     GOOGLE_EXPORT_MIME_MAP,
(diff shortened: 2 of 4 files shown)


ryannikolaidis:
Closing — superseded by an operator-layer fix.

FS-1981 (`platform-etl-orchestration` #1047, merged 2026-05-15) resolves the Google Drive / GCS / Dropbox / Microsoft OAuth token-expiry bug at the etl-operator: an `init-oauth-refresh` init container refreshes the connector's access token from its stored `refresh_token` before each source job, using provider client credentials configured on the operator (`OAUTH_CRED_*`). It needs no change to `unstructured-ingest` — the widget already persists the `refresh_token`, and the connector keeps consuming a (now freshly-refreshed) `oauth_token`.

This PR's approach — having the connector consume an `authorized_user` credential and self-refresh — is a parallel design the team didn't adopt; with FS-1981 in place it's redundant. Validated end-to-end on an SND: with the operator on a post-FS-1981 image + `OAUTH_CRED_GOOGLE_*`, a Drive source that had been silently indexing 0 files (expired token) processed its full corpus again.

Closing as superseded. Happy to reopen if the team decides connector-side self-refresh is still wanted (e.g. for single jobs that themselves run past the ~1h token lifetime, which the pre-job init refresh doesn't cover).

PR #738, opened 2026-06-23
fix(PLU-437): Azure OpenAI precheck

mpolomdeepsense:
### What
`AzureOpenAIEmbeddingConfig` inherited OpenAI's precheck, which validates the model against `client.models.list()`. On Azure that returns the base-model catalog (`text-embedding-3-small`, …), never deployment names — yet a deployment name is the only valid `model` value. Any deployment whose name differs from the base-model name (the normal Azure pattern) failed the precheck, making the embed step unusable. Reported by a customer running Azure OpenAI on AKS.

### Change
Override `run_precheck` on `AzureOpenAIEmbeddingConfig` to validate via a minimal live `embeddings.create(input="precheck", model=<deployment>)` instead of catalog membership. This checks auth, endpoint, and deployment existence using the only identifier Azure accepts, and works with any deployment name. The model is already validated against the model registry upstream, so the catalog check was also redundant.

- `DeploymentNotFound` (404) → `UserError` naming the deployment.
- All other API errors route through `wrap_error` (fixes a silent-failure path where non-404 API errors were swallowed and the precheck falsely passed).

### Tests
- New `test/unit/embed/test_azure_openai.py`: happy path, asserts `models.list()` is not called, `DeploymentNotFound`→`UserError`, auth-error-not-swallowed regression, unexpected-error wrapped.
- Existing `test/integration/embedders/test_azure_openai.py` still passes.

### Notes
Precheck now makes one small billable embeddings call per embedder init (previously a free `models.list()`). Flows into `platform-plugins-embed` via a dependency bump — no plugin code change needed.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/738?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.6.21]
+
+### Fixes
+
+- **fix(embed): validate Azure OpenAI deployments by reaching them, not via `models.list()`.** The Azure precheck inherited from OpenAI validated the model against `client.models.list()`, which on Azure returns the base-model catalog (`text-embedding-3-small`, …), never deployment names — so any deployment whose name differs from the base model (the normal Azure pattern) failed the precheck. `AzureOpenAIEmbeddingConfig.run_precheck` now validates via a minimal live `embeddings.create(input="precheck", model=<deployment>)` instead. `DeploymentNotFound` surfaces as a `UserError`; all other API errors route through `wrap_error`, closing a path where non-404 errors were silently swallowed.
+
 ## [1.6.20]
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Fixes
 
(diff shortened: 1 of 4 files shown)


awalker4:
I haven't touched the embedding module a whole lot. Do we even expose prechecks for non connector plugins? It's possible `wrap_in_fastapi` picks it up automatically, but then does anything in Platform call it? @potter-potter do you know?

Edit: I think you need to explicitly add `precheck_func` to `wrap_in_fastapi()` to hook up the plugin endpoint. Man, that wrapper is so dumb

potter-potter:
> I haven't touched the embedding module a whole lot. Do we even expose prechecks for non connector plugins? It's possible `wrap_in_fastapi` picks it up automatically, but then does anything in Platform call it? @potter-potter do you know?
> 
> Edit: I think you need to explicitly add `precheck_func` to `wrap_in_fastapi()` to hook up the plugin endpoint. Man, that wrapper is so dumb

Yeah. I have absolutely no idea. Haven't looked at the embed in u-i in years.

We should change codeowners for embed

tabossert (approved):

mpolomdeepsense:
@potter-potter @awalker4 bumping for approve

awalker4 (approved):
LGTM! Only comment is that the changelog is a bit verbose. I'd cut to down to what the customer impact was and the tldr on the fix

PR #722, opened 2026-05-27
fix(teradata): classify driver errors and surface user-fault TD messages

simoncoombes:
## Summary

Customer-fixable Teradata SQL destination errors (missing object, missing privilege, SQL syntax, schema mismatch) were being swallowed into a generic `"Unexpected job failure after 3 retries"` wrapper in the platform UI, hiding the actual server message. Mirrors the PLU-337 pattern already shipped for `teradata_vector_v2`.

Replaces #721 — clean single commit off latest `main`.

Linear: [PLU-377](https://linear.app/unstructured/issue/PLU-377/port-plu-337-error-surfacing-pattern-to-legacy-teradata-sql)

### `unstructured_ingest/error.py`
- **BREAKING:** `UserError.status_code` `401` → `422`. `401` is HTTP "Unauthorized" (unauthenticated) which is what `UserAuthError` covers. `UserError` is for invalid user input/config (HTTP 422). `UserAuthError` (`401`) and `RateLimitError` (`429`) override and are unaffected; `QuotaError` inherits without override and now reports `422`.

### `unstructured_ingest/processes/connectors/sql/teradata.py`
- `_USER_FAULT_TERADATA_CODES` (frozen via `MappingProxyType`):
  - `3807` object missing or no privilege
  - `3523` / `5612` / `5315` no privilege
  - `3706` / `3707` SQL syntax
  - `3753` / `3754` implicit type conversion / schema mismatch
- `_is_teradata_driver_error(exc)` — module-name check; lets non-teradatasql exceptions (`MemoryError`, `OSError`, etc.) bubble up unchanged.
- `_extract_teradata_error_code(exc)` — `re.findall` + last-match-wins (innermost / most specific server-side code).
- `_raise_classified_teradata_error(exc, *, host, table=None, direction: Literal["source","destination"]="destination", fallback_context="")`:
  - recognised user-fault code → `UserError` with `"Teradata reported: …"` carrying the raw TD message
  - anything else → existing `DestinationConnectionError` / `SourceConnectionError`
  - `direction` validated at runtime to catch typos that mypy can't
  - `fallback_context` forwarded to `_summarize_error` so the indexer can preserve its `"table 'X' not found or not accessible"` message for unrecognised codes
- `cursor.execute` wrapped in `TeradataUploader.{get_table_columns, delete_by_record_id, upload_dataframe, create_destination}` and `TeradataIndexer.precheck` table-probe. Each site re-raises non-teradatasql exceptions unchanged.

### Tests
[...]

PR #742, opened 2026-06-29
fix(teradata): auto-create destination table at upload time

tabossert:
## Summary

A Teradata **destination** connector with the database name left empty and a table name given (table not pre-created), with **"metadata as JSON" = Yes**, fails the workflow at upload with:

> `422: Error in uploader - Teradata error 3807 (object does not exist or user has no privilege on it) for 'uns_382'. ... [Error 3807] [SQLState 42S02] Object 'uns_382' does not exist.`

The table is supposed to be auto-created, but it never is.

## Root cause

The opinionated-table auto-create (`TeradataUploader.create_destination`, added in #654 for the "metadata as JSON" write mode) is reachable **only** through `TeradataUploader.init()`, and the **sole** caller of `init()` in the library is the local `Pipeline._run_initialization`. #655 then removed precheck's table-existence guard in favour of "auto-create instead of failing".

So when an external orchestrator (the Unstructured Platform) drives only the upload phase, `init()` never runs, the destination table is never created, and the first DB touch in `upload_dataframe` (`can_delete` → `get_table_columns` → `SELECT TOP 1 *`) hits Teradata **3807**, surfaced as the 422 `UserError` above. (The 422 wording itself is the #722 error-classification working correctly — the real defect is the missing table.)

## Fix

`upload_dataframe` now ensures the destination exists before its first table access by calling the already-idempotent `create_destination()` **once per uploader instance** (memoized via `_destination_ensured`). This brings the upload run-path to parity with the `init()` path and is independent of how the orchestrator drives the connector.

- **No clobbering:** `create_destination()`'s `DBC.TablesV` existence check returns early without issuing `CREATE` when the table already exists, so a user's pre-existing table is never recreated.
- **Cheap:** the existence probe runs at most once per run.
- **Covers** both a user-supplied table name that isn't pre-created (the reported case) and the no-table-name default-table case.

Mirrors the existing in-run-path creation pattern already used by the Neo4j connector.

## Behavior note

Auto-create is intentionally **not** gated on `metadata_as_json` — the uploader has no reference to the stager's config, and adding a required flag would re-break the orchestrator-driven path this fixes. This matches what `init()` already does today (it also unconditionally creates the opinionated table when missing), so a flattened-mode user pointing at a missing table now gets the opinionated table auto-created rather than a 3807 error — consistent with the #654/#655 "auto-create instead of failing" direction.

## Testing

- TDD: added 3 unit tests (named-but-missing table → auto-create; no-name → default table; existing table → no `CREATE`, idempotent). Confirmed red before the fix, green after.
- `test/unit/connectors/sql/test_teradata.py`: 98 passed. All SQL connectors: 161 passed. `ruff check`: clean.
- Patch version bump `1.6.22` → `1.6.23` + CHANGELOG entry.

Resolves Pylon #2098.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/742?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.6.23]
+
+### Fixes
+
+- **fix(teradata): auto-create the destination table during upload, not only during pipeline init.** Table auto-creation ran only via `TeradataUploader.init()`, whose sole caller is the local `Pipeline`. Orchestrators that drive only the upload phase (e.g. the Unstructured Platform) never created the table, so a not-pre-created destination failed with Teradata error 3807 (`Object '<table>' does not exist`). `upload_dataframe` now ensures the table exists itself — once per run, via the idempotent `create_destination()` whose `DBC.TablesV` check skips `CREATE` for existing tables, so a user's table is never recreated.
+
 ## [1.6.22]
 
 ### Enhancements
--- a/test/unit/connectors/sql/test_teradata.py
+++ b/test/unit/connectors/sql/test_teradata.py
@@ -1522,3 +1522,131 @@ def test_teradata_uploader_create_destination_no_privilege_raises_user_error(
         match=r"3523.*does not have the required privilege.*brand_new_table",
     ):
         uploader.create_destination(destination_name="brand_new_table")
+
+
+# --- Run-path auto-create (lazy create_destination in upload_dataframe) -------
+# init() is the only auto-create trigger and only the local Pipeline calls it, so
+# upload_dataframe must ensure the table exists itself — idempotently — or
+# orchestrators that skip init() fail with Teradata 3807 ("object does not exist").
+
+_OPINIONATED_COLUMNS = [
+    ("id",), ("record_id",), ("element_id",), ("text",), ("type",), ("metadata",),
+]
+
+
+def _opinionated_df() -> pd.DataFrame:
+    return pd.DataFrame(
+        {
+            "id": ["1"],
+            "record_id": ["r1"],
+            "element_id": ["e1"],
+            "text": ["hello"],
+            "type": ["NarrativeText"],
+            "metadata": ["{}"],
+        }
+    )
+
+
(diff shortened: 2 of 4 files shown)


tabossert:
https://app.usepylon.com/support/issues/views/inbox?issueNumber=2098

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
...
...only through `TeradataUploader.init()`, and the sole caller of `init()` in the library is the local `Pipeline._run_initialization`. PR #655 then removed precheck's table-existence guard in favour of "auto-create instead of failing". When an external orchestrator (e.g. the Unstructured Platform) drives only the upload phase, `init()` never runs, so a not-yet-created destination table is never created; the first statement in `upload_dataframe` (`can_delete` → `get_table_columns` → `SELECT TOP 1 *`) then fails with Teradata error 3807 (`Object '<table>' does not exist`), surfaced to the user as a 422 `UserError`. `upload_dataframe` now ensures the destination exists by calling `create_destination()` once per uploader instance before its first table access, bringing the upload run-path to parity with the `init()` path. `create_destination()` remains idempotent — its `DBC.TablesV` existence check returns early without issuing `CREATE` when the table already exists — so a user's pre-existing table is never recreated and the existence probe runs at most once per run. This covers both a user-supplied table name that is not pre-created and the no-table-name (default opinionated table) case.
```
awalker4: non blocking - this is pretty noisy for just trying to see what changed

awalker4 (approved):

PR #745, opened 2026-06-30
fix(FS-2108): download Jira attachment content to correct path with attachment-specific display name

mateuszkuprowski:
## Summary
Jira attachment downloads now write bytes to each attachment's own `source_identifiers` path (instead of a separate `attachments/` directory) and use the attachment filename as `display_name` instead of inheriting the parent issue title.

Fixes FS-2108

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/745?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


awalker4 (approved):
Lgtm pending the cubic security comment

PR #746, opened 2026-07-02
feat: weaviate auto schema option

mpolomdeepsense:
## Summary

Adds an opt-in `auto_schema` option to the Weaviate destination connector that lets Weaviate build the collection and its columns from the uploaded objects, so a destination collection no longer has to be created (with a full, correct schema) up front.

`auto_schema` defaults to `false`, so **existing behavior is unchanged unless the flag is set**.

## Why

With `flatten_metadata=true` the connector previously required the collection schema to be declared up front and dropped any property not already in it. `auto_schema=true` hands schema creation to Weaviate's native auto-schema so those fields land as columns automatically.

## Behavior

| Mode | `auto_schema=false` (default) | `auto_schema=true` |
|---|---|---|
| non-flatten | as before: nested `metadata`, nothing dropped, collection seeded from default config on `init` | Weaviate auto-creates the collection/columns on insert; nothing dropped |
| `flatten_metadata=true` | as before: conform to existing schema (unknown dropped, missing null-filled), collection must pre-exist | flattened metadata lands as **dynamic top-level columns**, nothing dropped, no predefined schema needed |

Requires `AUTOSCHEMA_ENABLED=true` on the Weaviate side (its default).

## Changes

- New `auto_schema` field on `WeaviateUploaderConfig` (and mirrored on `WeaviateUploadStagerConfig`).
- `precheck`: in `auto_schema` mode, don't require the collection to exist and don't read/enforce the schema; validate connectivity only.
- `run_data`: schema fetch + `conform_to_schema` now gated on `flatten_metadata and not auto_schema`; delete-by-record_id is skipped on the first run when the collection doesn't exist yet (Weaviate creates it on insert).
- Stager: extracted the value-normalization into a shared helper and apply it in **flatten + `auto_schema`** mode before flattening. Without this, `coordinates.points` reaches Weaviate as an unsupported array-of-arrays and every insert fails. Pure flatten (`auto_schema=false`) still passes raw values through for a user-declared schema.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/746?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.6.24]
+
+### Enhancements
+
+- **feat(weaviate): add `auto_schema` option to the destination connector.** New uploader option that defaults to `false`. When `false`, the connector behaves exactly as before: the collection must already exist (or, in non-flatten mode, is seeded from the default config), and in `flatten_metadata` mode each object is conformed to the existing schema (unknown properties dropped, missing ones set to null). When `true`, the connector skips the schema fetch/conform step and lets Weaviate create the collection and its columns from the uploaded objects on first insert, so a collection does not need to exist up front (requires `AUTOSCHEMA_ENABLED=true` in Weaviate). Combined with `flatten_metadata=true`, this allows dynamic, up-front-unknown metadata to land as top-level columns without a predefined schema.
+
 ## [1.6.23]
 
 ### Fixes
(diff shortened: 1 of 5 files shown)


potter-potter:
run_data adds a collections.exists() round-trip on every upload, in all modes - worth thinking about.

potter-potter (approved):
Looks good. With small comment. You choose if its important.

PR #748, opened 2026-07-06
fix(security): redact credentials from connector error logging

lawrence-u10d:
Superseded by #749 — the head branch was renamed, which auto-closed this PR. See #749 for the full description and review notes.


awalker4 (approved):
Awesome! My one comment is we should drop the IR references in an open repo.

PR #749, opened 2026-07-07
fix(security): redact credentials from connector error logging

lawrence-u10d:
## Summary

Remediates credential leakage in connector error handling: caught provider exceptions were interpolated into raised error messages and logs across ~33 connector call sites, and could embed API keys, passwords, DSNs, tokens, and service-account JSON.

**Core redaction**
- **Connector precheck / upload / path-validation errors** no longer carry provider exception text; `exc_info=True` frame dumps removed from credential-bearing paths (a rendered traceback's terminal line re-embeds `str(e)`).
- **Central fix:** `UnstructuredIngestError.wrap` no longer formats `str(error)` from foreign exceptions.
- **Diagnostics preserved via allowlist:** new `safe_error_summary()` renders the exception type name plus fields safe by construction — integer HTTP statuses, SDK error-code enums (SQLSTATE, `errno`, provider codes), opaque request/correlation IDs — e.g. `SlackApiError(status_code=401, error=invalid_auth)`. Free text (`str(e)`, `.args`, `.message`, bodies, URLs, headers) is never surfaced.
- **GCS service-account-JSON echo** guarded (the `Path(<json>).is_file()` → `OSError: File name too long` path).

**Refinements from cubic review (later commits)**
- **Preserve connector-authored guidance:** prechecks re-raise dependency-install hints (`requires_dependencies` `ImportError`) and connector-authored typed errors (missing index/db/collection, SCRAM-SHA-1 remediation, OAuth error codes, corpus-name mismatch) instead of flattening them to a bare type name.
- **Suppress the exception chain:** every redacted re-raise now uses `from None`, so provider text can't resurface via traceback logging (verified: 0 `safe_error_summary` re-raises still chain the source exception).
- **gcs guard narrowed** to `ENAMETOOLONG` so a genuine `OSError` on a real path isn't masked as "Invalid auth token value".
- **Elasticsearch bulk-upload failure logs** now allowlist safe fields (`_index`, `_id`, `status`, `error.type`) instead of only stripping the uploaded document, so document content in `error.reason` is no longer logged.
- **Guard test** pinning the `wrap` sibling pass-through: a connector error chained `from` a secret-bearing provider exception surfaces only its own sanitized message, never the chained cause.

Exception classes and control flow are unchanged at every site — only log and error-message payloads were sanitized. Merged current `main`; version bumped to **1.6.31** with CHANGELOG entry.

## Notes for reviewers

- **SharePoint is intentionally left identical to `main` — this PR no longer modifies `sharepoint.py`.** `main`'s #1.6.29 work rewrote the SharePoint error path to deliberately surface the real upstream HTTP status, a truncated (≤500 char) response body, and Microsoft correlation headers (the opposite of this PR's redaction). We defer to that newer, intentional design. ⚠️ That means SharePoint error messages can carry a truncated upstream body — a security follow-up for the team to reconcile with the redaction policy separately (see the pinned merge note).
- **Broader leak surface tracked separately:** connectors untouched by this PR still interpolate raw exceptions on connection paths (MotherDuck, VastDB, Teradata, DuckDB, Databricks volumes, Zendesk) plus `salesforce.py` private-key parsing and `chroma.py` upsert — filed as its own ticket, out of scope here.
- **error.py sibling-wrap:** the `UnstructuredIngestError.wrap` sibling pass-through keeps our own (sanitized) message by design rather than reducing it to a summary; the invariant is now defended by a guard test rather than changed.
- **neo4j integration tests**: assertions were remapped from driver message text to driver exception type names (`AuthError`, `ServiceUnavailable`, `ClientError`); they need a live neo4j container, so please watch integration CI.
- `errors_v2.py` contains an identical leaky `wrap` but is imported nowhere (dead module) — left untouched, candidate for deletion.
- Supersedes #748, which was auto-closed by a branch rename.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/749?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


awalker4 (approved):

paulkarayan:
I'm going to look through these as part of [FIE-197
](https://linear.app/unstructured/issue/FIE-197/improved-connector-error-logging-ingest-etl-orchestration)

they are completion issues, not blocking.

1. HIGH — the security objective isn't met: several connectors still leak on the exact path the PR targets

  The PR fixed the SQLIndexer/SQLUploader base prechecks in sql/sql.py, but sibling connectors override precheck and were missed — they
  still do logger.error(f"...{e}", exc_info=True) + raise ...Error(f"...{e}"), identical to the pattern redacted everywhere else:

  - MotherDuck — duckdb/motherduck.py:99-100 — highest value: the connection carries motherduck_token=.... A bad-token OperationalError
  embeds the connection string, and it's both logged (with full traceback) and raised into the API response.
  https://github.com/Unstructured-IO/unstructured-ingest/blob/556065e8/unstructured_ingest/processes/connectors/duckdb/motherduck.py#L9
  9-L100
  - VastDB — sql/vastdb.py:106-107 (and :190-191) — access_key/secret_key connector.
  https://github.com/Unstructured-IO/unstructured-ingest/blob/556065e8/unstructured_ingest/processes/connectors/sql/vastdb.py#L106-L107
  - Teradata — sql/teradata.py:311 — the raise goes through _summarize_error(host, e), but the logger.error(f"...{e}", exc_info=True)
  line still dumps the raw exception + traceback (host/user/password path).
  https://github.com/Unstructured-IO/unstructured-ingest/blob/556065e8/unstructured_ingest/processes/connectors/sql/teradata.py#L311
  - DuckDB — duckdb/duckdb.py:100-101 (same pattern; local, lower credential risk).
  - Databricks volumes — databricks/volumes.py:95, Zendesk — zendesk/client.py:219,245, s3 — fsspec/s3.py:153: unhandled-exception logs
  still exc_info=True.

  Failure scenario: MotherDuck workflow with a wrong token → OperationalError whose text/traceback includes
  md:?motherduck_token=<token> → written to logs and returned to the caller. This is precisely the leak class the PR's own description
  says it closes.

  2. MEDIUM — wrap_error "classified" branches still interpolate provider free-text (in-scope leak class, left in the handled path)

  The PR redacted only the fallback/unhandled branches of the fsspec/watsonx wrap_error methods; the classified 4xx/5xx branches still
  put provider text into UserError/ProviderError, which propagate to API responses:
  - fsspec/dropbox.py — return ProviderError(e.body) / UserError(...) with e.body (raw HTTP response body)
  - fsspec/box.py — UserAuthError(e.message), message = e.message or e
  - fsspec/azure.py — UserAuthError(e.reason), message
  - fsspec/gcs.py:135 — message = e.message or e
  - ibm_watsonx_s3.py — url = e.request.url (full request URL incl. query params) in messages, ProviderError(e)

  Failure scenario: a Dropbox 4xx whose e.body echoes the request (auth header / presigned URL) → surfaced to the caller. Lower
  confidence than #1 — depends on each provider's body/URL contents — but it's the same vulnerability the PR set out to fix, sitting
  one branch over.

  3. MEDIUM/LOW — safe_error_summary's allowlist is name-based, not shape-based; a secret in an allowlisted attr passes through

  _MACHINE_CODE_RE = [A-Za-z0-9_.\-]{1,64} accepts most API-key/token shapes (sk-live-..., AWS keys, hex tokens). The only thing
  preventing a leak is the attribute name allowlist (code, status, error_code, request_id, …). If any SDK stashes a credential in one
  of those string attrs, safe_error_summary surfaces it verbatim. The design is otherwise "safe by construction"; this is the one hole.
  error.py:34 (the regex) / error.py:38-45 (_safe_error_value).

  Failure scenario: an SDK raising with .code = "sk-live-…" → emitted in the redacted message. Low probability, worth a comment noting
  the guarantee rests on the attr-name list.

paulkarayan:
### Note: merge with `main` + SharePoint decision

Merged current `main` (now at `1.6.30`) into this branch to clear the conflicts. Resolution notes:

- **Version → `1.6.31`.** `main` had advanced to `1.6.30` and, notably, *reused* `1.6.26` for a different change (Confluence date population). This branch's security entry was therefore moved out of `1.6.26` into a fresh `## [1.6.31]` CHANGELOG section; `1.6.26` now reflects main's Confluence entry.

- **SharePoint: kept identical to `main` — this PR no longer modifies `sharepoint.py`.** The redaction edits this PR originally applied to the SharePoint connector are dropped as superseded. `main`'s `1.6.29` work (`_handle_client_request_exception`) *deliberately* surfaces the real upstream HTTP status, a **truncated (≤500 char) response body**, and Microsoft correlation headers on the raised error — the opposite direction from this PR's "reduce to a sanitized summary" approach. That is the newer, intentional design, so we defer to it rather than re-layer redaction on top.

  ⚠️ **Security follow-up (not addressed here):** because SharePoint error messages now carry a truncated upstream response body, that body could echo sensitive content. This is `main`'s deliberate diagnostics trade-off; flagging it so the security team can decide whether the SharePoint body-surfacing needs to be reconciled with the redaction policy (e.g. scrub the body) in a separate change.

- `confluence.py` and `jira.py` auto-merged cleanly; their precheck guidance-preservation / `from None` changes from this branch are intact.

Verified after merge: `test_sharepoint`, `test_confluence`, `test_jira`, `test_error` unit tests pass; ruff clean across changed files.

potter-potter:
Let me know when this is ready and I'll bring it through into the index/download/upload/staging images

potter-potter (approved):
LG

PR #750, opened 2026-07-07
Introduces local RAG Ingest MCP server (rag-ingest-mcp) on top of unstructured-ingest that pairs with transform-mcp (supports Chroma, Qdrant, pgvector)

CodesLikeIcarus:
## What

A local **STDIO MCP server** (`unstructured_ingest/mcp/`, run as `python -m unstructured_ingest.mcp`) that turns Transform MCP output into a queryable local RAG corpus. It pairs with the remote Transform MCP: Transform parses/chunks/(optionally) embeds and returns Element JSON; this server fetches that JSON **out of band** (it never enters the agent's context), embeds it locally iff Transform didn't, upserts it into a local vector store, and serves matched-space similarity search.

Three tools: `load_transform_output` (URLs, file paths, or directories; `embed_policy` auto/passthrough/local), `search`, `list_collections`.

## Design

- **The embedding-space invariant.** A collection is pinned to one `(provider, model, dimension)` on first load; loads with a different model are refused (`SpaceMismatch`), and every query is embedded into the pinned space. Corpus/query symmetry is structural, not a discipline — mixing embedding models silently ruins retrieval, so the server makes it impossible.
- **Three backends behind one `VectorStore` contract** (`stores/`), selected by `URAG_STORE_BACKEND`:
  - `chroma` (default) — embedded, zero-config; pin in collection metadata
  - `qdrant` — embedded local mode or a server (`URAG_QDRANT_URL`); pin in a reserved meta collection
  - `pgvector` — psycopg 3; auto-provisions the table + cosine HNSW index (≤2000 dims) on first load; pins in a `rag_ingest_spaces` table
- **Connector-shaped writes.** Rows are conformed by the corresponding ingest stagers (`ChromaUploadStager`, `QdrantUploadStager`, `SQLUploadStager`), so what lands in each store matches what the ingest pipeline's own connector would produce, and element ids are deterministic (re-loading upserts, never duplicates).
- Backend client libraries import lazily — an install only needs the extras for the backend it uses.

## Testing

- **Conformance suite** (`test/unit/mcp/test_stores.py`) runs identically against every backend and is the working definition of "supported backend": load → cosine ranking → space-guard refusal → idempotent re-load → metadata round-trip. Chroma and Qdrant run embedded with no services; pgvector runs against `URAG_TEST_PG_DSN` (verified vs a dockerized `pgvector/pgvector:pg17`).
- **Server-level suite** (`test/unit/mcp/test_server.py`) drives the FastMCP server in-memory over both embedded backends: load/list, directory expansion, per-source space-guard errors.
- **Real end-to-end (all three backends):** Transform MCP upload → `chunk_by_character` job → `download_url` → `load_transform_output` (local embed, `text-embedding-3-small`/1536) → `search`. Three real files (PDF/MD/HTML, 92 chunks); three topical queries each returned the expected document as the top hit with identical scores across backends.

## Docs

`unstructured_ingest/mcp/README.md` plus `examples/` with a getting-started guide per backend: setup, docker-compose where a service is needed, the agentic loop with Transform MCP, CLI inspection, and when (not) to pair the DB vendor's own MCP server with a corpus built here.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/750?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lawrence-u10d:
## Security review

Nice, well-structured PR — the space-pinning invariant, the streaming byte cap, deterministic idempotent upserts, and the parameterized pgvector SQL are all carefully done. A few security notes, one worth addressing before merge.

### 1. SSRF: the out-of-band fetch has no destination guard (please address)

`fetch.py` fetches arbitrary `http(s)` URLs with no validation of the resolved destination:

```python
with httpx.stream("GET", url, timeout=_FETCH_TIMEOUT_S, follow_redirects=True) as resp:
```

There's no block on private / loopback / link-local / cloud-metadata address ranges, and `follow_redirects=True` runs with no per-hop re-validation (so a guard on just the initial URL would still be bypassable via a redirect).

Because `sources` is a tool argument, in an agentic loop over untrusted documents it is effectively attacker-influenced (prompt injection). A crafted `sources=["http://<internal-host>/..."]` makes this local server reach whatever the host machine can reach — localhost/LAN services, unauthenticated internal endpoints — pull any JSON-array response into the corpus, and surface it back through `search`. Impact is bounded by the single-user local context and by the fact that the body must parse as a JSON array, but it's a real SSRF primitive in a new fetch path.

**Suggestion:** resolve the hostname once, reject private/loopback/link-local/metadata ranges, connect pinned to the validated IP (preserving Host + TLS SNI), and re-validate on every redirect hop.

### 2. Arbitrary local file / directory read via tool args

`fetch_elements` reads any local path (`""`/`file://` → `path.read_bytes()`), and `_expand_sources` does a recursive `rglob("*.json")` on any supplied directory. Under the same prompt-injection model, `sources=["/"]` or a path to a sensitive JSON file harvests local content into a searchable corpus. Single-user context limits the blast radius, but consider constraining reads to a configured root and documenting that `sources` is security-sensitive input.

### 3. pgvector SQL — clean ✅ (one non-security footgun)

The SQL construction is solid: strict identifier regex (`_check_name`) on every entry point, `psycopg.sql.Identifier` for table/index names, `sql.Literal` for the dimension, and bound params for all data including vector literals and `LIMIT`. No injection surface.

Footgun (not security): `CREATE TABLE IF NOT EXISTS {collection}` + `INSERT ... ON CONFLICT (id)` means a collection name colliding with a pre-existing table in the target database silently adopts/writes into it. Consider schema-qualifying collections (e.g. a dedicated schema) or documenting "use a throwaway database."

### 4. Note: stored `text_as_html` in metadata

The stagers land engine-supplied `text_as_html` into row metadata, which `search` returns verbatim. Not an issue for this server (it never renders HTML), but any downstream consumer that renders search-result metadata as HTML would need to sanitize it — worth a one-line note in the README.

None of the above are blockers except #1, which I'd like to see guarded before merge given it's a brand-new outbound-fetch path. Happy to help with the SSRF-guard implementation if useful.

PR #751, opened 2026-07-09
feat(connectors): add url source connector

paulkarayan:
Part of making "playground" (the Test / workbench flow) run on the mainstream processing path instead of its own bespoke code.

Today playground ingests uploaded/sample files through a hidden, one-off plugin (`plugin-utic-playground`) with its own indexer + downloader. This adds a real `url` source connector so ingest can instead be a normal connector — "fetch this list of {url, filename}" — the same way every other source works.

**What it does**
- `UrlIndexer`: yields `FileData` for a configured list of `{url, filename}`.
- `UrlDownloader`: SSRF-safe fetch — validates every resolved address (`is_global` allowlist), pins the httpx transport to the validated IP, revalidates each redirect hop (no DNS-rebind TOCTOU). Private-IP guard is connector config (`allow_private_ips`), not an env var.
- Registered in the source registry; httpx is a lazy `url` extra. 14 unit tests.

Once released, playground's uploaded-file ingest becomes a plain `url` source — no special plugin.

**Linkages**
- platform-libs `pk/url-source-connector-type` declares this as a `SourceConnectorType` (needed for it to be usable in workflows).
- platform-plugins `pk/playground-notify` handles the uploader side of the same goal.
- Downstream: platform-api rewrites the playground DAG to emit this `url` source (separate PR, once this is released and plugin manifests reference it).

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/751?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulkarayan:
Addressed cubic:
- **P1 (path traversal):** fixed — `_safe_filename` reduces caller filenames to a basename and rejects pure-traversal names; tested.
- **Duplicate filenames:** fixed — `precheck` rejects duplicate basenames.
- **Coverage (default guard not exercised through the pipeline):** added `test_download_blocks_private_by_default` — runs `UrlDownloader.run` with `allow_private_ips=False` and asserts the fetch is blocked.
- **Streaming/memory (P2):** deferred — the only current caller is playground (20 MB cap), and streaming with per-redirect revalidation adds real complexity. Worth doing if this connector is used for large general downloads; filing as follow-up rather than in this PR.
- **Spike redirect test (P2):** the spike is a smoke script; the real per-hop revalidation coverage is `test_redirect_target_is_revalidated` (monkeypatches `_validate_and_pin`, `allow_private=False`, asserts the post-redirect host is validated).

aballman:
The SSRF design is right and I verified it: the socket pins to the validated IP with no connect-time re-resolution (no TOCTOU), every resolved record is validated, each redirect hop is revalidated, and Host/SNI stay on the real hostname. `allow_private_ips` defaults to deny.

Blocking:
- Lint is red on 3.11/3.12/3.13, and `changelog`/`check-version` are failing. Needs green CI and a CHANGELOG entry.
- The module docstring says "NOT YET REGISTERED ... follow-ups", but this PR adds the `add_source_entry("url", ...)` registration in `connectors/__init__.py`. Fix the docstring.
- Is `scripts/url_connector_spike.py` meant to ship, or is it a leftover spike?

Worth addressing:
- `is_global` alone misses IPv6 transition addresses: `64:ff9b::a9fe:a9fe` (NAT64), `2002:a9fe:a9fe::1` (6to4), and `::7f00:1` (IPv4-compatible) all embed a private/metadata IPv4 but return `is_global == True`. Only reachable behind NAT64/DNS64, but I'd reject reserved addresses or re-check the embedded IPv4. The same guard in platform-plugins#1668 has the identical gap, which is the argument for factoring `_validate_and_pin` into one shared util instead of two copies.
- `is_global`'s handling of CGNAT `100.64.0.0/10` and IPv4-mapped is Python-patch dependent (correct on >=3.11.9/3.12.4). This connector supports 3.11+, so either raise the floor or note the caveat.
- The response body is read fully into memory with no size cap. A hostile or oversized URL can OOM the worker.

paulkarayan:
Dropped `scripts/url_connector_spike.py` in 17685cf8 — throwaway spike; the connector is covered by `test/unit/connectors/test_url.py`.

paulkarayan:
Addressed the SSRF address-validation gaps (1fd48914):
- **IPv6 transition addrs:** `_is_public_address` now rejects IPv6 that embeds a private/metadata IPv4 — NAT64 `64:ff9b::/96`, 6to4 `2002::/16`, IPv4-mapped `::ffff:…`, IPv4-compatible `::a.b.c.d`. Tests cover NAT64/6to4/mapped/compat + a public IPv6.
- **CGNAT version-dependence:** rejected `100.64.0.0/10` explicitly, so it no longer depends on the Python patch level of `is_global`.

On the shared-util suggestion: I mirrored the same fix into platform-plugins#1668 rather than factor a cross-repo util, because that downloader is being retired by the playground collapse (its index/downloader are replaced by this `url` connector). Not worth a shared dependency for code that's going away.

aballman:
The IPv6 transition hardening looks right. `_is_public_address` extracts and re-checks the embedded IPv4 for NAT64/6to4/compat, the explicit CGNAT reject makes it Python-version-independent, and the tests cover all four cases (`64:ff9b::`, `2002::`, `::7f00:1`, `100.64/10`). Streaming to disk, the spike removal, and the docstring fix are all addressed too.

One low nit: `192.0.0.0/24` special-use (e.g. `192.0.0.9`) still passes `is_global`. It isn't a metadata or internal target, so I wouldn't hold on it. Otherwise this looks good to release.

paulkarayan:
changing to draft since i want to entertain migration to platform plugins

paulkarayan:
Superseded by platform-plugins#1751, which moves this url connector into a self-contained `url` plugin (playground's only consumer, hidden). Keeps unstructured-ingest untouched. The SSRF-safe fetcher, the two open review fixes here (temp-file prefix, NAT64 NSP), plus four further hardening fixes (allow_private_ips env-only, redacted non-retryable upstream errors, IPv4 address preference, per-entry identifiers) all live in #1751. Closing.

PR #753, opened 2026-07-10
fix(sharepoint): surface real upstream HTTP status in downloader instead of masking as "Site not found"

qued:
## Summary

`SharepointDownloader._fetch_file` caught every `ClientRequestException` from the site/file lookup and re-raised it as `SourceConnectionError("Site not found")`, discarding the underlying HTTP status and body. Every upstream failure — `401`, `403`, `404`, `429`, or `5xx` — therefore surfaced identically to users as `400 ... Site not found`, making auth / permission / throttle problems indistinguishable and undiagnosable. The retry classifier also matched on that already-rewritten string, so genuine `429` throttles never retried.

## Changes

- **Shared status→typed-error mapping.** Extracted the status handling into a module-level helper used by `SharepointIndexer` and `SharepointDownloader`: `401`/`403` → `UserAuthError`, `404` → `NotFoundError`, `429` → `RateLimitError`, `5xx` → `SourceConnectionNetworkError` (transient/connection-class, not a user fault). It logs the real `status_code`, response body, and Microsoft correlation headers (`request-id`, `client-request-id`, `x-ms-ags-diagnostic`, `SPRequestGuid`, `Retry-After`, `WWW-Authenticate`) **before** applying any label.
- **Pass the real status through on the error, not just the logs.** The raised error carries the true HTTP status code stamped on the instance (shadowing the typed class's default, so e.g. a `403` reports `403` rather than `UserAuthError`'s `401`, and an unmapped `4xx` reports its real code rather than `422`) and a truncated response body in its message — so the true condition is visible to callers, not only in logs.
- **Stop masking in the downloader.** `_fetch_file` no longer wraps with `@SourceConnectionNetworkError.wrap` (which coerced sibling typed errors back to `400`); the catch-all is replicated inline so genuinely unrecognized failures still surface as a `SourceConnectionNetworkError`, while typed errors carry their real status.
- **Indexer parity.** The indexer's site lookup, path resolution (`_get_target_drive_item`), and file listing (`get_files`) in `run_async` now route through the same shared helper instead of re-raising a generic `SourceConnectionError`, so index-time auth/throttle/not-found failures are classified consistently with the downloader.
- **Fix retry.** `retry_on_status_code` now reads the real status code stamped on the typed error (with type/string checks as fallback) instead of the masked message, so genuine `429`s and `503`s retry (matching the OneDrive downloader's throttle set) and then surface with their true status.

## Tests

- Downloader: real status passes through — `403` → `UserAuthError` reporting `403` (not `401`), an unmapped `4xx` (`409`) → `UserError` reporting its real code, `404` → `NotFoundError`, `429` → retries then `RateLimitError`, `5xx` → `SourceConnectionNetworkError` (not `UserError`) reporting `503`; `401` does not retry; an unknown non-HTTP exception → `SourceConnectionNetworkError` (the inline catch-all). The response body is carried on the error and truncated when long, and the real status + correlation headers are logged.
- Retry boundary: a retriable `503` followed by a non-retriable `403` stops on the `403` (two attempts) and surfaces it.
- Indexer `run_async`: site `401` → `UserAuthError`, `429` → `RateLimitError`, path-resolution `404` → `NotFoundError`, file-listing `429` → `RateLimitError`.
- Existing SharePoint + OneDrive unit tests pass (76 total); `ruff format`/`check` clean.

## Follow-ups (not in this PR)

- `UnstructuredIngestError.wrap` only re-raises the exact wrapped class, so a sibling typed error raised under **any** `.wrap` gets coerced to that class. A broader fix would let it pass through any `UnstructuredIngestError` (this would also fix the already-coerced `NotFoundError` in this file and OneDrive's `FileNotFoundError`).
- `run_async` makes blocking Office365 SDK calls (site resolution and `get_files().execute_query()`) on the event-loop thread. This predates this PR and is also present in the OneDrive path; wrapping them in `asyncio.to_thread` is orthogonal to this masking fix and best done as its own change.
- Classify the indexer's permissions batch fetch (`_fetch_permissions_raw`) — a raw `requests.post` to the Graph `$batch` endpoint (a different error surface: `requests` errors + per-item batch statuses) living in the shared `OnedriveIndexer` parent, so it needs its own classifier plus OneDrive coverage.
- Cache site resolution across files in a pipeline (currently one `get_by_url` per file).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Fixes
 
-- **fix(sharepoint): surface the real upstream HTTP status in the downloader instead of masking every error as "Site not found".** `SharepointDownloader._fetch_file` caught every `ClientRequestException` and re-raised it as `SourceConnectionError("Site not found")`, discarding the underlying HTTP status and body — so `401`/`403`/`404`/`429`/`5xx` all surfaced identically as a `400`, and the retry classifier (matching the rewritten message) never retried genuine `429` throttles. The status→typed-error mapping is now shared between the indexer and downloader (`401`/`403` → `UserAuthError`, `404` → `NotFoundError`, `429` → `RateLimitError`), preserves the status/body and Microsoft correlation headers for diagnosis before labeling, and `429`s and `503`s are now retried (matching the OneDrive downloader) and surfaced with their real status.
+- **fix(sharepoint): surface the real upstream HTTP status instead of masking every error as "Site not found".** SharePoint upstream errors now map to typed exceptions carrying the real status code and response body (instead of a generic `400`), and genuine throttles are retried.
 
 ## [1.6.28]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Fixes
 
-- **fix(sharepoint): surface the real upstream HTTP status instead of masking every error as "Site not found".** SharePoint upstream errors now map to typed exceptions carrying the real status code and response body (instead of a generic `400`), and genuine throttles are retried.
+- **fix(sharepoint): surface the real upstream HTTP status instead of masking every error as "Site not found".** SharePoint upstream errors now map to typed exceptions carrying the real status code and response body (instead of a generic `400`), and genuine throttles are retried, honoring the server's `Retry-After` backoff.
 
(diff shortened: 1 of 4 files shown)


potter-potter (approved):
LG

PR #760, opened 2026-07-16
fix(security): redact MotherDuck token from precheck connection errors

paulkarayan:
## Summary

Follow-up to #749 ([FIE-197](https://linear.app/unstructured/issue/FIE-197)). `MotherDuckUploader.precheck` overrode the base precheck that #749 fixed, so it still logged (with an `exc_info` traceback) and raised the caught exception — leaking `md:?motherduck_token=<token>` from the connection string into both logs and the API response. Routes the error through `safe_error_summary`, drops `exc_info`, and re-raises `from None`, matching the base `sql` precheck idiom; dependency-install hints and connector-authored typed errors are re-raised unchanged.

## Verification

- `ruff check` clean (pinned 0.15.1); module imports.
- End-to-end check: a token in the driver exception is absent from both the raised message and the logs.
- codex + cubic: no issues.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/760?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.7.2]
+
+### Fixes
+
+- **fix(security): redact MotherDuck token from precheck connection errors.** `MotherDuckUploader.precheck` now routes caught exceptions through `safe_error_summary` and drops `exc_info`, so a bad-token `OperationalError` no longer leaks `md:?motherduck_token=<token>` into logs or the API response.
+
 ## [1.7.1]
 
 ### Fixes
--- a/test/unit/connectors/motherduck/test_motherduck.py
+++ b/test/unit/connectors/motherduck/test_motherduck.py
@@ -1,0 +1,46 @@
+import io
+import logging
+from contextlib import contextmanager
+from unittest.mock import patch
+
+import pytest
+
+from unstructured_ingest.error import DestinationConnectionError
+from unstructured_ingest.logger import logger
+from unstructured_ingest.processes.connectors.duckdb.motherduck import (
+    MotherDuckAccessConfig,
+    MotherDuckConnectionConfig,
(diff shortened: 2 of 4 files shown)


awalker4 (approved):

awalker4 (approved):
Clean adoption of the #749 idiom — ordering the ImportError/typed re-raise ahead of the redacted branch is exactly right, so dependency-install hints and connector-authored errors pass through untouched and only genuinely unexpected exceptions get summarized.

One thing to keep an eye on across the whole set (not a blocker here): `from None` plus dropping `exc_info` orphans the cause entirely, so for a bare driver exception carrying no allowlisted attrs the summary can collapse to just the type name. That's a fine trade for closing the concrete `motherduck_token` leak; worth revisiting only if we later decide ingest operator logs are a trusted debug sink rather than an exposed surface.

paulkarayan:
> One thing to keep an eye on across the whole set (not a blocker here): `from None` plus dropping `exc_info` orphans the cause entirely, so for a bare driver exception carrying no allowlisted attrs the summary can collapse to just the type name. That's a fine trade for closing the concrete `motherduck_token` leak; worth revisiting only if we later decide ingest operator logs are a trusted debug sink rather than an exposed surface.

this makes a ton of sense. agree re the right approach for now. also note that we could split the sinks in the future vs. a full on change. but that's a design decision for later.

PR #761, opened 2026-07-16
fix(security): redact provider exceptions from missed connector error sinks

paulkarayan:
## Summary

Follow-up to #749 ([FIE-197](https://linear.app/unstructured/issue/FIE-197)). Closes the sibling connectors #749 missed for both log and response sinks:
- **Logs / precheck:** VastDB and DuckDB `precheck` overrides adopt the base `sql` idiom; Databricks/Zendesk/S3 unhandled-exception logs route through `safe_error_summary` with `exc_info` dropped.
- **`wrap_error` returns:** the Zendesk/S3/Databricks classified branches route returned messages through `safe_error_summary` instead of raw `e`/`str(e)`/provider text (originally opened as #766, now **folded in here** so these three connectors ship as one PR).
- **Correctness:** Zendesk now classifies HTTP 500 as `ProviderError` (was `> 500`, which dropped 500 to a raw `return e`) — flagged by cubic.

The "File not found" guidance strings and the unclassified `return e` escape hatches are left intact.

> Part of the FIE-197 redaction follow-up set. Bumped to 1.7.1 + CHANGELOG entry; redaction unit tests added (guarded with `pytest.importorskip` where a connector SDK is not in the unit env, matching `test_sftp.py`). The six parallel `## [1.7.1]` entries will conflict at merge — only the first merges cleanly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.7.3]
+
+### Fixes
+
+- **fix(security): redact provider exceptions from missed connector error sinks.** VastDB/DuckDB `precheck` overrides and the Databricks/Zendesk/S3 unhandled-exception logs no longer interpolate raw exceptions or dump `exc_info` tracebacks; the Zendesk/S3/Databricks `wrap_error` classified branches route returned messages through `safe_error_summary`; Zendesk now classifies HTTP 500 as `ProviderError`.
+
 ## [1.7.2]
 
 ### Fixes
--- a/test/unit/connectors/databricks/test_volumes.py
+++ b/test/unit/connectors/databricks/test_volumes.py
@@ -1,0 +1,75 @@
+import logging
+
+import pytest
+
+from unstructured_ingest.error import ProviderError, UserAuthError, UserError
+from unstructured_ingest.processes.connectors.databricks.volumes_native import (
+    DatabricksNativeVolumesAccessConfig,
+    DatabricksNativeVolumesConnectionConfig,
+)
+
+SECRET = "SECRETpassword=hunter2 key=AKIAEXAMPLE"
+
+
+def _connection_config() -> DatabricksNativeVolumesConnectionConfig:
+    return DatabricksNativeVolumesConnectionConfig(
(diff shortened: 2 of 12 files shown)


awalker4 (approved):

awalker4 (approved):
The log redaction is right and the duckdb/vastdb precheck rewrites match the base idiom.

Main thing: this is only half of the fix for databricks/s3/zendesk. The classified branches still return the raw exception into the typed error (`str(UserAuthError(e)) == str(e)`), and since the caller does `raise self.wrap_error(e)` that raw text reaches the response via `status_code_text` — the more-exposed surface than the logs this PR closes. #766 is what fixes that, and it's stacked on this branch. They close one hole together, so I'd land them as a pair (or fold #766 in) rather than merge this alone as "logging-leak done" — otherwise main briefly reads as fixed while the connector still leaks the same secret to the API response.

Separate boundary note, not this PR's to fix: the unclassified `return e` fallbacks still surface raw `str` through the platform wrapper's generic branch (api_generator `status_code_text`). That's the platform-side backstop the ownership model wants (non-canonical → generic 500), which the deployed wrapper doesn't do yet.

paulkarayan:
> The log redaction is right and the duckdb/vastdb precheck rewrites match the base idiom.
> 
> Main thing: this is only half of the fix for databricks/s3/zendesk. The classified branches still return the raw exception into the typed error (`str(UserAuthError(e)) == str(e)`), and since the caller does `raise self.wrap_error(e)` that raw text reaches the response via `status_code_text` — the more-exposed surface than the logs this PR closes. #766 is what fixes that, and it's stacked on this branch. They close one hole together, so I'd land them as a pair (or fold #766 in) rather than merge this alone as "logging-leak done" — otherwise main briefly reads as fixed while the connector still leaks the same secret to the API response.
> 
> Separate boundary note, not this PR's to fix: the unclassified `return e` fallbacks still surface raw `str` through the platform wrapper's generic branch (api_generator `status_code_text`). That's the platform-side backstop the ownership model wants (non-canonical → generic 500), which the deployed wrapper doesn't do yet.

i will add this to the linear ticket and follow up.

PR #762, opened 2026-07-16
fix(security): redact provider exceptions from Teradata error logs

paulkarayan:
## Summary

Follow-up to #749 ([FIE-197](https://linear.app/unstructured/issue/FIE-197)). The seven `logger.error(f"...{e}", exc_info=True)` sites in the Teradata connector (source/destination precheck, table probe, create table, read schema, delete, insert batch) dumped the raw driver exception plus a full traceback into logs; the Teradata Go driver embeds connection details in its messages, so these leaked host/user/password text. Each log now routes through `safe_error_summary` with `exc_info` dropped. The raised errors already went through the connector's safe `_summarize_error` / `_raise_classified_teradata_error` helpers and are unchanged.

## Open design question (not changed here)

This connector **intentionally** preserves raw driver text on some paths — `_raise_classified_teradata_error` surfaces `Teradata reported: {exc}` for user-fault codes and chains `from exc` so tracebacks stay in logs (see its docstring; authored in #722 by Simon <simon@unstructured.io>). That predates the redaction policy and conflicts with it. Recommendation: switch `from exc` -> `from None` (stops traceback re-exposure) while keeping the user-fault message. Deferred to the connector owner (Simon, #722) rather than silently rewritten.

## Verification

- `ruff check` clean (pinned 0.15.1); **99 unit tests pass**.
- codex + cubic: no issues.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/762?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.7.8]
+
+### Fixes
+
+- **fix(security): redact provider exceptions from Teradata error logs and user-fault responses.** The seven `logger.error(..., exc_info=True)` sites route through `safe_error_summary` with `exc_info` dropped; `_raise_classified_teradata_error` no longer interpolates the raw driver message into the user-fault `UserError` (`Teradata reported: {exc}`) — it surfaces only the numeric code + fixed descriptor — and both classified raises use `from None`, so the driver text (host/user/password) can't reach the response or resurface via traceback logging.
+
 ## [1.7.7]
 
 ### Fixes
--- a/test/unit/connectors/sql/test_teradata.py
+++ b/test/unit/connectors/sql/test_teradata.py
@@ -1,3 +1,4 @@
+import logging
 from unittest.mock import MagicMock
 
 import pandas as pd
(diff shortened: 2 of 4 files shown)


awalker4 (changes requested):
The log redaction is right, and `_summarize_error` is already safe (it only ever returns controlled strings, never `str(raw)`), so the connection-error precheck path is fine.

The blocker is the classified user-fault path. `_raise_classified_teradata_error` still interpolates the raw driver exception into the response:

    raise UserError(f"Teradata error {code} ({descriptor}){target}. Teradata reported: {exc}") from exc

`{exc}` is the same Go-driver text this PR's own premise says embeds host/user/password, and it lands on the response surface — more exposed than the logs we just fixed. This is reachable (e.g. 3523 no-privilege on the source-table probe). Worth flagging that the proposed `from exc` → `from None` change wouldn't close it: the credential is in the message string, independent of the chaining.

The safe head is already built — `code`, `descriptor` (static map), and `table` are all safe and genuinely debuggable — so dropping the `. Teradata reported: {exc}` tail (or swapping it for `safe_error_summary(exc)`) closes the leak while keeping the useful part. Since it's a known credential leak on the response path and the fix is one line in a file you're already touching, I'd fold it in here rather than hand the whole thing to #722. Happy to leave just the `from exc`/`from None` chaining decision to Simon as the genuine design question.

awalker4 (approved):

PR #763, opened 2026-07-16
fix(security): redact provider text from fsspec/watsonx wrap_error branches

paulkarayan:
## Summary

Follow-up to #749 ([FIE-197](https://linear.app/unstructured/issue/FIE-197)). #749 redacted only the fallback branches of these `wrap_error` methods; the classified 4xx/5xx branches still interpolated raw provider free-text into the typed errors returned to API callers — dropbox `e.body`/`e.error`, box `e.message`, azure `e.reason`, gcs `e.message`, and watsonx raw `e` plus the full request URL (query string can carry tokens/signatures). Routes the message payloads through `safe_error_summary` and strips the query string from the watsonx log URL. The dropbox 4xx/5xx if/else collapsed since both arms now return the same summary.

**Correctness:** watsonx now classifies HTTP 500 as `ProviderError` (was `> 500`, which dropped 500 to a raw `return e`) — flagged by cubic (P1).

gcs's `Forbidden`/`Bad Request`/`File-not-found` branches and dropbox's `RateLimitError` are left as connector-authored guidance.

> Part of the FIE-197 redaction follow-up set. Bumped to 1.7.1 + CHANGELOG entry; redaction unit tests added (guarded with `pytest.importorskip` where a connector SDK is not in the unit env, matching `test_sftp.py`). The six parallel `## [1.7.1]` entries will conflict at merge — only the first merges cleanly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.7.5]
+
+### Fixes
+
+- **fix(security): redact provider text from fsspec/watsonx wrap_error classified branches.** Dropbox/Box/Azure/GCS/watsonx `wrap_error` route classified-branch messages through `safe_error_summary` instead of raw provider text (`e.body`/`e.message`/`e.reason`/full request URL); watsonx strips the URL query string from logs and now classifies HTTP 500 as `ProviderError` (was `> 500`).
+
 ## [1.7.4]
 
 ### Enhancements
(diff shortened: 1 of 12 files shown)


awalker4 (approved):
Routing the classified branches through `safe_error_summary` is consistent with the rest, and stripping the watsonx query string is a good catch (presigned signatures/tokens).

Two non-blocking nits while you're in here: `if response_code > 500` misses exactly 500, which then falls through to the raw `return e` — `>= 500` closes that. And dropbox's `AuthError` branch still `raise`s where its siblings `return` — pre-existing, just inconsistent.

PR #764, opened 2026-07-16
fix(security): redact raw exceptions from Salesforce/Chroma non-precheck raises

paulkarayan:
## Summary

Follow-up to #749 ([FIE-197](https://linear.app/unstructured/issue/FIE-197)). Folds in the two non-precheck raise sites flagged but deferred during #749 — Salesforce private-key parsing (the crypto exception can echo key material) and Chroma upsert (raw driver exception). Both now route through `safe_error_summary` and re-raise `from None`, matching the #749 chain-suppression idiom.

## Verification

- `ruff check` clean (pinned 0.15.1); both modules import.
- codex + cubic: no issues.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/764?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.7.6]
+
+### Fixes
+
+- **fix(security): redact raw exceptions from Salesforce/Chroma non-precheck raises.** Salesforce private-key parse failures and Chroma upsert failures route the caught exception through `safe_error_summary` and re-raise `from None`, so key material / driver text no longer surfaces in the raised error or its chained traceback.
+
 ## [1.7.5]
 
 ### Fixes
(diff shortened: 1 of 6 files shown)


awalker4 (approved):
Both good. The salesforce private-key path especially — a crypto exception can echo key material, so summarizing it and re-raising `from None` is the right move. Matches the idiom.

PR #765, opened 2026-07-16
docs(security): note safe_error_summary allowlist rests on attr names

paulkarayan:
## Summary

Follow-up to #749 ([FIE-197](https://linear.app/unstructured/issue/FIE-197)). Documents the one residual hole the review flagged: `safe_error_summary` is safe by construction of the `_SAFE_ERROR_ATTRS` **name** allowlist, not by value shape — `_MACHINE_CODE_RE` would pass a credential shaped like a machine code (`sk-live-...`, an AWS key, a hex token). The added comment records that the guarantee rests on keeping the attr list limited to fields that never carry free text. Comment-only; no runtime change.

The `wrap()` sibling-family branch (formats an already-sanitized `UnstructuredIngestError` message rather than a summary) is intentional and pinned by #749's guard test, so it is left unchanged.

## Verification

- `ruff check` clean (pinned 0.15.1); **14 error unit tests pass**.
- codex + cubic: no issues.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/765?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.7.1]
+
+### Fixes
+
+- **docs(security): document that `safe_error_summary`'s allowlist rests on attribute names.** Adds a comment in `error.py` recording that `_MACHINE_CODE_RE` is a shape filter, not a secret detector, so the guarantee depends on keeping `_SAFE_ERROR_ATTRS` limited to fields that never carry free text. Comment-only; no runtime change.
+
 ## [1.7.0]
 
 ### Enhancements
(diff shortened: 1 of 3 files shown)


awalker4 (approved):
Good to have this written down. The allowlist being safe by construction of the attr *names* rather than the shape regex is the honest bound, and stating that keeping `_SAFE_ERROR_ATTRS` free-text-free is the actual invariant is exactly the right framing.

PR #766, opened 2026-07-16
fix(security): redact provider text from Zendesk/S3/Databricks wrap_error returns

paulkarayan:
## Summary

Follow-up to #749 ([FIE-197](https://linear.app/unstructured/issue/FIE-197)). Companion to `pk/redact-connector-logging-leaks`: those `wrap_error` methods also returned typed errors carrying the raw provider exception into the API response — Zendesk `UnstructuredIngestError(str(e))` plus its classified `UserAuthError`/`RateLimitError`/`UserError`/`ProviderError(e)`, S3's provider `Error.Message`, and Databricks' classified branches. Routes each through `safe_error_summary`, mirroring the fsspec/watsonx classified-branch fix. The "File not found" guidance string and the unclassified `return e` escape hatches are left intact.

## Stacking

Based on `pk/redact-connector-logging-leaks` (same three files), so this PR targets that branch as its base. Merge that PR first, or fold this in.

## Verification

- `ruff check` clean (pinned 0.15.1); all 3 modules import.
- codex + cubic: no issues.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/766?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/unstructured_ingest/processes/connectors/databricks/volumes.py
+++ b/unstructured_ingest/processes/connectors/databricks/volumes.py
@@ -81,18 +81,18 @@ class DatabricksVolumesConnectionConfig(ConnectionConfig, ABC):
             if (message_split[0].endswith("auth")) or (
                 "Client authentication failed" in error_message
             ):
-                return UserAuthError(e)
+                return UserAuthError(safe_error_summary(e))
         if isinstance(e, DatabricksError):
             reverse_mapping = {v: k for k, v in STATUS_CODE_MAPPING.items()}
             if status_code := reverse_mapping.get(type(e)):
                 if status_code in [401, 403]:
-                    return UserAuthError(e)
+                    return UserAuthError(safe_error_summary(e))
                 if status_code == 429:
-                    return RateLimitError(e)
+                    return RateLimitError(safe_error_summary(e))
                 if 400 <= status_code < 500:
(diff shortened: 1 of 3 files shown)


paulkarayan:
Folded into #761 — the Zendesk/S3/Databricks `wrap_error` return-redactions (plus the Zendesk HTTP-500 boundary fix) were cherry-picked onto pk/redact-connector-logging-leaks so the log-leak and return-leak fixes for those three connectors ship as one PR. Closing this stacked branch; no unique commits are lost.

awalker4 (approved):
This is the other half of #761 — closes the classified-branch return leaks on databricks/s3/zendesk, including dropping the `str(e)` default in the s3 `Error.Message` lookup. Reads correct.

Just needs to land after #761 (or fold the two together), since it targets that branch. The unclassified `return e` escape hatches remain the one residual, but that's the platform-wrapper boundary rather than this PR's to close.

PR #769, opened 2026-07-19
fix(stager): preserve single-record NDJSON partition output

CyMule:
## Summary

Fix blob-storage staging when a partition produces an NDJSON file containing exactly one record.

The Google Drive E2E test exposed this because `foo.txt` produces a single partition element. The stager passed the NDJSON file to a streaming JSON-array parser, which treated the record as the root object and silently emitted zero elements.

## Changes

- Stream `.ndjson` inputs with an NDJSON-aware reader.
- Keep the existing streaming path for JSON-array inputs.
- Add regression coverage for both single-record and multi-record NDJSON.
- Update the Google Drive expected fixtures with the `drive_id` field introduced by #768.

## Root cause

The streaming implementation added in #735 expected a top-level JSON array. Single-record NDJSON is instead a single root object, so the parser returned no items. Multi-record NDJSON happened to trigger the existing fallback path, which is why the original unit test passed.

## Verification

- `uv run --locked --no-sync pytest -q test/unit/processes/utils/test_blob_storage.py`
- `uv run --locked --no-sync ruff check unstructured_ingest/processes/utils/blob_storage.py test/unit/processes/utils/test_blob_storage.py`
- `uv run --locked --no-sync ruff format --check unstructured_ingest/processes/utils/blob_storage.py test/unit/processes/utils/test_blob_storage.py`

The credentialed Google Drive E2E test was not run locally.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/769?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.7.11]
+
+### Fixes
+
+- **fix(stager): preserve single-record NDJSON partition output.** Blob-storage staging now reads `.ndjson` inputs with an NDJSON-aware streaming reader instead of treating them as top-level JSON arrays, preventing single-record partition output from being silently dropped. Adds regression coverage for single- and multi-record inputs and updates the Google Drive expected output with the `drive_id` field introduced in 1.7.10.
+
 ## [1.7.10]
 
 ### Fixes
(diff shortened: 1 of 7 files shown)


potter-potter (approved):
LG

PR #771, opened 2026-07-20
Refresh downloads when the remote source is newer

CyMule:
## Summary

- compare local and remote modification times in the correct direction
- accept both epoch and ISO-formatted remote timestamps
- treat timezone-naive ISO timestamps as UTC
- add regression coverage for newer, equal, older, malformed, and timezone-sensitive values

## Why

The freshness comparison was reversed, and ISO timestamps were not parsed. A newer remote document could therefore leave an older local download in place without an error.

## Impact

Cached files are refreshed when the remote source is newer. Missing or unparseable modification timestamps retain the existing conservative download behavior.

## Validation

- `uv run --locked --no-sync pytest -q test/unit/pipeline/test_download.py` — 16 passed
- Ruff check and format check passed for all changed Python files
- changelog and package versions both set to 1.7.11

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/771?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - **Refresh cached downloads when the source is newer.** Download freshness checks now compare local and remote modification times in the correct direction and accept both epoch and ISO-formatted source timestamps.
 - **Stamp downloaded files with the source modification time.** Downloaded files take their modification time from the source whenever it is an epoch or ISO-formatted timestamp, even when the source creation time is missing or unparseable, so freshness checks compare against the remote time rather than the download time.
+- **Report Databricks Volumes modification times in epoch seconds.** The Databricks SDK reports modification times in milliseconds; they are now converted to seconds so freshness checks no longer treat every indexed file as newer than its local copy.
 
 ## [1.7.10]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.7.11]
+
+### Fixes
+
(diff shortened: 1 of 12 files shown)


dizon-u10d (approved):
Looks good to me. I did not find any open changes-requested blockers to carry forward; the existing Cubic reviews were comments only. The shared timestamp parser now handles epoch and ISO values, rejects non-finite values, treats naive ISO timestamps as UTC, and is used consistently by both the freshness check and download response stamping. The Databricks Volumes millisecond-to-seconds conversion is covered by a focused unit test and the expected integration fixture output was updated.

Validation I ran locally:
- `uv run --locked --group test pytest -q test/unit/test_utils.py::test_parse_timestamp_handles_epoch_and_iso_values test/unit/test_utils.py::test_local_timezone_does_not_shift_iso_timestamps test/unit/interfaces/test_downloader.py test/unit/pipeline/test_download.py test/unit/connectors/databricks/test_volumes.py::test_indexed_file_reports_modification_time_in_epoch_seconds` - 27 passed, 1 skipped
- `uv run --locked --group lint ruff check ...` on the changed Python files - passed
- `uv run --locked --group lint ruff format --check ...` on the changed Python files - passed

PR #772, opened 2026-07-20
Rewind files before NDJSON fallback parsing

CyMule:
## Summary

- rewind the input stream before NDJSON fallback parsing
- add a regression test using valid two-record NDJSON with an unrecognized extension

## Why

Format detection reads from the file before the fallback parser iterates its lines. Without rewinding, valid NDJSON can be parsed from end-of-file and return zero records without an error.

## Impact

The fallback now reads from the beginning of the stream. Recognized JSON and NDJSON paths are unchanged.

## Validation

- `uv run --locked --no-sync pytest -q test/unit/utils/test_data_prep.py` — 1 passed
- Ruff check and format check passed for all changed Python files
- changelog and package versions both set to 1.7.11

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/772?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.7.11]
+
+### Fixes
+
+- **Preserve NDJSON records after format detection.** The fallback parser now rewinds the input stream before reading line-delimited JSON, preventing valid records from being skipped after the detection pass.
+
 ## [1.7.10]
 
 ### Fixes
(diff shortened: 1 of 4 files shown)


dizon-u10d (approved):
Looks good to me. The fallback parser now rewinds after the failed JSON probe before trying NDJSON, which matches the failure mode in the PR description, and the new test covers the unrecognized-extension path that previously returned no records.

PR #774, opened 2026-07-20
Apply tar extraction filters on supported Python runtimes

CyMule:
## Summary

- detect the existing `tarfile.tar_filter` capability instead of gating it on the Python minor version
- apply the current tar extraction policy on supported Python 3.11 maintenance releases as well as Python 3.12+
- preserve the existing warning-and-extract behavior when extraction filters are unavailable
- add focused coverage for benign archives, path traversal rejection, and the compatibility fallback

## Why

Tar extraction filtering is currently enabled only when `sys.version_info >= (3, 12)`. Extraction filters are also available on maintained Python 3.11 patch releases supported by this package, so the version gate can leave available protection unused.

Capability detection applies the mitigation wherever the runtime provides it without changing the filter policy or breaking older Python 3.11 patch releases still permitted by the package metadata.

This draft intentionally does **not** switch from `tar_filter` to the stricter `data_filter`, fail closed on older patch releases, or change `requires-python`; those would be separate compatibility and product decisions.

## Impact

On Python runtimes exposing `tarfile.tar_filter`, archive extraction uses the same filter already selected for Python 3.12+. Runtimes without the API retain their current behavior and receive an updated warning recommending Python 3.11.4 or later.

## Validation

- `uv run --locked --no-sync pytest -q test/unit/utils/test_compression.py` — 3 passed
- `uv run --locked --no-sync ruff check unstructured_ingest/utils/compression.py test/unit/utils/test_compression.py`
- `uv run --locked --no-sync ruff format --check unstructured_ingest/utils/compression.py test/unit/utils/test_compression.py`

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/774?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.7.11]
+
+### Fixes
+
+- **fix(security): apply tar extraction filters on supported runtimes.** Tar extraction now detects the existing `tarfile.tar_filter` capability instead of gating it on the Python minor version, so supported Python 3.11 maintenance releases receive the same filtering already used on Python 3.12+. Older patch releases retain the existing warning-and-extract behavior.
+
 ## [1.7.10]
 
 ### Fixes
--- a/test/unit/utils/test_compression.py
+++ b/test/unit/utils/test_compression.py
@@ -1,0 +1,57 @@
+import io
+import logging
+import tarfile
+from pathlib import Path
+
+import pytest
+
+from unstructured_ingest.utils.compression import uncompress_tar_file
+
+
+def _write_tar_member(archive_path: Path, member_name: str, content: bytes) -> None:
+    with tarfile.open(archive_path, "w") as archive:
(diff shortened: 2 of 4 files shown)


dizon-u10d (approved):
I didn't find any blocking issues here. The capability check preserves the existing tar filter policy while enabling it on runtimes that expose `tarfile.tar_filter`, the fallback behavior is unchanged apart from the clearer warning, and the new tests cover benign extraction, traversal rejection when filters exist, and the no-filter compatibility path.

PR #775, opened 2026-07-21
fix(slack): Account for `xoxe.` prefix when classifying slack token

ds-filipknefel:
Slack token is prefixed with `"xoxe."` when obtained via refresh - as referenced in the documentation examples https://docs.slack.dev/authentication/using-token-rotation/#refresh. Update the token classification function to take the prefix into account.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## [1.7.14]
+
+### Fixes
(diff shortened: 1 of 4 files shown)


mateuszkuprowski (approved):
That's exactly how it should be, let's go with it!

potter-potter (approved):

PR #777, opened 2026-07-27
bug(PLU-526): user error classification for postgres

potter-potter:
A user's ngrok tunnel to their private Postgres keeps dropping mid-run.
`precheck()` already wraps connection errors, but `SQLUploader.upload_dataframe`
called `get_cursor()` bare — so a `psycopg2.OperationalError` at write time
bubbled up unwrapped and the plugin host mapped it to a **500** (platform fault,
retryable). It's a customer-side outage, not ours.
### Change
- Wrap the write loop the same way `precheck()` does: pass through
  `ImportError` / typed `UnstructuredIngestError`, otherwise re-raise as
  `DestinationConnectionError` (status_code 400 → 4xx/user, non-retryable).
- Regression tests: raw driver error → `DestinationConnectionError(400)`;
  typed ingest error passes through unwrapped.
### Scope
Base `SQLUploader` covers postgres, sqlite, and singlestore. The same latent
gap in snowflake / vastdb / duckdb / motherduck is deferred to a follow-up
hardening PR; teradata already handles it.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/777?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Fixes
 
-- **Surface mid-upload SQL destination failures as connection errors.** SQL uploaders only wrapped connection failures in `precheck()`; a destination reachable at precheck could still drop during the write loop (e.g. a customer's tunnel to a private database going down), letting a raw driver exception surface as a 500/platform fault. `upload_dataframe` now re-raises such failures as `DestinationConnectionError` (matching `precheck()`), so they are classified as user/downstream faults. Applied to the base SQL uploader (postgres, sqlite, singlestore) and the snowflake, vastdb, duckdb, and motherduck uploaders; teradata already had equivalent handling.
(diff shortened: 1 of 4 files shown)


dizon-u10d (approved):
Looks good to me. I verified the shared SQLUploader path now preserves ImportError and existing UnstructuredIngestError subclasses, while unexpected mid-upload failures are re-raised as DestinationConnectionError with the existing sanitized safe_error_summary behavior, matching the precheck classification path.

I also checked the version/changelog bump and ran the focused SQL unit tests successfully. Local make check-version is blocked on this machine by the repo script requiring GNU sed >= 4.3, but the PR's GitHub check-version job is green for this head.

PR #778, opened 2026-07-28
feat(PLU-511): ACL digest (permissions_version) for OneDrive/SharePoint

danielle-unstructured-io:
## What

Adds a per-file **ACL digest** (`permissions_version`) so an ACL-only change can trigger reprocessing under incremental (`reprocess_all = false`), following the mechanism recommended in the [spike](https://app.notion.com/p/3a02c3765a0a805da24ce23761ceed82). This is the connector-layer slice of PLU-511; it pairs with the invoker changes in platform-plugins and the compare/write-back in platform-etl-orchestration (#1250).

- `utils/acl.py` — `compute_permissions_version(permissions_data, denied_permissions_data=None)`: canonicalizes (recursive sort) then `sha256`. Returns `None` only when there is no ACL signal at all; an empty list is a real state (all access revoked) and yields a stable digest.
- `data_types/file_data.py` — new `permissions_version` field on `FileDataSourceMetadata`.
- `processes/connectors/onedrive.py` — the indexer computes the digest over the permissions it already batch-fetches (SharePoint inherits this base class). Adds call-count telemetry proving the digest issues **0 additional Graph calls**.
- `test/unit/utils/test_acl.py` — determinism / order-independence / empty-vs-None behavior.

## Validated

Proven end to end on an SND against a real SharePoint site: an ACL-only change (content unchanged) reprocesses exactly the changed file (spike scenarios S2/S8). See PLU-511 for evidence.

## Notes

- **Draft.** Branch is behind `main` and needs a rebase before merge.
- Known limitation surfaced during validation: SharePoint `siteGroup` ACL changes are not captured (pre-existing PLU-370 `_extract_identity_ids_from_raw` exclusion), so a site-group-only change won't move the digest. Documented in the spike; follow-up TBD.

PLU-511

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/778?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test/unit/connectors/test_onedrive.py
+++ b/test/unit/connectors/test_onedrive.py
@@ -350,9 +350,7 @@ class TestDriveItemToFileDataSync:
                 "grantedToV2": {"user": {"id": "user-1"}},
             }
         ]
-        file_data = indexer.drive_item_to_file_data_sync(
-            drive_item, raw_permissions=raw_perms
-        )
+        file_data = indexer.drive_item_to_file_data_sync(drive_item, raw_permissions=raw_perms)
         assert file_data.metadata.permissions_data is not None
         assert len(file_data.metadata.permissions_data) == 3
 
@@ -362,13 +360,24 @@ class TestDriveItemToFileDataSync:
         file_data = indexer.drive_item_to_file_data_sync(drive_item)
         assert file_data.metadata.permissions_data is None
 
-    def test_permissions_none_when_empty_list_passed(self):
-        # empty list (e.g. 403 fall-back) leaves the field as None rather than
-        # writing an all-empty placeholder
+    def test_permissions_captured_when_empty_list_passed(self):
+        # Empty list is a genuine "all access revoked" state (distinct from a
+        # failed fetch, which is now None): it produces permissions_data plus a
+        # stable digest so revocation is detected (S9).
         indexer = _make_indexer()
         drive_item = _make_drive_item()
         file_data = indexer.drive_item_to_file_data_sync(drive_item, raw_permissions=[])
+        assert file_data.metadata.permissions_data is not None
+        assert file_data.metadata.permissions_version is not None
+
+    def test_permissions_none_when_fetch_unavailable(self):
+        # None means the permission fetch was unavailable this run; leave the
+        # ACL fields unset rather than fabricating a revocation.
+        indexer = _make_indexer()
+        drive_item = _make_drive_item()
+        file_data = indexer.drive_item_to_file_data_sync(drive_item, raw_permissions=None)
         assert file_data.metadata.permissions_data is None
+        assert file_data.metadata.permissions_version is None
 
 
 def _batch_response(status_code: int = 200, responses: list[dict[str, Any]] | None = None) -> Mock:
@@ -430,7 +439,7 @@ class TestFetchPermissionsRaw:
             "item-f2.docx": "c",
         }
 
-    def test_per_item_403_yields_empty_for_that_item(self):
+    def test_per_item_403_yields_none_for_that_item(self):
         indexer = _make_indexer()
         items = [_make_drive_item(f"f{i}.docx") for i in range(2)]
         body = _batch_response(
@@ -447,7 +456,19 @@ class TestFetchPermissionsRaw:
             result = indexer._fetch_permissions_raw(items, access_token="tok")
 
         assert result["item-f0.docx"][0]["grantedToV2"]["user"]["id"] == "a"
-        assert result["item-f1.docx"] == []
+        # A forbidden sub-response is "unavailable", not "no permissions": None,
+        # so it is never mistaken for a revocation.
+        assert result["item-f1.docx"] is None
+
+    def test_per_item_200_empty_value_is_empty_list(self):
+        # A 200 with an empty value list is a genuine "no permissions" state
+        # (all access revoked): represented as [] and kept distinct from None.
+        indexer = _make_indexer()
+        items = [_make_drive_item("f.docx")]
+        body = _batch_response(responses=[{"id": "0", "status": 200, "body": {"value": []}}])
+        with patch("requests.post", return_value=body):
+            result = indexer._fetch_permissions_raw(items, access_token="tok")
+        assert result == {"item-f.docx": []}
 
     def test_envelope_401_raises_user_auth_error(self):
         indexer = _make_indexer()
@@ -466,7 +487,7 @@ class TestFetchPermissionsRaw:
         body.text = "internal server error"
         with patch("requests.post", return_value=body):
             result = indexer._fetch_permissions_raw(items, access_token="tok")
-        assert result == {"item-f0.docx": [], "item-f1.docx": []}
+        assert result == {"item-f0.docx": None, "item-f1.docx": None}
 
     def test_429_triggers_retry_then_succeeds(self):
         indexer = _make_indexer()
@@ -481,8 +502,10 @@ class TestFetchPermissionsRaw:
                 }
             ]
         )
-        with patch("requests.post", side_effect=[throttled, success]) as mock_post, \
-             patch("tenacity.wait_exponential.__call__", return_value=0):
+        with (
+            patch("requests.post", side_effect=[throttled, success]) as mock_post,
+            patch("tenacity.wait_exponential.__call__", return_value=0),
+        ):
             result = indexer._fetch_permissions_raw(items, access_token="tok")
         assert mock_post.call_count == 2
         assert result["item-f.docx"][0]["grantedToV2"]["user"]["id"] == "x"
@@ -490,14 +513,16 @@ class TestFetchPermissionsRaw:
     def test_network_errors_exhaust_retries_then_degrade(self):
         indexer = _make_indexer()
         items = [_make_drive_item(f"f{i}.docx") for i in range(2)]
-        with patch(
-            "requests.post",
-            side_effect=requests.exceptions.ConnectionError("dns down"),
-        ) as mock_post, \
-             patch("tenacity.wait_exponential.__call__", return_value=0):
+        with (
+            patch(
+                "requests.post",
+                side_effect=requests.exceptions.ConnectionError("dns down"),
+            ) as mock_post,
+            patch("tenacity.wait_exponential.__call__", return_value=0),
+        ):
             result = indexer._fetch_permissions_raw(items, access_token="tok")
         assert mock_post.call_count == 5  # stop_after_attempt(5)
-        assert result == {"item-f0.docx": [], "item-f1.docx": []}
+        assert result == {"item-f0.docx": None, "item-f1.docx": None}
 
     def test_malformed_sub_response_id_skipped(self):
         indexer = _make_indexer()
@@ -510,8 +535,9 @@ class TestFetchPermissionsRaw:
         )
         with patch("requests.post", return_value=body):
             result = indexer._fetch_permissions_raw(items, access_token="tok")
-        # malformed entries are skipped; default empty list remains
-        assert result == {"item-f.docx": []}
+        # malformed entries are skipped; the item never got a 200, so it stays
+        # None (unavailable), not [] (which would imply a genuine revocation)
+        assert result == {"item-f.docx": None}
 
 
 class TestSingletonBleedRegression:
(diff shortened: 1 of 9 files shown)


potter-potter:
Small suggestion: leave a breadcrumb comment about the SharePoint `siteGroup` gap at the digest computation site (`drive_item_to_file_data_sync`), so the next person extending PLU-511 sees that the digest's coverage is bounded by `permissions_data`.

It's a known, already-documented limitation (PLU-370) — not something to fix here — just worth making visible from the digest path:

```python
# NOTE: the digest inherits permissions_data's coverage. SharePoint siteGroup
# grants are excluded (see _extract_identity_ids_from_raw / PLU-370), so
# site-group-mediated ACL changes don't move the digest. Known limitation.
```

The existing `_extract_identity_ids_from_raw` docstring already captures the "why" (siteGroup numeric IDs aren't resolvable via Graph in the item-permissions call), so no need to duplicate that — this is just a pointer from the digest code.

potter-potter:
Not on the diff so I can't anchor it inline, but flagging on `sharepoint.py:335` (`SharepointIndexer.run_async` → `_flush`):

```python
raw_permissions=perms_by_id.get(di.id, []),
```

Minor, but worth tightening since this PR establishes the None-vs-`[]` contract:

- Under the new contract, `None` = "fetch unavailable → skip the digest" and `[]` = "genuine 200-empty → revoked → emit a real digest."
- `_fetch_permissions_raw` always returns an entry for every id (`None` when unavailable, `[]` only for a true revocation), so this default is never actually hit today.
- But if that dict ever became partial, a missing id would default to `[]` and **fabricate a revocation digest** — exactly the failure mode the `1dde29d4` fix set out to prevent. OneDrive's `run_async` was updated to `perms_by_id.get(drive_item.id)` (defaulting to `None`); SharePoint's still passes `[]`.

Suggest matching OneDrive: `raw_permissions=perms_by_id.get(di.id)` (drop the `[]`) so a missing id degrades to "skip digest" rather than "revoked."

unstructured_ingest/processes/connectors/onedrive.py
```diff
@@ -575,17 +587,36 @@ async def run_async(self, **kwargs: Any) -> AsyncIterator[FileData]:
...
+                if raw_permissions is not None:
+                    digests_computed += 1
+                items_indexed += 1
                 yield await self.drive_item_to_file_data(
                     drive_item=drive_item,
-                    raw_permissions=perms_by_id.get(drive_item.id, []),
+                    raw_permissions=raw_permissions,
                 )
 
+        logger.info(
```
potter-potter: N4 (SharePoint telemetry gap): this telemetry block only runs for OneDrive. `SharepointIndexer.run_async` overrides this method entirely (via its `_flush` helper) and has no equivalent, so a SharePoint index run emits **no** ACL-digest telemetry — even though SharePoint is the connector this POC exists to prove, and "marginal cost captured via call-count telemetry" is a named PLU-511 acceptance criterion.

To be clear, it's an evidence gap, not a functional one: SharePoint uses the same `_fetch_permissions_raw`, so the digest genuinely costs 0 extra Graph calls there too — you just can't see the proof in a SharePoint run's logs.

Suggestion: thread the same three counters (`permission_batch_calls` / `digests_computed` / `items_indexed`) through `SharepointIndexer._flush` and log once after the final flush. Count only the `_fetch_permissions_raw` batch calls as digest-attributable (not the site/drive-resolution calls). Even better would be a small shared helper used by both `run_async`s so the two paths can't drift again — the same override-drift that left `sharepoint.py:335` on the stale `[]` default.

Minor related nit (N5): the "0 additional API calls" sentence is a hardcoded string, not a measured value. If SharePoint gets its own copy, worth deriving it rather than asserting it.

danielle-unstructured-io: Done in 8e6804df. Pulled the three counters into a shared `_AclDigestTelemetry` helper that both OneDrive and SharePoint `run_async` use, so SharePoint now logs the same telemetry and the two paths can't drift. Only `_fetch_permissions_raw` batch calls are counted (record_batch), not the site/drive-resolution calls. For N5, reworded the log to state the digest "hashes already-fetched data and makes no Graph calls of its own" instead of asserting a hardcoded 0. Added a unit test for the counter (None -> no digest, [] -> digest, only batch calls counted).

danielle-unstructured-io:
Addressed the two non-anchored notes in 8e6804df:

- `sharepoint.py` `_flush` now uses `perms_by_id.get(di.id)` (defaults to None -> skip digest) instead of `[]`, matching OneDrive and the None-vs-`[]` contract from 1dde29d4, so a missing id can't fabricate a revocation.
- Added the siteGroup breadcrumb NOTE at the digest site in `drive_item_to_file_data_sync`, pointing to `_extract_identity_ids_from_raw` / PLU-370.

Thanks for the careful review.

awalker4:
A few things from a read-through, mostly around the new None-vs-empty contract plus some cleanup I'd want before this merges.

**`extract_permissions([])` returns `[{}]`**

Relaxing the guard to `is not None` makes the empty-list branch reachable for the first time, and `extract_permissions` short-circuits a falsy input to `[{}]` rather than to the normalized all-empty triple. Two effects worth deciding on deliberately rather than inheriting:

- `permissions_data` is now `[{}]` on a fully-revoked record where it used to be unset, and that value flows downstream to destinations that write ACL fields.
- "No one has access" ends up with two different digests: `[{}]` when it arrives as an empty fetch, and `[{"read": {"users": [], "groups": []}}, ...]` when it arrives as permission entries whose roles don't map to any operation. Drifting between those two representations would look like an ACL change and cause a spurious reprocess, even though effective access is identical in both.

Returning the normalized all-empty triple from the `not raw_permissions` branch would collapse the two. Either way, `test_permissions_captured_when_empty_list_passed` currently only asserts `is not None`, so the actual value isn't pinned by anything — worth asserting the shape.

**`_parse_batch_response` can raise on a null body**

`sub.get("body", {}).get("value", [])` — the `{}` default only applies when the key is absent, so a sub-response of `{"id": "0", "status": 200, "body": null}` raises `AttributeError` and takes down the whole index run rather than degrading. Separately, a 200 whose body is present but has no `value` key falls through to `[]`, which under the new contract is a fabricated revocation — the exact case the fetch-vs-revocation change set out to prevent. Something like `(sub.get("body") or {}).get("value")`, with a missing `value` treated as unavailable, covers both.

**Version bump**

This adds a public field to `FileDataSourceMetadata` and a new public `unstructured_ingest/utils/acl` module. A minor bump reads more accurately than a patch for a new consumable surface.

**Changelog**

Three entries, where two of them describe fixes made to the first during review. From outside the PR that reads as churn rather than as anything a consumer acts on — a single entry describing the digest and its None-vs-empty semantics covers what changed.

**The telemetry helper**

`_AclDigestTelemetry`'s own docstring notes the digest's marginal API cost is "zero by construction, not a separately measured value," which I think is exactly right — but if so, the counter isn't measuring anything a reader couldn't determine from the code. Meanwhile it emits a six-line prose log line on every index run and is a private name imported across module boundaries. I'd drop it and record the number wherever the cost analysis is being written up.

**Internal references in the code**

There's a fair amount of internal ticket and scenario shorthand sitting in comments, docstrings, and test names here, including inside a log message that ships to production. This repo is public, so none of it means anything to a reader who doesn't have the internal context open. Worth a pass to either drop them or restate the reasoning inline so the code stands on its own.

**Small one**

`_canonicalize`'s dict branch is redundant. `json.dumps(..., sort_keys=True)` already sorts keys both in the inner sort key and in the final encode, so the list-sorting branch is the only part doing real work.

danielle-unstructured-io:
Thanks Austin, this was a great pass. Rundown of what landed:

**`_parse_batch_response` null/malformed body** — fixed in `50c698a1`. `(sub.get("body") or {}).get("value")` now degrades a null/absent body or a missing `value` key to `None` (unavailable, skip the digest) while a present-but-empty `value` stays `[]` (genuine revocation). Added tests for the null-body and missing-`value` cases.

**`extract_permissions([])` returning `[{}]`** — fixed in `50c698a1`. The empty branch now falls through to the normalized all-empty read/update/delete triple, so a revoked record and a no-op-role record produce the same digest (no spurious reprocess from representation drift). Both the `extract_permissions` test and the drive-item test now pin the exact shape instead of just `is not None`.

**Version bump** — agreed, bumped to a minor (`1.8.0`) for the new public field + `utils/acl` module.

**Changelog** — collapsed to a single entry describing the digest and its None-vs-empty semantics.

**Telemetry helper** — dropped entirely in `1c1d383d`. You're right that the cost is zero by construction, so the counter wasn't measuring anything derivable from the code, and it was the one place internal shorthand shipped in a production log line. Cost is now documented once in the analysis (microbenchmark + write-up) rather than emitted at runtime.

**`_canonicalize` dict branch** — I looked at this closely and I'd push back: it's load-bearing, not redundant. `json.dumps(sort_keys=True)` canonicalizes dict *key* order but never reorders *list elements*, so the dict branch is what lets the recursion descend into the nested `users`/`groups` lists to sort them. Dropping it makes `compute_permissions_version(_perms(["a","b"])) != compute_permissions_version(_perms(["b","a"]))` — i.e. `test_reordered_users_same_digest` fails (confirmed by removing the branch and running it). Determinism regardless of principal ordering is the whole point of the canonicalization, so I've kept it. Happy to add a comment making that intent explicit if it read as redundant.

One remaining item from your public-repo note: the shipped log line is gone with the telemetry, but there are still `PLU-511`/scenario refs in source comments and test names. Most restate the reasoning inline already; I can do a sweep to drop the bare tags if you'd prefer.

danielle-unstructured-io:
Did the internal-reference sweep in `b107d44c` — dropped the ticket/scenario shorthand from comments, docstrings, and test docstrings so the code reads on its own in the public repo (restated the reasoning inline where a ref was carrying it, and kept the code pointer to `_extract_identity_ids_from_raw`). Left the CHANGELOG's ticket ref, since every entry in this repo carries one.

That should cover everything from your pass — thanks again.

awalker4 (approved):
LGTM!

unstructured_ingest/processes/connectors/onedrive.py
```diff
@@ -458,11 +521,13 @@ def _parse_batch_response(
             di = drive_items[idx]
             status = sub.get("status")
             if status == 200:
-                by_id[di.id] = sub.get("body", {}).get("value", [])
+                # A present-but-empty value list is a genuine "no permissions"
+                # (revoked) state and stays []. A null/absent body or a missing
+                # value key is malformed/unavailable and degrades to None (skip
+                # the digest) rather than a fabricated revocation.
+                by_id[di.id] = (sub.get("body") or {}).get("value")
```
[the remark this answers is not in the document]
danielle-unstructured-io: Fixed in 80381f8e. `_parse_batch_response` now validates the shape before trusting a 200: `body` must be a dict and `value` a list, otherwise the item degrades to `None` (skips the digest) instead of crashing the run or fabricating a revocation. A present-but-empty `[]` still means revoked.

test/unit/connectors/test_onedrive.py
```diff
@@ -447,7 +469,39 @@ def test_per_item_403_yields_empty_for_that_item(self):
...
+        with patch("requests.post", return_value=body):
+            result = indexer._fetch_permissions_raw(items, access_token="tok")
+        assert result == {"item-f.docx": []}
+
+    def test_per_item_null_body_degrades_to_none(self):
+        # A 200 with a null body must not crash the index run; it degrades to
+        # None (unavailable), not [] (which would be a fabricated revocation).
+        indexer = _make_indexer()
+        items = [_make_drive_item("f.docx")]
+        body = _batch_response(responses=[{"id": "0", "status": 200, "body": None}])
```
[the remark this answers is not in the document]
danielle-unstructured-io: Fixed in 80381f8e. Added a parametrized test covering `body=[]`, `body="invalid"`, and non-list `value` shapes — all degrade to `None`. Pairs with the parser guard in the same commit.

unstructured_ingest/processes/connectors/onedrive.py
danielle-unstructured-io: Update for the record: the telemetry described above was later removed entirely in 1c1d383d, per @awalker4's review — the digest's marginal API cost is zero by construction (it hashes permissions already fetched for ACL metadata), so a runtime counter measured nothing derivable from the code and shipped internal shorthand in a prod log line. Cost is now captured once in analysis (microbenchmark + PLU-511 write-up) instead. So N4/N5 are moot; resolving.

PR #781, opened 2026-08-08
test: skip embedder integration tests on transient provider errors

paulkarayan:
## What & why

The embedder integration tests call real hosted embedding providers and assert success. When a provider is momentarily rate-limited / throttled / unavailable (5xx) / timing out, the test should SKIP, not FAIL - the code under test is fine, the provider blinked. Auth / misconfig / quota errors must still FAIL. This is the unstructured-ingest counterpart of the SDE fix in platform-plugins (#1890), tracked under FIE-341.

Unlike the plugin `/invoke` tests, there's no HTTP status here - a transient failure surfaces as a raised exception, and the library wraps provider errors non-uniformly (OpenAI/Voyage/Together let timeouts bubble raw, Vertex wraps only auth, Mixedbread doesn't wrap). So `test/integration/embedders/utils.py` gets a `skip_on_transient_provider(provider_label)` guard that detects "transient" three ways, strongest first, and re-raises everything else:

1. the library's typed transient classes (`RateLimitError` 429, `ProviderError` 5xx, `TimeoutError` 408, `EmbeddingEncoderConnectionError`);
2. a transient HTTP status read off the raw SDK exception (`{408, 429, 500, 502, 503, 504}`);
3. a class-name fallback for well-known raw SDK transient exceptions (`ResourceExhausted`, `APITimeoutError`, ...).

`UserAuthError` (401), `UserError` (422), `QuotaError`, and `AssertionError` (wrong-output) are deliberately excluded from all three tiers, so a bad key / wrong model / real assertion failure still reds the build. Applied to both call sites per provider (`Embedder.run` and the raw `embed_query` / `embed_documents` paths) for the hosted providers: openai, azure_openai, vertexai, voyageai, octoai, togetherai, mixedbread, bedrock.

## Linked ticket

FIE-341

## Impact

Test infra plus two small, related library fixes in `unstructured_ingest/embed/vertexai.py`. The transient-skip guard exposed a gap in Vertex's `wrap_error`: a genuine quota exhaustion would have been misread as a transient blip and skipped instead of failing. Both fixes close that gap so quota still reds the build:

- `google.api_core.exceptions.ResourceExhausted` (HTTP 429, legacy `TextEmbeddingModel` transport) now maps to `QuotaError`, mirroring how OpenAI/OctoAI map `insufficient_quota`.
- `google.genai.errors.ClientError` (code 429, the Gemini-family google-genai transport) also maps to `QuotaError` — google-genai raises its own error type, not `api_core.ResourceExhausted`, so without this a Gemini 429 would surface raw and the 429/`ResourceExhausted` transient rules would swallow it.

`QuotaError` is excluded from all three transient tiers, so these are the only consumer-facing behavior change: a Vertex quota error now fails loudly rather than surfacing raw. HuggingFace is intentionally left unguarded — it runs a local sentence-transformers model (no hosted API, no rate-limit risk), so its `validate_raw_embedder` call passes no `provider_label` and falls to `nullcontext()`.

## Risk / rollback

Low; test-only; revert the PR. Fail-safe by construction: if none of the three tiers matches, the exception re-raises and the test fails exactly as today - never a false pass, and no tier can match an auth/quota/validation error.

## How it was verified

`uvx ruff check` clean on the touched files. `pytest test/integration/embedders/test_utils.py` — 50 passed: parametrized unit tests over `_is_transient_provider_error` and the skip/re-raise guard covering every tier (typed transient classes, 408/429/5xx statuses, class-name markers) plus the fail cases (`UserAuthError`/`UserError`/`QuotaError`/`AssertionError`/bare `ProviderError` all re-raise, no false pass), and a regression test driving the real Vertex encoder path to prove a bare no-embedding `ProviderError` FAILS rather than skips. The provider tests themselves need real creds, so they were not run end to end here.

## Worked Example

```
$ uvx ruff check test/integration/embedders/
All checks passed!

$ python -m pytest test/integration/embedders --collect-only -q
35 tests collected
```

Skip messages route through the repo's existing `safe_error_summary()` redaction helper so no raw exception text / secrets leak into CI output.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/781?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulkarayan:
Note on the transient-skip classifier: `ProviderError` is classified transient-vs-contract by inspecting `__cause__`/`__context__` (a bare `ProviderError` = contract defect → fails; a wrapped one → transient → skips). This is a **stopgap** — it's exact within the current embedder scope but couples skip-behavior to how the error was raised. Longer-term hardening (an explicit `transient` signal / a dedicated contract-error type) is tracked in **FIE-365**: https://linear.app/unstructured/issue/FIE-365/harden-embedder-transient-skip-replace-providererror-cause-heuristic

potter-potter (approved):
Solid implementation. The 3-tier classifier is well-designed and the unit test coverage in `test_utils.py` (50 parametrized cases) covers every tier and failure class exhaustively.

The two Cubic P2s are both addressed in this PR: `ResourceExhausted → QuotaError` fix in `vertexai.py` and the new unit tests.

One question worth tracking (not a blocker): the `ProviderError` bare-vs-wrapped heuristic (`__cause__ or __context__`) is a documented stopgap (FIE-365). Is there any current path in the library where a bare contract assertion `ProviderError` could be raised inside an `except` block? If not, FIE-365 covers it without any changes here.

paulkarayan:
@potter-potter thank you!

>  Is there any current path in the library where a bare contract assertion ProviderError could be raised inside an except block?
No such path exists in the current embedder scope. Every ProviderError in the embedders is produced by wrap_error() gated on a 5xx status. 

bedrock, openai, octoai, togetherai, voyageai all map http_response_code >= 500 -> ProviderError) and raised via raise self.wrap_error(e=e) inside the `except … as e`  so it's always a wrapped transient 5xx. 

tldr: no code updates needed.

PR #782, opened 2026-08-11
fix(pinecone): make vector_length optional on create_destination

micmarty-deepsense:
## What

`PineconeUploader.create_destination` declares `vector_length` as a required positional argument with no default. Any caller that cannot resolve the embedding dimension ahead of time fails with:

```
TypeError: create_destination() missing 1 required positional argument: 'vector_length'
```

## Why it matters

The embedding dimension is not always knowable before the upload step runs. It is unavailable whenever the embedding model is not one the platform can look up, for example a self-hosted OpenAI-compatible embedding endpoint, and also whenever a workflow has no embedder node at all. In those cases every record in the upload fails.

The failure is also broader than it needs to be. It happens at call binding, before the method body executes. `create_destination` already short-circuits on an existing index and returns without ever reading `vector_length`, so users whose index is already provisioned, who need no dimension whatsoever, still cannot get past the call.

## The change

Pinecone was the only connector implementing `create_destination` with a required `vector_length`. Weaviate and AstraDB already default it to `None`, and Databricks and Teradata do not accept the parameter at all. This brings Pinecone in line with that convention, so auto-create degrades to a no-op when the index exists and the dimension is unknown, which is the behaviour every other destination already has.

Adds a unit test asserting the method is callable without a dimension.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/782?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.8.1]
+
+### Fixes
+
+- **fix(pinecone): make `vector_length` optional on `PineconeUploader.create_destination`.** The parameter was a required positional with no default, so any caller that could not determine the embedding dimension upfront crashed with `TypeError: create_destination() missing 1 required positional argument: 'vector_length'` before the method body ran. This happened even when the target index already existed and no dimension was needed, since the failure occurred at call binding. Pinecone was the only connector implementing `create_destination` with a required `vector_length`: Weaviate and AstraDB already default it to `None`, and Databricks and Teradata do not accept it. Pinecone now matches, so auto-create is a no-op when the index exists and the dimension is unknown.
+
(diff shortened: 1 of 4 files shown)


awalker4 (approved):
LGTM. Maybe we can add a one-liner to ensure there is a useful message if the index doesn't exist, per Cubic's comment.

PR #788, opened 2026-08-14
fix(SAP-6): sftp host public key fetch

mpolomdeepsense:
# fix(SAP-6): make SFTP host-key pinning work, and stop leaking credentials in logs

Pinning a correct `host_public_key` on the SFTP connector failed against Azure Blob Storage SFTP, reporting a mismatch against an ECDSA key that was never configured. Fixing that surfaced three adjacent defects in the same path.

### 1. Constrain host-key negotiation to the pinned key's algorithm family

SSH separates a key's **format** (`ssh-rsa` — what a key blob or `known_hosts` entry names) from the **public key algorithm** negotiated in KEXINIT (`ssh-rsa`, `rsa-sha2-256`, `rsa-sha2-512`; RFC 8332). One RSA key serves all three.

`paramiko.SSHClient` promotes the literal known-hosts name (`ssh-rsa`) to the front of its preference list. Against a server that has dropped SHA-1 `ssh-rsa` — Azure among others — that entry matches nothing, so negotiation falls through to the next mutual algorithm, normally `ecdsa-sha2-nistp256`, which the pin does not cover.

`get_client` now passes `disabled_algorithms={"keys": ...}` so only the pinned key's family is offered. paramiko's own `Transport.connect()` performs this expansion correctly; `SSHClient` never routes through it.

Side effect: the `paramiko<5` cap added in 1.9.2 is no longer load-bearing for these servers.

### 2. Report host-key and authentication failures without echoing secrets

`BadHostKeyException.__str__` prints both host keys in full. Beyond log noise, showing the key the server just presented invites pasting it into `host_public_key` — which trusts whatever answered and defeats the pin.

When `host_public_key` is set, `BadHostKeyException` and `AuthenticationException` are re-raised as `UserAuthError` with a short message and `from None`. The `from None` also keeps paramiko's frames — which hold the password as a local in `client.py` `_auth` — out of the chained traceback.

Before:
```
Host key for server 'x.blob.core.windows.net' does not match: got 'AAAAB3NzaC1yc2EAAAAD...', expected 'AAAAB3NzaC1yc2EAAAAD...'
```
After:
```
SFTP host key verification failed for x.blob.core.windows.net: public keys do not match
```

### 3. `SftpDownloader.connector_type` was a `FieldInfo`, not a string

The dataclass assigned pydantic's `Field(default=CONNECTOR_TYPE, init=False)` where every sibling downloader uses a bare literal, so the `FieldInfo` object became the value. With no explicit `download_dir`, building the default download path raised:

```
TypeError: unsupported operand type(s) for /: 'PosixPath' and 'FieldInfo'
```

`SftpDownloader` was the only connector affected.

### 4. Strip credentials from URLs written to logs

`DataSanitizer.sanitize_url` dropped query parameters but returned `netloc` verbatim — and userinfo (`user:password@`) lives in `netloc`. An fsspec-style `remote_url` carrying credentials therefore reached connector error logs in the clear through `log_connection_failed`.

```
s3://myuser:SUPERSECRETPW@bucket/path   ->   s3://***@bucket/path
```

Userinfo is dropped whole rather than masking only the password, since schemes like `https://<token>@host` carry the secret in the username. This affects all connectors, not just SFTP.

## Testing

- Verified against a live Azure Blob Storage SFTP endpoint: with the account's RSA host key pinned, negotiation lands on `rsa-sha2-512` and a full index → download → partition run completes.
- New unit tests: pinned-family negotiation, host-key mismatch message (asserts neither key appears), authentication failure message, and `sanitize_url` credential stripping.
- Full unit suite green (1308 passed).

## Notes for reviewers

- Only `ssh-ed25519` and `ssh-rsa` are pinnable. Azure publishes RSA and ECDSA host keys, so RSA is the only option there. ECDSA support was implemented and deliberately dropped from this PR.
- The clean error messages apply only on the pinned path; with `host_public_key` unset, `get_client` uses fsspec's own `_connect` and raw paramiko exceptions still surface.
- `sanitize_url` is shared code touched by every connector — worth a second look if you'd prefer it split out.
- Out of scope: plugin pods render tracebacks through loguru with `diagnose=True`, which prints frame locals and therefore paramiko's `password`. That is a `platform-libs` / deployment concern and is not fixable in this repo.

Version bumped to 1.9.3 with a matching CHANGELOG section.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/788?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


potter-potter (approved):
LG

PR #789, opened 2026-08-14
fix: skip elements with no embedding in qdrant, chroma and astradb

awalker4:
## What & why

Elements with empty text are never embedded — `embed_documents` embeds only elements with truthy text, so the rest arrive at the upload stager with no `embeddings` key. Three destinations map that key onto a vector field, and each mishandled its absence differently:

- **Chroma** — the uploader passes one embedding per id, so a missing one becomes a null entry and Chroma rejects the entire batch with `ValueError: Expected a 1-dimensional array, got a 0-dimensional array nan in upsert`.
- **Astra DB** — the stager raises `No vectors provided. Please enable an Unstructured embedding provider or configure Astra to generate embeddings`, which points the user at embedding configuration they already have.
- **Qdrant** — the point is upserted with an empty vector. Qdrant *accepts* it rather than rejecting it, storing a point that no vector search can ever return.

All three now override the `should_include` hook on `UploadStager`, matching Milvus and Pinecone.

## Linked ticket

PLU-431

## Scope

Narrower than it could be, deliberately. Pinecone already filters these elements. LanceDB stores the resulting null vector without complaint and KDB.AI indexes on a declared vector column, but neither has a reported failure behind it, and both would add surface area to a layer we are migrating away from.

Two exclusions are behavioural rather than incidental: Astra DB keeps unembedded elements when `astra_generated_embeddings` is set, and Weaviate is untouched, because both read a missing vector as a request to vectorize server-side. Filtering there would silently discard data for anyone relying on a server-side vectorizer.

## Verification

- **Qdrant, end to end against a real server** (stager + uploader, unnamed single-vector collection): before, 2 points stored and vector search returned 1 of 2; after, 1 point stored and 1 of 1. The empty vector is accepted on 1.7.4, 1.12.0 and 1.19.0, so this is a silent-data issue rather than a rejected write.
- **Chroma** rejection reproduced in-process against `chromadb` with the exact list shape the uploader builds.
- **Astra DB** raise reproduced by running the stager over a text-less element.
- New unit tests cover the predicate and both the `.json` and `.ndjson` run paths for all three, plus the Astra-generates-vectors case.
- Full unit suite: **1305 passed**.

## Follow-ups (separate PRs)

- Write errors interpolate only `str(e)`, which is empty for transport-level exceptions such as `httpx` timeouts and `ResponseHandlingException`. A failure recorded as `Qdrant error:` with no detail is unactionable; preserving the exception type would fix that, and the same pattern appears across several connectors.
- The Qdrant uploader sets no client timeout, so httpx's 5-second default applies, and it issues every batch for a record concurrently with `wait=True`. That combination wants a timeout, a retry, and a bound on concurrency.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/789?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.9.3]
+
+### Fixes
+
+- **fix(qdrant, chroma, astradb): skip elements that have no embedding.** Elements with empty text are not embedded, and each of these destinations failed differently on them: Chroma rejected the entire batch, Astra DB raised an error asking the user to configure an embedding provider they already had, and Qdrant stored a point with no vector that search can never return. All three now drop those elements before upload, matching Milvus and Pinecone. Astra DB keeps them when it is configured to generate the vectors itself, and Weaviate is unchanged, since both treat a missing vector as a request to vectorize server-side.
+
 ## [1.9.2]
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Fixes
 
-- **fix(qdrant, chroma, astradb): skip elements that have no embedding.** Elements with empty text are not embedded, and each of these destinations failed differently on them: Chroma rejected the entire batch, Astra DB raised an error asking the user to configure an embedding provider they already had, and Qdrant stored a point with no vector that search can never return. All three now drop those elements before upload, matching Milvus and Pinecone. Astra DB keeps them when it is configured to generate the vectors itself, and Weaviate is unchanged, since both treat a missing vector as a request to vectorize server-side.
+- **fix(qdrant, chroma, astradb): skip elements the embedder left without an embedding.** Elements with empty text are not embedded, and each of these destinations failed differently on them: Chroma rejected the entire batch, Astra DB raised an error asking the user to configure an embedding provider they already had, and Qdrant stored a point with no vector that search can never return. All three now drop those elements during staging. An element that has text but no embedding is kept, since that means no embedder ran at all and each destination's existing error should still surface rather than a workflow silently uploading nothing. Astra DB drops nothing when configured to generate the vectors itself, and Weaviate is unchanged, since both treat a missing vector as a request to vectorize server-side.
 
 ## [1.9.2]
 
--- a/test/unit/processes/connectors/test_vector_destination_skips_unembedded.py
+++ b/test/unit/processes/connectors/test_vector_destination_skips_unembedded.py
@@ -1,0 +1,101 @@
+"""Vector destinations must not stage elements that have no embedding.
+
+Elements with empty text are skipped by the embedder and arrive at the stager without
+an "embeddings" key. Qdrant, Chroma and Astra DB each map that key onto a vector field
+and cannot use the element without it, so they drop it before upload.
+"""
+
+import json
+from pathlib import Path
+
+import pytest
+
+from unstructured_ingest.data_types.file_data import FileData, SourceIdentifiers
+from unstructured_ingest.processes.connectors.astradb import (
+    AstraDBUploadStager,
+    AstraDBUploadStagerConfig,
+)
+from unstructured_ingest.processes.connectors.chroma import (
+    ChromaUploadStager,
+    ChromaUploadStagerConfig,
+)
+from unstructured_ingest.processes.connectors.qdrant.cloud import (
+    CloudQdrantUploadStager,
+    CloudQdrantUploadStagerConfig,
+)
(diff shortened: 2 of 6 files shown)


dizon-u10d (commented):
Blocking: cubic-dev-ai's open AstraDB finding remains valid. `UploadStager.run` applies `should_include` before `conform_dict`; with the default `astra_generated_embeddings=False`, any nonempty element from a pipeline with no embedder is now silently omitted instead of reaching AstraDB's existing "No vectors provided" configuration error. Please preserve that error for nonempty text while continuing to drop the empty-text elements that an embedder intentionally leaves without embeddings.

I also assessed the other open bot threads: an `embeddings: []` value is not emitted by the normal embedder path and should continue to surface the existing invalid-vector error rather than be silently dropped; the two test-coverage suggestions are useful but non-blocking.

Validated with `uv run --locked --group test pytest -q test/unit/processes/connectors/test_vector_destination_skips_unembedded.py test/unit/interfaces/test_upload_stager.py` (22 passed), the related existing connector suites (23 passed, 13 skipped), Ruff on the changed Python files, and `git diff --check`. The repository's `check-version` command cannot run locally because macOS BSD sed is older than its required version; GitHub's matching check is green.

awalker4:
Good catch, that regression was real. Filtering on the absence of the `embeddings` key also swallowed the no-embedder case, where *every* element lacks the key — so a workflow with no embedder went from a clear "No vectors provided" error to a silent empty upload. Fixed in e69a5534 by narrowing the predicate to what the embedder actually skips: `bool(text) or "embeddings" in element`. An element with text but no embedding is now kept and still reaches the existing error.

Applied the same narrowing to Qdrant and Chroma, since they had the same shape of problem without a bespoke message — a no-embedder pipeline would have silently staged nothing where Chroma previously failed loudly. It also makes the predicate mean what its comment claims, and it's a smaller behavioural delta.

Agreed on leaving `embeddings: []` alone — the embedder never emits it, and an invalid vector value should keep surfacing the existing error rather than being silently dropped. Both test suggestions taken: the run-path tests now assert *which* element survived rather than just the count, and there's run-path coverage for `astra_generated_embeddings=True` plus cases pinning that the "No vectors provided" error still fires for a text-bearing element and for an empty vector.

Worth flagging that this leaves the predicate here deliberately different from Milvus and Pinecone, which still filter on key presence alone and therefore still have the silent-empty-upload behaviour for a no-embedder pipeline. I left them out of this PR rather than widen it, but it's a real inconsistency and probably worth its own change.

Full unit suite: 1311 passed.

dizon-u10d (approved):
My earlier blocking comment about AstraDB is addressed. In e69a5534, `AstraDBUploadStager.should_include` (astradb.py:346) now narrows to `bool(element_dict.get("text")) or "embeddings" in element_dict`, so a nonempty-text element from a no-embedder pipeline is kept and still hits `conform_dict`'s "No vectors provided" error (astradb.py:378) instead of being silently dropped. The same narrowing was applied to Chroma and Qdrant for consistency. Confirmed via `uv run --locked --group test pytest -q test/unit/processes/connectors/test_vector_destination_skips_unembedded.py test/unit/interfaces/test_upload_stager.py` (28 passed, up from 22 at the last commit, reflecting the new empty-embeddings and full-pipeline tests). Traced `UploadStager.run` to confirm `should_include` runs before `conform_dict` on every code path, so the raw `text`/`embeddings` keys are still present when the filter evaluates.

On cubic's two follow-up notes against e69a5534 (qdrant.py:86): the first was really about the code comment overstating Qdrant's behavior as a "loud failure" when it actually just leaves an unusable empty-vector point; the comment has since been rewritten to say exactly that ("leaving a point no search can return") rather than claiming a hard error, so the comment now matches the code and I don't read this as something the PR still needs to fix — turning it into a real Qdrant-side rejection would be a separate, more invasive change than this PR's scope. The second (add a full-pipeline test asserting `vector={}` for a no-embedder Qdrant element) is a reasonable coverage addition but non-blocking.

Also checked CHANGELOG.md and the version bump for 1.9.3 (consistent with the fix), and confirmed Weaviate/Milvus/Pinecone are correctly left untouched per the changelog's stated rationale (author flagged Milvus/Pinecone as a known, separate follow-up, not in scope here).

No remaining blockers from my side. Approving.

PR #790, opened 2026-08-18
fix(PLU-504): surface skipped Teams channels in a per-run summary

potter-potter:
## Summary

Follow-up to #787 (merged) addressing two non-blocking review observations, which are the same underlying issue: **a Teams crawl can finish "successfully" while silently omitting an entire channel's files**, with only a scattered inline log line as evidence.

Two scenarios trigger a benign per-channel skip today:
- A forbidden **shared**-channel `403` — assumed cross-tenant, so skipped rather than raised. But a same-tenant **Application Access Policy** that scopes the app away from that channel's site produces the identical `403`; the response alone can't distinguish "expected" from "misconfigured."
- The precheck (`_probe_files_read_scope_sync`) validates scope against `/groups/{team_id}/drive/root` — the group's default drive — but private/shared channels live on **separate sites**. An app scoped away from one channel's site passes precheck cleanly and only fails at runtime, silently.

Both are closed the same way the reviewer proposed: a per-run summary of skipped channels (count + names + reason), so a clean-looking run can't hide a channel that quietly disappeared.

## Changes

- `_get_channel_files_folder_sync` now returns `(folder, skip_reason)` instead of `Optional[dict]`. The reason travels with the skip (unprovisioned 404, forbidden shared `403` — flagged as possibly cross-tenant *or* app-scoped-away, unexpected 4xx, missing drive/item id).
- `_run_team_async` accumulates skipped channels and, at the end of the crawl, emits one consolidated `WARNING` naming every skipped channel with its reason and an `indexed N of M channel(s)` count — explicitly calling out the AAP misconfiguration possibility. Inline per-channel logs are unchanged.
- No behavior change to what gets indexed, error propagation, or the precheck. This is purely observability. The runtime error-propagation guarantees from #787 (standard/private 401/403, 429, 5xx propagate as typed errors) are untouched.
- Version bump `1.10.0` → `1.10.1` + CHANGELOG entry.

## Test plan

- [x] `test/unit/connectors/test_sharepoint.py` — 104 passed. Updated resolver tests for the tuple return; added coverage asserting the consolidated summary names the skipped channel (both an unprovisioned skip and a forbidden shared-channel skip, the latter asserting the "scoped away" ambiguity is surfaced).
- [x] `ruff check` clean on touched files.
- [ ] CI green.

## Notes

This does not attempt to *distinguish* a benign cross-tenant 403 from an AAP misconfiguration (impossible from the response alone) or to make precheck cover per-channel sites — it makes the runtime outcome **visible** so an incomplete crawl is auditable. Non-blocking follow-up to an already-merged, well-reviewed PR.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/790?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Fixes
 
-- **fix(sharepoint): surface skipped Teams channels in a per-run summary.** A Teams crawl skips a channel whose files folder isn't provisioned or returns a forbidden shared-channel 403, but those skips were only inline log lines a long run could bury — so a run that finished with zero errors could still be silently missing an entire channel's files (e.g. a cross-tenant shared channel, or an app scoped away from a channel's own site by an Application Access Policy, which is indistinguishable from the 403 alone). `_run_team_async` now emits one consolidated WARNING at the end naming every skipped channel with its reason and an `indexed N of M` count, so an incomplete crawl is visible rather than hidden.
+- **fix(sharepoint): surface skipped Teams channels in a per-run summary.** A Teams crawl skips a channel whose files folder isn't provisioned or returns a forbidden shared-channel 403, but those skips were only inline log lines a long run could bury — so a run that finished with zero errors could still be silently missing an entire channel's files (e.g. a cross-tenant shared channel, or an app scoped away from a channel's own site by an Application Access Policy, which is indistinguishable from the 403 alone). `_run_team_async` now emits one consolidated summary at the end naming every skipped channel with its reason and an `indexed N of M` count. Severity is calibrated so it stays meaningful: a suspicious skip (a forbidden channel or unexpected response — a possible silent miss) is a WARNING worth investigating, while a purely benign not-yet-provisioned channel is an INFO note rather than an alarm that fires on every run.
 
 ## [1.10.0]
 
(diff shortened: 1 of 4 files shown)


dizon-u10d (approved):
Reviewed the current head. The outstanding cubic inline comment does not appear to be a blocker: the consolidated summary is intentionally emitted after a completed crawl, while cancellation, downstream failure, or deliberate early consumption does not constitute the clean run this change is meant to summarize.

Validation: locked dependencies are current; the focused SharePoint suite passed (104 tests); lint and diff whitespace checks passed; and GitHub currently reports 28 passing checks with none failing. The local version-check command is limited by macOS's BSD sed requirement, but the package version and changelog entry are aligned and the repository's Linux CI version gate is passing.

awalker4 (approved):

PR #791, opened 2026-08-18
fix(PLU-504): redact Microsoft Graph response bodies in SharePoint errors/logs

potter-potter:
## Summary

Follow-up to the SharePoint Teams work (#787, #790) closing a review block: the Teams/SharePoint error paths interpolated the **raw Microsoft Graph response body** into raised error messages and logs after only a 500-char truncation. **Truncation is not redaction** — and this repo's own error policy already says so:

> `safe_error_summary`: *"Exception text (str(e), .args, .message, response bodies) routinely embeds credentials, connection strings, or request payloads, so it is never included. Only machine-readable fields are surfaced..."*

So the new code was inconsistent with the repo's established bar (and inconsistent within the same file — network errors used `safe_error_summary`, but HTTP error paths dumped the raw body).

## Changes

- Replaced `_truncate_body(resp.text)` (raw body, truncated, no redaction) with `_safe_graph_detail(...)` at every call site (`_probe_files_read_scope_sync`, `_list_channels_sync`, `_get_channel_files_folder_sync`, and the shared `_handle_client_request_exception` used by the downloader + indexer in both site and Teams modes).
- `_safe_graph_detail` parses the Graph error JSON and surfaces **only allowlisted, non-sensitive fields** — the enum-like `error.code` (e.g. `accessDenied`, `itemNotFound`) and the opaque `request-id`/`client-request-id` — and **never** the raw body or the free-text `message`. Non-JSON/unparseable bodies yield `"no additional detail"` (the HTTP status is still carried on the message).
- Removed `_truncate_body` / `_MAX_BODY_CHARS`.
- **Latent bug fixed:** this module imports a custom `ValueError` from `unstructured_ingest.error`, which shadows the builtin. A plain `except ValueError` would therefore *not* catch `json.JSONDecodeError` (a builtin `ValueError` subclass), so any non-JSON error body would have raised mid-error-handling. Now catches `builtins.ValueError`. (This was caught by the test suite — 29 error-path tests exercise non-JSON bodies.)
- Version bump `1.10.1` -> `1.10.2` + CHANGELOG.

## Why this is the right fix (not just "Graph bodies are usually benign")

Graph error bodies for these read endpoints don't actually echo the bearer token (it lives in the request header). But the repo's policy is body-agnostic: don't surface raw upstream bodies at all. Extracting structured, curated fields satisfies that bar and removes the argument entirely, while keeping diagnosis useful (status + Graph code + request-id).

## Test plan

- [x] `test/unit/connectors/test_sharepoint.py` — **111 passed**. Rewrote the three tests that asserted the old "truncate-but-include body" behavior to assert redaction (raw body / free-text message never appears in error or log; safe `code`/`request-id` do). Added `TestSafeGraphDetail` covering JSON parsing, `client-request-id` fallback, non-JSON bodies, empty/None, no-`error` object, and a JSON array (non-object) body.
- [x] `ruff check` clean on touched files.
- [ ] CI green.

## Downstream

Once released as `1.10.2`, the consuming image bump can pin `>=1.10.2` to satisfy the reviewer's "use a release that sanitizes these bodies" block.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/791?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dizon-u10d (approved):
Reviewed the current head. The prior cubic-dev-ai sanitizer concerns are addressed: the affected paths now emit only bounded, sanitized Graph code/request-id detail, while malformed and oversized bodies safely fall back without replacing the typed error.

Validated with `git diff --check`, `uv lock --check`, `uv run --locked --group test pytest -q test/unit/connectors/test_sharepoint.py` (116 passed), and `uv run --locked --group lint ruff check test/unit/connectors/test_sharepoint.py unstructured_ingest/processes/connectors/sharepoint.py`. I also checked the version/changelog against the repository's enforcing version-sync script; 1.10.2 is aligned. No blocking issues found.

PR #794, opened 2026-08-25
fix(databricks-volumes): make source precheck call Databricks (PLU-637)

paulkarayan:
## What & why

**Problem:** The Databricks Volumes source connection check never contacted Databricks, so a connector that could not work reported a successful connection test and then failed at job time. Both the indexer and the downloader `precheck` did nothing but call `get_client()`, and under personal-access-token auth the SDK's auth provider is a static `Authorization` header, so constructing the client makes zero network calls. An invalid token, an unreachable workspace host, a blocked egress path, a nonexistent catalog/schema/volume, and a missing Unity Catalog grant all passed the check. The first real request was `dbfs.list` inside the indexer's `run`, so the user found out via a failed job, with nothing pointing at the connection. It also made the check useless as a support signal: for this connector, "the connection test passed" carried no information.

**Change:** The indexer precheck now calls `current_user.me()` and then takes the first entry of a non-recursive listing of the configured volume path. `DatabricksVolumesDownloader.precheck` is removed rather than fixed: Platform Test Connection only ever exercises the indexer, so a downloader precheck has no path to run down, and this was the only downloader in the repo that defined one. Failures keep raising through the connector's existing `wrap_error`, so responses stay mapped to the typed user/auth/rate-limit/provider errors and stay redacted.

## Linked ticket

[PLU-637](https://linear.app/unstructured/issue/PLU-637)

## Why two calls on the indexer

They answer different questions, and separating them is what makes the failure readable.

`current_user.me()` proves the credentials and the host: a rejected token or an unreachable workspace fails here, before anything about the volume is in play. It is also what the uploader precheck in the same file already does, so source and destination now behave the same way.

The listing proves the rest of what users actually get wrong: that the catalog, schema and volume resolve, and that `USE CATALOG`, `USE SCHEMA` and `READ VOLUME` are granted. Those are exactly the failures a client constructor cannot see and a bare auth check does not cover.

Cost is bounded. `dbfs.list` is a generator in `databricks-sdk`, so nothing is requested until the first item is pulled; taking one entry issues a single request and stops, rather than paging a volume that may hold many files. It is pinned to `recursive=False` regardless of the indexer's own `recursive` setting, so a recursive source does not walk its whole tree during a connection test. An empty path is still valid and passes, matching how the fsspec source precheck treats an empty directory.

## Impact

**Users:** A Volumes source that was already broken now fails the connection test instead of passing it, which is the point of the change but is a visible difference: a connector that "tested green" yesterday can report an error today. The error is the real one and arrives at the moment the user can act on it, rather than as a failed job later. This matters most in locked-down deployments, where the plausible failures (egress allowlists, TLS trust roots, Unity Catalog grants) are precisely the ones only a live call can catch. Working connectors are unaffected: two extra requests during a connection test, none during a job.

**Wire contract / clients:** No API or config shape changes. Errors were already routed through `wrap_error`, so the error types a caller sees are the existing `UserAuthError` / `UserError` / `RateLimitError` / `ProviderError`, and `safe_error_summary` redaction is unchanged. The precheck lives on the shared `DatabricksVolumesIndexer` base class, so `databricks_volumes`, `databricks_volumes_aws`, `databricks_volumes_azure` and `databricks_volumes_gcp` all inherit it. None of them override `precheck`. The downloaders now inherit the `BaseProcess` no-op, like every other downloader in the repo. The Volumes uploader already made a live call and is untouched. The Delta Tables connectors (`databricks_volume_delta_tables` and the SQL `databricks_delta_tables`) have their own cursor-based prechecks and are not affected.

**Deployment:** The new calls go to the same workspace host the connector already uses, so any environment that could run the connector can run the precheck. On an air-gapped or proxied deployment the precheck will now fail where it previously passed, which is the intended behavior.

## Risk / rollback

Low. One connector module plus its tests, no schema or config change, revert-safe as a single commit. The behavior change is one-directional: connectors that were genuinely reachable still pass.

## How it was verified

Unit tests only. New tests drive the indexer precheck against a mocked `WorkspaceClient`: a rejected token (`Unauthenticated`) raises `UserAuthError` without leaking the token into the message, a nonexistent volume path (`NotFound`) raises `UserError`, a missing read grant (`PermissionDenied`) raises `UserAuthError`, the listing is asserted to be non-recursive and scoped to the resolved path, and an empty path passes. They fail on `main` and pass with the fix. The full unit suite and `make check-ruff` are clean.

Not verified against a live Databricks workspace: that would need a real workspace plus a deliberately invalid token and a volume with the read grant withheld, which is not available here. The generator behavior of `dbfs.list` was confirmed directly against the pinned `databricks-sdk` (0.88.0) rather than assumed.

## Proof

Reproduced against `main` by stashing the fix and running the new tests, which pass only because the precheck contacts nothing:

```
$ pytest test/unit/connectors/databricks/test_volumes.py -q --no-cov --tb=line
.....FFFF.FF.                                                            [100%]
E   Failed: DID NOT RAISE <class 'unstructured_ingest.error.UserAuthError'>
E   Failed: DID NOT RAISE <class 'unstructured_ingest.error.UserError'>
E   Failed: DID NOT RAISE <class 'unstructured_ingest.error.UserAuthError'>
E   AssertionError: Expected 'list' to be called once. Called 0 times.
E   Failed: DID NOT RAISE <class 'unstructured_ingest.error.UserAuthError'>
E   AssertionError: Expected 'me' to have been called once. Called 0 times.
6 failed, 7 passed in 4.25s
```

The mocked client recorded zero calls to `current_user.me` and zero to `dbfs.list`, which is the defect stated as an assertion.

With the fix applied:

```
$ pytest test/unit/connectors/databricks/ -q --no-cov --tb=short
69 passed in 2.27s

$ make test-unit
1478 passed, 1 warning in 79.48s

$ make check-ruff
All checks passed!
```

## Dependencies / merge order

none

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/794?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


potter-potter:
Unless you know something I don't, I'd completely remove the whole downloader `precheck`. A precheck on a downloader will never get run — this looks like wasted code that somehow got copied in here. (not by you. it has existed for some reason for a while)

Platform Test Connection remaps `source` → `index` only (`dag_node_from_request_node_spec` in etl-operator), so check-executioner never starts the downloader sidecar. Job preflight is off by default. And this is the only downloader in unstructured-ingest that even defines `precheck` — Hubert added a no-op `get_client()` copy onto it in #160 when the source landed, and nobody else followed that pattern.

The indexer change is the actual fix. I'd drop `DatabricksVolumesDownloader.precheck` and the downloader precheck tests.

awalker4:
Agree with Potter's read — Test Connection only ever exercises the indexer side, so a downloader-level precheck genuinely can't run in the path that matters here. Dropping it outright rather than leaving it half-alive is the right call.

Verified the indexer fix directly: pulled the branch, reverted just the `precheck` change, and all the new tests fail against that baseline and pass with the fix applied — matches the repro in the description. `current_user.me()` catching bad credentials before the listing call, and the listing itself proving the catalog/schema/volume resolve, covers the two failure modes that used to slip through silently.

One small thing worth a pass before merge: the description's "Why two calls on the indexer" section still frames the downloader as making its own `current_user.me()` call, which was true of an earlier revision but not the one that landed — worth a quick edit so it doesn't read as describing something the diff doesn't do.

awalker4 (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
...ient()`. Under personal-access-token auth the SDK's auth provider is a static `Authorization` header, so building the client makes no request and the connection check passed for an invalid token, an unreachable host, or a volume the user has no access to. The first real contact was `dbfs.list` inside the indexer's `run`, so the failure surfaced as a failed job after the user had been told the connector was fine. The indexer precheck now calls `current_user.me()` (credentials and host) and then takes the first entry of a non-recursive listing of the configured volume path (the catalog, schema and volume resolve, and the Unity Catalog `USE CATALOG` / `USE SCHEMA` / `READ VOLUME` grants are in place); an empty path stays valid. Failures keep routing through the connector's existing error wrapping, so responses are still classified into the typed user/provider/rate-limit errors and redacted. This is a visible behaviour change: a Volumes source that reported a successful connection test while being unusable now fails the test, and the four Volumes source connectors (`databricks_volumes`, `databricks_volumes_aws`, `databricks_volumes_azure`, `databricks_volumes_gcp`) all inherit the fix.
```
awalker4: I would just slim down the changelog to be human readable (i.e shorter 😅 )

paulkarayan:
@potter-potter @awalker4 both of your points are in, and the branch is rebased onto main and mergeable again.

**potter-potter, downloader precheck: dropped.** `DatabricksVolumesDownloader.precheck` is gone, along with its three unit tests and the `_downloader` helper that only existed to drive them. Your read checks out in this repo too: it was the only `Downloader` subclass that defined `precheck` at all, so removing the override falls back to the `BaseProcess` no-op every other downloader already has. The indexer precheck is untouched and is the actual fix.

**awalker4, changelog: slimmed.** Both entries are down to three sentences and one, from eight and four. The internal service name that was in there is out of the public changelog as well.

**awalker4, the description:** you are right that it no longer matches the diff, and it is worse now that the downloader precheck is gone. Two sentences need to go: "The downloader has no path of its own, so it calls `current_user.me()`" under **Change**, and the downloader half of **How it was verified**. I have left the body alone rather than rewriting a reviewed description under you; pk is editing it.

**Rebase:** onto `main` at 1.11.12, so this is now 1.11.13. The only conflicts were `CHANGELOG.md` and `__version__.py`, both the usual version-bump collision, no code conflict.

CI: green on `e2bfadb2`, 27 pass and 11 skipped. 68 databricks unit tests pass locally, ruff clean.

Nothing else outstanding from my side.

PR #799, opened 2026-08-28
fix(ibm-watsonx-s3): retry COS IncompleteBody and RequestTimeTooSkewed

potter-potter:
## Summary
- Retry IBM COS `IncompleteBody` and `RequestTimeTooSkewed` during Iceberg upload instead of failing the job as `ProviderError`.
- s3fs maps those errors to `OSError`/`PermissionError` and will not retry them; the Iceberg transaction has not committed, so a fresh delete+append is safe.
- Other `EINVAL` / `AccessDenied` failures stay non-retryable (including when those boto Codes sit on `__cause__`). Exhausted retries surface as `DestinationConnectionError`. Bumps to 1.11.6.

## Test plan
- [x] Unit: `test/unit/connectors/ibm_watsonx/test_ibm_watsonx_s3.py` (74 passed)
  - Classifier: distinctive AWS messages, boto `Code`, `__cause__`/`__context__` chain (including generic outer EINVAL/AccessDenied with the real Code on the inner ClientError)
  - Uploader: retry, exhausted-retry → `DestinationConnectionError`, non-retryable `EINVAL`/`AccessDenied` → `ProviderError`
  - `upload_data_table` re-raises retryable COS errors so `upload_dataframe`'s loop can start a fresh Iceberg transaction
- [x] CI unit tests + changelog/version-sync checks (green on first push; this commit is tests-only)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.11.6]
+
+### Fixes
+
+- **fix(ibm-watsonx-s3): retry COS IncompleteBody and RequestTimeTooSkewed during Iceberg upload.** s3fs maps those COS/S3 errors to `OSError`/`PermissionError` and then refuses to retry them. The uploader was wrapping them as `ProviderError`, turning a transient COS flake into a hard job failure. Both are now classified as retryable (without treating every `EINVAL` or `AccessDenied` as retryable), re-raised out of `upload_data_table` so `upload_dataframe`'s retry loop can start a fresh Iceberg transaction, and surfaced as `DestinationConnectionError` if retries are exhausted. The Iceberg transaction has not committed when these fire, so retrying delete+append is safe.
+
 ## [1.11.5]
 
 ### Fixes
--- a/test/unit/connectors/ibm_watsonx/test_ibm_watsonx_s3.py
+++ b/test/unit/connectors/ibm_watsonx/test_ibm_watsonx_s3.py
@@ -682,6 +682,33 @@ def test_is_retryable_cos_incomplete_body_walks_cause_chain():
     assert is_retryable_cos_incomplete_body(outer) is True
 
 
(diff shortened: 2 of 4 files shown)


dizon-u10d (approved):
Solid fix. Traced the full error path: `upload_data_table`'s inner `_upload_data_table` now lets `is_retryable_cos_io_error` exceptions bubble out as-is instead of swallowing them into `ProviderError`, the outer `upload_data_table` wrapper does the same, and `upload_dataframe`'s tenacity loop (now using `retry_if_exception` with `_retryable_upload_error`) picks them up for retry. On exhaustion they become `DestinationConnectionError`, which is the right signal for a connectivity flake, not a hard `ProviderError` failure.

The two classifier functions are appropriately tight:
- `is_retryable_cos_incomplete_body` restricts message matching to `OSError(EINVAL)` (exactly what s3fs `translate_boto_error` emits), so generic `EINVAL` and `InvalidArgument` are not caught.
- `is_retryable_cos_request_time_too_skewed` omits the type guard but uses a substring specific enough that false positives are negligible; `AccessDenied` / `InvalidArgument` are explicitly excluded and tested.
- Structured boto `Code` matching walks the full exception chain for both, handling the `set_cause=True` path where s3fs attaches the original `ClientError` as `__cause__`.

Cycle detection in `_walk_exception_chain` via `id()` is correct. The `_retryable_upload_error` closure captures `RESTError` from the enclosing scope correctly, preserving the original RESTError retry behavior while adding the COS flake cases.

Test coverage is thorough: exact-match, rejection (wrong errno, wrong message, wrong type), both `__cause__` and `__context__` chain walking, boto code matching, retry-success, retry-exhaustion, and non-retryable passthrough for `EINVAL`/`AccessDenied`. Fixture uses `max_retries_connection=2`, so the `call_count == 2` assertions are correct. Version bump to 1.11.6 and CHANGELOG entry are present and consistent.

PR #801, opened 2026-08-28
test(outlook): seed CI fixture mail and run the Outlook e2e without enforcing it yet

micmarty-deepsense:
## Why

`outlook.sh` has not run in CI since the e2e jobs moved to Python 3.12 on 2026-02-06. `test-src.sh` only runs scripts outside `full_python_matrix_tests` when the interpreter is 3.10, and `outlook.sh` was never added to that array. It is *also* listed in `tests_to_ignore`, so even when it did run its exit code was discarded.

The result is the worst kind of green: `src_e2e_test` reports success while the Outlook connector is not exercised at all. As far as I can tell from the run history, there is no confirmed-passing run of this test in the repository.

## What this enables, and what it deliberately does not

The Python-version gate is removed, so nightly and dispatched runs actually execute `outlook.sh`, and the fixture-update workflow can regenerate `test_e2e/expected-structured-output` for Outlook for the first time. The `tests_to_ignore` entry **stays**: the test runs for fixture generation and evidence, but its exit code does not yet gate CI. Connector e2e enforcement is consolidating in the orchestration repo, so enforcing here would build a gate in the place it is being moved away from. If and when the team wants enforcement here after all, it is the one `tests_to_ignore` line.

## Why the gate could not just be flipped

Turning it on needs fixture mail that actually exists in the target mailbox and survives being re-run. The connector names each downloaded file `sha256(message.id)[:16] + ".eml"`, and `message.id` is Graph-assigned and unstable across a folder move. A tear-down-and-recreate seeding strategy would therefore churn every fixture filename on every run.

So seeding is **create-if-absent**, keyed on a fixed extended-property marker rather than on `message.id`.

## What is here

A seeding script under `test_e2e/env_setup/outlook/`, following the shape of the existing `env_setup` provisioning for airtable, couchbase and sftp. It authenticates through the connector's own connection config rather than reimplementing token handling, and it fails distinctly and immediately on a 403, because that specifically means the app registration is missing `Mail.ReadWrite` rather than anything being wrong with the script.

The seeded set is deliberately small and generic. Five fixtures seed successfully and reproducibly against the live mailbox today: a thread starter, a message carrying inline quoted history, one with a small attachment, a cross-folder move target, and a target for mutable state (read, flag, category). Attachments are kept trivial on purpose: they are partitioned through the live hosted API, so elaborate ones would make fixtures drift with that API's releases rather than with connector behaviour.

The two Graph-native threading fixtures, reply and forward, do **not** seed, and are deferred behind an opt-in `SEED_DEFERRED_FIXTURES=1`. `createReply` comes back as a Graph 400 `ErrorInvalidReferenceItem`, and `createForward` fails client-side inside the SDK with no HTTP response at all. Both happen when the base message was only ever created through the raw message-create API and never actually sent. The root cause is unconfirmed and Microsoft documents no such constraint, so the deferral is a parking spot rather than a diagnosis. The practical consequence is that a default seeding run seeds the five that work and exits zero, which is what `outlook.sh` requires, since it invokes the seeder under `set -e`.

Two robustness fixes fell out of getting there:

- `main()` isolates failures per fixture, so one fixture's Graph request failing no longer aborts the whole run. A genuine provisioning failure, such as a missing permission or a missing seed folder, still exits immediately.
- `FollowupFlag`'s own constructor defaults its three date sub-fields to empty `DateTimeTimeZone` objects, which Graph rejects with `TimeZoneNotSupportedException`. The fix nulls them explicitly so only `flagStatus` is sent.

Also in the diff: `test_e2e/src/outlook.sh` and `test_e2e/test-src.sh` (the invocation path; the python-matrix gate removed, the `tests_to_ignore` entry deliberately kept per the section above), the version bump to 1.11.11, and its `CHANGELOG.md` entry.

## Branch history

Rebased onto `main`, with the history rebuilt into four commits: the seeder plus the CI run-enable, the per-fixture fixes and the threading deferral, the version bump, and the non-enforcing exit-code hold. Nothing under `.github/` is touched by this branch.

## Sequencing

Lands after #802 and #803. #803 changes every fixture's expected-output filename (a hash of the message id) and #802 changes `data_source.version`, so fixtures generated before those merge would be regenerated again immediately.

## Before this can go green

The first seeding run creates messages whose Graph-assigned ids, and therefore whose fixture filenames, cannot be known in advance. `check-diff-expected-output.py` will fail until someone runs the suite once with `OVERWRITE_FIXTURES=true` against the live mailbox and commits the resulting fixture JSON. That is a deliberate two-step.

It is also not something this PR's own checks will show either way. The live e2e jobs in this repo never run on pull requests: they run nightly, on manual dispatch, and post-merge on push to `main`. And with the exit code non-enforcing, a seeding failure in those runs shows up in the run log and as missing fixture regeneration rather than as a red job — the distinct 403 fast-fail still prints, it just does not gate. The fixture regeneration is a separate deliberate step someone has to take.

Two things worth a reviewer's attention:

- **The app registration now holds `Mail.ReadWrite`** (creation) on top of `Mail.Read` (existence checks), with admin consent, and real mailbox writes have since succeeded against it. This no longer stalls on provisioning.
- **`test_e2e/src/outlook.sh` is edited** to invoke the seeder after its existing credential gate. Without that there is no invocation path at all, but it is the one change here that goes beyond test data and configuration.

## An upstream bug worth knowing about

In `office365-rest-python-client` 2.6.2, `Message.create_reply()` and `create_forward()` both return `self` rather than the draft they build internally, unlike `MailFolder.copy()` in the same version which returns its own object correctly. Used naively this silently operates on the wrong message. There is a small documented bypass helper for it here.

## Checks

`ruff check`, `ruff format --diff`, `py_compile`, `shellcheck` and `shfmt` via the repo's own recipe are clean on every touched file, plus zero-network smoke tests exercising each build path. The seeder has also been run against the live mailbox, where the five fixtures above seed reproducibly.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/801?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.11.12]
+
+### Fixes
+
+- **test(outlook): seed CI fixture mail and run the Outlook E2E test without enforcing it yet.** `outlook.sh` was listed in `all_tests` but never added to `full_python_matrix_tests`, so `test-src.sh` skipped it on every run once CI standardized on Python 3.12, including the fixture-regeneration dispatch, which is why no Outlook fixtures could ever be produced. The python-version gate is removed so dispatch and nightly runs actually exercise the connector and can regenerate `test_e2e/expected-structured-output`; the test stays in `tests_to_ignore` for now, so its exit code does not gate CI while e2e enforcement consolidates in the orchestration repo. Because the test needs mail that actually exists in the target mailbox, and the connector derives each filename from `sha256(message.id)` (Graph-assigned, and unstable across a folder move), a tear-down-and-recreate strategy would churn every fixture name per run. `test_e2e/env_setup/outlook/` therefore seeds create-if-absent, keyed on a fixed extended-property marker, covering attachments, threading, a cross-folder move, and mutable read/flag/category state, and exits distinctly on a 403 so a missing `Mail.ReadWrite` grant is not mistaken for a code fault. The two Graph-native threading fixtures (reply and forward) currently fail against a never-sent base message and are deferred behind an opt-in (SEED_DEFERRED_FIXTURES=1), so a default seeding run seeds the remaining five and exits zero.
+
 ## [1.11.11]
 
 ### Fixes
--- a/test_e2e/env_setup/outlook/README.md
+++ b/test_e2e/env_setup/outlook/README.md
@@ -1,0 +1,120 @@
+# Outlook fixture seeding
+
+`seed_fixtures.py` guarantees that a fixed, known set of fixture messages exists in the
+Outlook mailbox the `outlook.sh` end-to-end test reads from, before that test runs. It is
+create-if-absent: every time it runs, it checks whether each fixture already exists and only
+creates the ones that do not.
+
+Fixture *content* lives in `fixtures.py`, one `SeedMessage` per fixture, each with a comment
+explaining the specific question that fixture exists to answer. This file explains how the two
+scripts fit into CI and why they have to work the way they do.
+
+This is a public repository. Every fixture is generic, synthetic test mail: no customer name,
+no real correspondence, no deployment-specific detail, anywhere in `fixtures.py`, in this file,
+or in commit history touching either. If you add a fixture, keep it that way.
+
+## How this gets invoked from CI
+
+`test_e2e/src/outlook.sh` calls `seed_fixtures.py` itself, right after its existing check that
+`MS_CLIENT_ID` / `MS_CLIENT_CRED` / `MS_TENANT_ID` / `MS_USER_EMAIL` are all set (and before it
+runs the actual ingest). `outlook.sh` is itself invoked by `test_e2e/test-src.sh`'s main loop,
+which is invoked by the `src_e2e_test` job in `.github/workflows/e2e.yml` (nightly / on-demand /
+post-merge only, never on pull requests) and by the `update-fixtures-and-pr` job in
+`.github/workflows/ingest-test-fixtures-update-pr.yml`. Both jobs already set the four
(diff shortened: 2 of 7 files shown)


micmarty-deepsense:
This PR should land after #802: #802 changes `data_source.version` (now a real, non-deterministic Graph change key instead of null) and, via a switch to immutable message ids, every fixture's expected output filename (a hash of the message id). Any fixtures seeded here should be generated against #802's code rather than the current version, and `data_source.version` should be treated as volatile in the comparator once this e2e is re-enabled. Both PRs currently bump the package to the same next patch version; whichever merges second will need to bump again to stay unique.

micmarty-deepsense:
Update: the earlier ordering note in this thread now needs a correction, the id-stabilization work split out of #802 into its own PR. This PR should land last, after both #802 and #803:

- #802 changes `data_source.version` (now a real, non-deterministic Graph change key instead of null).
- #803 switches to Graph's immutable message ids, which changes every fixture's expected output filename (a hash of the message id) and is its own coordinated, ship-gated rollout.

Any fixtures seeded here should be generated against both PRs' code, and `data_source.version` should be treated as volatile in the comparator once this e2e is re-enabled. All three PRs currently bump the package to the same next patch version; whichever merges last will need to bump again to stay unique.

PR #802, opened 2026-08-31
fix(outlook): emit real record versions and fully page folder/message listings

micmarty-deepsense:
## What was broken

The Outlook connector's per-message `data_source_version` was always blank: the change-key accessor queried the JSON key with the wrong casing, so the value it read never matched what the Graph API actually sends. Folder and message listings also only fetched a single page from the Graph API, so any mailbox or folder with more items than the default/requested page size was silently truncated rather than fully enumerated.

## What changes for users

- `data_source_version` now carries the real Graph change-key string for each message instead of always being null, so incremental change detection can correctly skip unchanged messages.
- Large folders and mailboxes with many folders are now fully enumerated instead of stopping at the first page.
- Message listing now requests only the fields the connector actually reads (`$select`), instead of Graph's default projection, which includes the full HTML body.

Message-identity stabilization (immutable message ids) is split out into its own follow-up PR, since it changes the record identifier itself and is a separate, coordinated rollout event; see #803.

## Note for reviewers

- The Outlook end-to-end test is currently skipped in CI, so its fixtures will be regenerated the next time it's re-enabled. See #801, which re-enables that test and seeds its fixtures. #801 needs to land after both this PR and #803: this PR changes `data_source.version` (now a real, non-deterministic change key instead of null), and the immutable-id PR changes every fixture's expected output filename (a hash of the message id). #801's fixtures must be generated against both PRs' code, and `data_source.version` will need to be excluded or treated as volatile in the fixture comparator once the e2e is re-enabled. On versioning: the merge order is #802 first, then #803, then #801 — stated in prose on purpose. Version numbers are re-picked against `main` at each merge (unrelated merges consume numbers faster than any reserved sequence survives; an earlier reservation scheme here went stale within days). This branch currently carries 1.11.9; renumber at merge time if `main` has moved again.
- This PR and #803 are sibling branches off `main`, not a stack, and both append tests to the same region of `test/unit/connectors/test_outlook.py`, so whichever of the two merges second resolves that one conflict at merge time. The connector module itself merges cleanly between them.
- `TestChangeKeyRawPropertyLookup` is a canary on the underlying SDK's property-casing bug, not regression coverage for this PR's fix - it's meant to fail (on purpose) if a future office365-rest-python-client upgrade fixes the casing mismatch, which is the cue to simplify back to the typed accessor. The regression coverage for the version fix itself is `TestMessageToFileDataVersion`.
- `get_all()` eagerly drains every page of a folder/mailbox into memory before `run()` yields any records, trading memory for correctness against the previous silent single-page truncation. Streaming enumeration is a reasonable follow-up if a mailbox's full listing turns out too large to hold at once, but isn't needed for this fix.
- The pagination tests now also pin the `$select` projection as concrete field names (so silent drift in `MESSAGE_SELECT_FIELDS` fails a test rather than being asserted tautologically) and cover a listing larger than one Graph page, asserting the enumeration comes back complete and in order.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/802?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.11.9]
+
+### Fixes
+
+- **fix(outlook): emit real record versions and fully page folder/message listings.** The Outlook connector read a message's change key through a typed accessor that queried the wrong JSON key casing, so `data_source_version` was always null and incremental change detection could never skip an unchanged message. Message listing also requested an out-of-range page size (Graph documents `$top` for `/messages` as 1..1000) and never followed `@odata.nextLink`, while folder listing had no page size set at all and was silently capped at Graph's default; either way, only the first page was ever returned. Listings now page through every result and, for messages, request only the fields the connector actually reads via `$select`, trading the memory of holding a full page against the previous silent truncation.
+
 ## [1.11.8]
 
 ### Enhancements
--- a/test/unit/connectors/test_outlook.py
+++ b/test/unit/connectors/test_outlook.py
@@ -13,7 +13,6 @@ from unstructured_ingest.processes.connectors.outlook import (
     OutlookConnectionConfig,
     OutlookIndexer,
     OutlookIndexerConfig,
-    _prefer_immutable_ids,
 )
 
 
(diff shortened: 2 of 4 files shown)


micmarty-deepsense:
@claude review

awalker4 (approved):
lgtm

PR #805, opened 2026-09-05
fix(none): classify a rejected credential during indexing

paulkarayan:
## What & why

**Problem:** A customer whose source connector credentials are wrong gets a run that reports success and moves nothing, with no error anywhere telling them why. On an SND, a source over a bucket holding three files returned `total_docs: 0` and `processing_status: SUCCESS` with bad credentials, while the same bucket with good credentials returned 3, and `/jobs/{id}/node-errors` carried nothing for the failed one. A connection check on that identical connector said "Authentication failed" the whole time, so we already knew and only the job path did not say so. Customers re-run these jobs assuming a blip; Biogen escalated one at Sev 2 after repeatedly retrying runs that had done nothing (ENG-1362).

**Change:** `FsspecIndexer.get_file_info()` now wraps its listing in the same try/except `precheck()` has always had, raising through `wrap_error`. That turns fsspec's bare `PermissionError` into the classified `UserAuthError` the rest of the stack can act on. The failure logs through a new `log_listing_failed` rather than reusing `log_connection_failed`: both fire for the same connector against the same endpoint, one during precheck and one during the job, so sharing a template left a log search or an alert unable to tell the two phases apart. Both wrappers delegate to one helper, so sanitization and `exc_info` stay identical and only the sentence differs.

**Blast radius:** 3/5 -- a base class every fsspec connector inherits, converting a silent path into a raised error.

## Linked ticket

none. This is the connector-side half of two tracked, unassigned defects:

- https://linear.app/unstructured/issue/ENG-1362/ingestreprocess-job-reports-completed-instead-of-failing-when-s3 (Backlog) -- Biogen, Sev 2, denied S3 ListBucket reported as COMPLETED with no error surfaced.
- https://linear.app/unstructured/issue/DTPL-773/etl-jobs-that-fail-at-the-source-connector-0-docs-terminate-as (Triage) -- the job-status model has no job-level FAILED state, so a connector-level failure is structurally unrepresentable.

This PR does not close DTPL-773; that is a job-status-model change. It removes the reason a connector failure arrives unclassified in the first place.

Client-facing follow-up: a Teradata workflow run reported Finished after processing no files -- https://app.usepylon.com/issues?issueNumber=2335

## Impact

- **Customers:** a run whose source credentials are rejected now fails with an authentication error attached, instead of reporting success having indexed nothing. Anyone hitting a bad key, a rotated secret, or a bucket policy that denies listing stops silently getting empty runs. A source that is genuinely empty, or whose files are all filtered out, is unaffected: an empty listing is still an empty listing, and only a raised exception changes behaviour.
- **Internal (devs / ops / other teams):** this is a shared library. `FsspecIndexer` is the base for six connectors -- S3, GCS, Azure, Box, Dropbox and SFTP -- and all six inherit the change. OneDrive does not: `OnedriveIndexer` subclasses `Indexer` directly (`unstructured_ingest/processes/connectors/onedrive.py:244`) and has its own listing path, so it keeps whatever classification it had. That breadth is the point (the same swallow existed for all of them) and it is also the risk: any of those whose backend raises during a listing now surfaces that as a classified error rather than an empty result. Support gains an actual error to point at instead of "the job says it finished".
- **Wire contract / clients:** none. No signature, return type, or response shape changes. `get_file_info()` returns exactly what it returned before on every path that previously returned; only the failure path differs, and it previously escaped as an unclassified builtin.
- **Deployment target considerations:** the library ships inside plugin images, so SaaS, DI, in-VPC, on-prem and SND all pick it up when those images rebuild. Nothing target-specific: no new dependency, config key, or network call. I proved it against an SND's minio; I have not exercised it against real AWS, GCS or Azure endpoints.

## Risk / rollback

Moderate for a two-file diff, because it is a base class every fsspec connector inherits. It converts a previously-silent path into a raised error, so a connector that has been quietly returning empty on a recoverable listing hiccup would now fail the run instead. That is the intended correction, but it is a behaviour change and it is why the blast score is 3. Straight revert, no migration and no persisted state.

The filtering below the listing is deliberately left OUTSIDE the try. A malformed entry is our own bug, and relabelling it a connection failure would send the reader off to re-test perfectly good credentials. A test pins that.

## How it was verified

Ran the real `S3Indexer` against an SND's minio before and after, good and bad credentials, and read all four outcomes. Because this is a base class the other fsspec connectors inherit, I ran every one of their suites explicitly, not just the file I touched: `test/unit/connectors/fsspec/` (s3, gcs, azure, box, dropbox, sftp, plus the base) is 97 passed, and the full unit suite is 1471 passed / 35 skipped. Those are unit suites in this repo; I did not run the downstream consumers that vendor this library as a wheel. NOT verified: the full chain end to end on a live job. Every link from `UserAuthError(401)` through the plugin library's in-band error line, the ETL invoker's `>= 400` check, and into the console's node-errors display is read from source, but I did not rebuild the plugin image and re-run the SND job to watch it, so that last mile is reasoned rather than observed.

## Proof

**Repro (snd+local).** On pk-snd, put three files in a minio bucket, then built two S3 sources over that same non-empty bucket differing only in credentials, plus a workflow and a run for each. The non-empty bucket is the point: an earlier attempt used an EMPTY bucket, which cannot separate "auth failed" from "auth fine, nothing there". That run proved nothing and was discarded.

**Observed.** Identical bucket, three files, only credentials differ:

| source | total_docs | processing_status |
| --- | --- | --- |
| good credentials | 3 | FAILED (documents failed downstream) |
| bad credentials | 0 | SUCCESS |

`GET /jobs/{id}/node-errors` returned `{"nodes": []}` for the bad-credential job, so nothing recorded that the source could not authenticate.

Isolated to the line by running the real `S3Indexer` against that minio:

```
good creds: precheck -> returns,              get_file_info -> returns 3
bad creds:  precheck -> raises UserAuthError, get_file_info -> raises raw PermissionError
```

`precheck()` wraps through `wrap_error` (fsspec.py:109-143); `get_file_info()`, the path a job takes, had no try/except. An unclassified error carries no `status_code`, so downstream the run is indistinguishable from one that legitimately found nothing.

**Failing test.** `test/unit/connectors/fsspec/test_fsspec.py::TestFsspecIndexerListingErrorsAreClassified`

```
uv run --locked --no-sync pytest -q test/unit/connectors/fsspec/test_fsspec.py

fsspec.py reverted to origin/main:  6 failed, 21 passed in 1.44s
this head:                          27 passed in 1.15s

  FAILED ...::test_rejected_credential_is_wrapped[ls]
  FAILED ...::test_rejected_credential_is_wrapped[find]
  FAILED ...::test_the_failure_is_logged_as_a_listing_not_a_precheck[ls]
  FAILED ...::test_the_failure_is_logged_as_a_listing_not_a_precheck[find]
  FAILED ...::test_missing_prefix_is_wrapped[ls]
  FAILED ...::test_missing_prefix_is_wrapped[find]
E   PermissionError: The Access Key Id you provided...
E   FileNotFoundError: no such prefix
```

Same revert across the whole `test/unit/connectors/fsspec/` directory is `6 failed, 91 passed`, against `97 passed` at this head.

**After.** Same probe, same bucket, same credentials, before -> after:

| case | before | after |
| --- | --- | --- |
| good creds, `precheck` | returns | returns (unchanged) |
| good creds, `get_file_info` | returns 3 | returns 3 (unchanged) |
| bad creds, `precheck` | `UserAuthError` | `UserAuthError` (unchanged) |
| bad creds, `get_file_info` | `PermissionError` | **`UserAuthError`** |

Full unit suite 1471 passed, 35 skipped. ruff `check` and `format --check` clean, no blanket suppressions added.

**Risk noted in the proof.** The last mile is reasoned, not observed (see verification above). `hydrate_file_data` has the SAME unguarded `client.ls` on the single-file path and was left alone as out of scope. This is a shared library and I ran its own suite, not every consumer's.

## Dependencies / merge order

none. Independent of the console-side change that renders a zero-file run (https://github.com/Unstructured-IO/platform-workflow-builder/pull/1913 and its follow-up), which improves what the user reads whether or not this lands. This one makes the underlying cause reportable.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/805?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulkarayan:
## Repro-first proof

**Bugfix** — A source connector whose credentials are rejected now fails the run instead of reporting a success that indexed nothing.

### Reproduced the broken state
- **Environment:** `snd+local`
- **How:** On pk-snd, put THREE files in a minio bucket (s3://fie-authprobe), then built two S3 source
connectors over that same non-empty bucket differing only in credentials, plus a workflow and a
run for each.
  good creds source 5e869ad3-fbe4-486b-8866-f3f0ff2dcc72, job bab5369c-1ced-4896-bc45-b9cab81fff16
  bad creds  source 281686fb-4e70-4e72-946c-56b6efa388e5, job f3ec9e07-96e3-4167-823f-76a73245956f
The non-empty bucket is the point. An earlier attempt pointed the bad-credential source at an EMPTY
bucket, which cannot distinguish "auth failed" from "auth fine, nothing there"; that run proved
nothing and was discarded.
- **Observed:** Identical bucket, three files in it, only the credentials differ:
  good creds -> total_docs 3, processing_status FAILED (documents failed downstream)
  bad creds  -> total_docs 0, processing_status SUCCESS, status COMPLETED
GET /jobs/{id}/node-errors returned HTTP 200 {"nodes": []} for the bad-credential job, so nothing
anywhere recorded that the source could not authenticate. A connection check on that same connector
correctly returned FAILURE "Connection test failed.\nAuthentication failed." So the platform knows
the credentials are bad on one path and reports success on the other.

Isolated to the line by running the real S3Indexer against the same minio:
  good creds: precheck -> returns,               get_file_info -> returns 3
  bad creds:  precheck -> raises UserAuthError,  get_file_info -> raises raw PermissionError
precheck() wraps its listing through wrap_error (fsspec.py:109-143); get_file_info(), which is the
path a job takes, had no try/except at all. An unclassified PermissionError carries no status_code,
so downstream the run is not distinguishable from one that legitimately found nothing.
- **Evidence:** `job f3ec9e07-96e3-4167-823f-76a73245956f on pk-snd (bad creds, 3-file bucket, SUCCESS/0 docs) against bab5369c-1ced-4896-bc45-b9cab81fff16 (good creds, same bucket, 3 docs)`

### Failing test (red)
- `test/unit/connectors/fsspec/test_fsspec.py::TestFsspecIndexerListingErrorsAreClassified` (unit) — committed
- transcribed by the author (not captured by a runner)

```
Reverted fsspec.py to origin/main and re-ran; 4 red, 21 passed:
  FAILED ...TestFsspecIndexerListingErrorsAreClassified::test_rejected_credential_is_wrapped[ls]
  FAILED ...TestFsspecIndexerListingErrorsAreClassified::test_rejected_credential_is_wrapped[find]
  FAILED ...TestFsspecIndexerListingErrorsAreClassified::test_missing_prefix_is_wrapped[ls]
  FAILED ...TestFsspecIndexerListingErrorsAreClassified::test_missing_prefix_is_wrapped[find]
E   PermissionError: The Access Key Id you provided...
E   FileNotFoundError: no such prefix
```

### Fix
- Wrap the listing calls in FsspecIndexer.get_file_info() in the same try/except precheck() already
uses, raising through self.wrap_error(e=e). S3ConnectionConfig.wrap_error (s3.py:138-142) already
maps PermissionError to UserAuthError, which carries status_code 401 (errors_v2.py:60-62), so the
error now reaches the plugin layer classified. The filtering below the listing is deliberately left
OUTSIDE the try: a malformed entry is our own bug and must not be relabelled a connection failure.
- **Files:** `unstructured_ingest/processes/connectors/fsspec/fsspec.py`

### Proof it's resolved
- **Test green:** yes
- **Environment:** `snd+local`
- **Evidence:** `Same probe, same minio bucket, same credentials, before -> after:
  good creds  precheck       returns            -> returns            (unchanged)
  good creds  get_file_info  returns 3          -> returns 3          (unchanged)
  bad creds   precheck       UserAuthError      -> UserAuthError      (unchanged)
  bad creds   get_file_info  PermissionError    -> UserAuthError      (THE CHANGE)
Suite: 4 red before the fix, 25 passed after on the fsspec file; full unit suite 1444 passed,
60 skipped. Pinned-ruff check and format --check both clean; no blanket suppressions added.`

---
_Auto-generated from this branch's `.proof.toml` (repro-first proof gate). Advisory._

tabossert:
Approving. I reverted `fsspec.py` to `origin/main` myself and got exactly the four failures the body claims, the filter staying outside the try is a real design boundary and a test pins it, and nothing in the diff can regress a successful listing. Four notes, none blocking.

`CHANGELOG.md:5` ends with "Inherited by every fsspec-backed connector (s3, gcs, azure, box, dropbox, sftp, onedrive)" and the PR body repeats it. OneDrive's indexer is not an `FsspecIndexer`: `onedrive.py:244` is `class OnedriveIndexer(Indexer)`. The complete set of `FsspecIndexer` subclasses is six: s3, gcs, azure, box, dropbox, sftp. (`onedrive.py:844` does carry `location_shape=LocationShape.FSSPEC_URL`, but that is a URL-shape marker on the destination registry entry, not a base class.) One word to delete in both places.

`hydrate_file_data` (`fsspec/fsspec.py:232`) still has the identical unguarded `client.ls`. You flag it as out of scope in the body and that's fair. One thing worth knowing before you decide: `git grep hydrate_file_data -- unstructured_ingest test` returns only the definition, so it has zero in-repo callers and nothing here would catch a regression in it. Same five lines if you want it.

The new log line is phase-ambiguous, and no test pins it. You reuse `log_connection_failed` with the same `connector_type` and the same endpoint f-string `precheck()` passes at `fsspec.py:138`, so its endpoint template ("Failed to validate %s connection to %s: %s") now fires from two phases with every argument identical except the exception. Both call sites pass `exc_info=True`, so the attached traceback does name `get_file_info` vs `precheck`; it's anything keyed on the message string alone that can't tell them apart. A distinct message or a phase argument fixes it. Separately, `log_connection_failed` is replaced with a bare `MagicMock` at `test_fsspec.py:353` and never asserted anywhere in the file: I mutated the endpoint f-string to garbage and `TestFsspecIndexerListingErrorsAreClassified` still went 6/6 green.

Question, which caller reaches `get_file_info()` without having run `precheck()`? `Pipeline._run_initialization` prechecks every step, and your own probe shows precheck already raising `UserAuthError`. But it then flattens it to `f"[{type(e).__name__}] {e}"` and raises `PipelineError` (`pipeline.py:149-153`), dropping `status_code`. So the SND result implies the platform driver drives the indexer directly and skips precheck. If that's the ETL plugin shell, say so in the body. And if `Pipeline` users are also affected, `pipeline.py:149` is a second defect this PR doesn't reach.

Last thing: 806 also writes `1.11.7` and its own `## [1.11.7]` heading, and `check-version` is its own CI job (`.github/workflows/unit_tests.yml:57-62`), so whichever of the two lands second goes red and re-bumps.

tabossert (approved):

paulkarayan:
Thanks Trevor, all four land. I reproduced the mutation before touching anything and it went exactly as you said.

**OneDrive: you're right, and it was wrong in two places.** `git grep -n "FsspecIndexer)" unstructured_ingest/` returns six and only six -- s3, gcs, azure, box, dropbox, sftp. `OnedriveIndexer` is `class OnedriveIndexer(Indexer)` at onedrive.py:244 with its own listing path, so it inherits nothing here. Dropped from the CHANGELOG on the branch; the PR body is going the same way.

**The log line: fixed, and the test now kills the mutant.** I ran your mutation first -- replaced the endpoint f-string with the constant `"GARBAGE-MUTANT"` and `TestFsspecIndexerListingErrorsAreClassified` came back 6 passed, 19 deselected. Nothing read that string, because `log_connection_failed` was a bare MagicMock at test_fsspec.py:353 with no assertion behind it.

Two changes. `log_listing_failed` is new alongside `log_connection_failed`, both delegating to one `_log_endpoint_failure` helper so sanitization and `exc_info=True` stay identical and only the template differs; `log_connection_failed`'s text is byte-for-byte unchanged. Indexing now reads "Failed to list s3 contents at s3://bucket/prefix", precheck still reads "Failed to validate s3 connection to s3://bucket/prefix". And the MagicMock is gone, replaced by caplog assertions on the message, on the endpoint, and on the absence of the precheck sentence. Three mutations that used to survive now fail both parametrizations:

```
garbage endpoint constant          2 failed, 6 passed
revert to log_connection_failed    2 failed, 6 passed
delete the log call entirely       2 failed, 6 passed
restored                          27 passed
```

**Your question: it is the ETL plugin shell, and it never calls precheck on that path.** Traced it rather than guessing.

`IndexerJob.run()` in platform-plugins (`platform_plugins/plugins/index/index/app.py:70-101`) builds the indexer from `SOURCE_CONNECTOR_CONFIG_REGISTRY` and goes straight to `indexer.run_async()` / `indexer.run()`. `IndexerJob.precheck` is a sibling method, and the plugin library wraps the two as separate HTTP endpoints -- `POST /invoke` calls `job.run`, `GET /precheck` calls `job.precheck` -- with nothing inside `/invoke` reaching for the other. The orchestration side confirms it: `IndexInvoker.invoke_plugin()` POSTs to `/invoke` and has zero references to `/precheck`. `/precheck` is only reached by `run_preflight_gate()` in plugins_controller, which is behind `enable_preflight: bool = False` and, when on, runs once at pod startup before the invoker is even constructed, not per job. So `Pipeline` is not what the platform runs, and a job's listing has never been preceded by a precheck.

**pipeline.py:149-153 is a second defect and I'm leaving it.** `failures[...] = f"[{type(e).__name__}] {e}"` then `raise PipelineError("Initialization failed")` does drop `status_code`, so the CLI/Pipeline path still loses the classification this PR restores. It is a real bug, it is not on the path the customer hit, and fixing it means deciding what a multi-step failure raises when two steps fail with different classifications. That is its own change, not a rider on this one. Happy to file it or take it next, whichever you prefer.

**`hydrate_file_data`: leaving it, and I'll say why rather than just calling it out of scope.** Zero in-repo callers cuts both ways. There is nothing to regress, but there is also nothing to prove a fix works against, and the function is not a five-line copy of `get_file_info` -- it raises `ValueError` on zero files and on more than one, so "the listing came back empty" is already an error there rather than a silent zero. Wrapping it would be an untested change to dead code in the same commit as a tested change to live code. If it acquires a caller, it should get the same treatment plus a test that exercises it.

Adjacent issue found: instantiating the base `FsspecIndexer` directly leaves `connector_type` as the pydantic `FieldInfo` rather than a string, because the base declares it `Field(default=CONNECTOR_TYPE, init=False)` on a plain `@dataclass` while every real subclass assigns the string. Only bites direct instantiation, which is tests. Not changed.

**On the 1.11.7 collision.** Both PRs prepend to CHANGELOG.md line 1, so the second one hits a textual conflict whether or not the number matches -- renumbering now wouldn't save the rebase. `git merge-tree` across the two current heads conflicts on CHANGELOG.md and nothing else; `fsspec.py` auto-merges. My recommendation is this one first: it is the defect fix with the SND evidence behind it and the shorter body, and https://github.com/Unstructured-IO/unstructured-ingest/pull/806 reads better as diagnostics layered on a source path that already classifies its failures. Whichever goes second rebases onto main, moves its entry under a fresh `## [1.11.8]` heading, and bumps `__version__` to match. I have not renumbered either branch, so they are both still 1.11.7 as you found them.

Full unit suite is 1471 passed / 35 skipped, ruff check and format clean.

PR #806, opened 2026-09-06
feat(none): report which zero a run hit when it indexed nothing

paulkarayan:
## What & why

**Problem:** A customer's workflow finished successfully having moved zero files, and every way of arriving at zero looked like the same zero. Teradata asked us whether their source was empty, unreachable, or filtered out, and nothing in the run could tell those apart: `FsspecIndexer.get_file_info()` listed and filtered in one expression, so the pre-filter count was gone before anyone could read it; the glob/size `Filterer` logged a DEBUG line per dropped file and counted nothing; `JobNodeStats` only counts records that became a `FileData`, so a record that never got that far is invisible to it. An empty prefix, a prefix holding only directory markers and zero-byte keys, and `file_glob` rejecting every file all reach the platform as the identical fact -- 0 documents, run succeeded.

**Change:** Keep the pre-filter listing count and report it next to the surviving count as `source.listing.returned` / `source.listing.retained`, and report `filter.<stage>.received` / `filter.<stage>.retained` per filter stage (`indexed`, `downloaded`, `uncompressed`, so a file that survives all three passes is not counted three times). Both ride the OTel span the work is already inside -- the channel the platform already points somewhere via `ProcessorConfig.otel_endpoint` / `OTEL_EXPORTER_OTLP_ENDPOINT` -- and each zero case also emits one consolidated `WARNING` naming which zero it was, following the end-of-run summary precedent from https://github.com/Unstructured-IO/unstructured-ingest/pull/790. Both families land on the root `ingest process` span, not on `indexer` or `filter step`, so query them there: `IndexStep.run` is a generator function under `@instrument`, which opens and closes its span around the call that merely builds the generator, and `FilterStep` reports after the instrumented `__call__` has already returned. That is pre-existing decorator behavior, not something this change introduces, but whoever writes the dashboard query needs it. Reporting only: nothing new is raised, and the telemetry call swallows its own errors, which matters in a base class every fsspec connector inherits. Two internal signatures move with it, both source-compatible: `FilterStep.__call__` and `Pipeline.apply_filter` each gain `stage: str = "indexed"`. All three `apply_filter` call sites pass an explicit stage; nothing else in the repo calls it.

**Blast radius:** 2/5 -- additive counters and warnings on a shared fsspec base class and the filter step, plus one early return that replaces a crash. No behavior change to what gets indexed.

## Linked ticket

Client-facing follow-up: Teradata could not get an answer to "why did this run move nothing?" -- https://app.usepylon.com/issues?issueNumber=2287

## Impact

Customers see a named cause instead of a bare zero: a run that indexes nothing now says whether the listing came back empty, came back full of directory markers and zero-byte keys, or was emptied by the configured globs, both as a `WARNING` in the job logs and as span attributes wherever the platform points OTLP. Nothing changes for a run that moves files. Scope worth knowing before you rely on it: the listing counts exist for the fsspec family only (s3, gcs, azure, box, dropbox, sftp inherit `FsspecIndexer.get_file_info` and none of them override it); the filter counts are pipeline-wide and apply to every connector; the missing-credentials cause is handled separately in https://github.com/Unstructured-IO/unstructured-ingest/pull/805; and neither number reaches `JobNodeStats`, which still has no slot for a record that failed to become a `FileData`.

## What the second commit changes

The first commit gave `Pipeline.apply_filter` a required `stage` while `FilterStep.__call__` got a default, so a subclass or an out-of-repo caller passing only `records` would have hit a `TypeError`. The second gives it the same default.

It also rewords the all-dropped warning. It used to assert that the configured filters dropped everything. A record can also fail to be retained by erroring: `file_data_from_file` raises on a malformed cached FileData, `run_async` catches it and returns `None`, and that `None` counts as not-retained. The warning would then name the glob config as the cause of something it had no part in. It now says what was observed and keeps the config as context, with the reason recorded at the site.

## The third commit: an uncompress that extracts nothing

`clean_results` returns `None`, not `[]`, when everything it is handed is falsy. The post-download path guards for that; the post-uncompress path handed the `None` straight to `apply_filter`, whose list comprehension iterated it and died with `TypeError: 'NoneType' object is not iterable`. The full trigger is an uncompress step, a configured filterer, and every archive extracting to zero regular files -- which is the exact everything-got-dropped case this PR exists to name, so a bare TypeError there is the worst available outcome.

The guard is at the call site, matching the post-download one, rather than `records or []` inside `apply_filter`. An empty uncompress is not a zero that filtering produced, and swallowing it in `apply_filter` would put `filter.uncompressed.received=0` on the span and log a filter warning for a drop the filters had no part in.

Pre-existing on main. Taken here because the call site is one of the lines this PR edits.

## Risk / rollback

Low. Additive reporting behind a try/except that never lets telemetry fail a job -- revert the PR to back it out. One caveat to read the numbers by: `source.listing.retained` is counted before `sample_n_files`, so on a sampled run it is larger than the number of files actually indexed.

## How it was verified

Drove the three zero cases through `FsspecIndexer.get_file_info()` against a stubbed fsspec client and read the log lines and span attributes myself, before and after; the two zeros print different sentences after the change and printed nothing at all before it. New unit tests cover the indexer counts, the per-stage filter counts, and the three discriminating warning sentences themselves. The strings were the part of this change nothing asserted: swapping one indexer arm's text for the other's left the full unit suite green at 1506 passed. Six caplog assertions close that, and both string mutations now fail. Full unit suite is 1507 passed after the fix. Not exercised against a live S3 bucket or on a mirror: this is library code with no deployable surface of its own, so there is no environment to deploy it to -- the platform picks it up when it bumps the dependency.

## Proof

Repro (local): drove `FsspecIndexer.get_file_info()` over a stubbed fsspec client with three listings that all end in zero indexed files -- an empty listing, a listing of a directory marker plus a zero-byte `_SUCCESS` key, and a control listing holding one real file.

Observed on `origin/main` (7e1f22ec): the empty-listing case and the all-dropped case are indistinguishable. `get_file_info()` returns `[]` in both, emits nothing at all, and `run()` takes `total_files=len(files)` off the already-filtered list. The pre-filter count is computed inside the same expression that filters it away and is never held. The glob/size `Filterer` logs one DEBUG line per dropped file and keeps no count. Both causes reach the platform as the same fact.

Failing test before the fix -- `test/unit/test_zero_file_diagnostics.py::TestIndexerReportsWhatTheListingReturned::test_a_listing_our_own_filter_empties_is_reported_as_that_instead` and its siblings, red on the attributes that did not exist:

```
E       KeyError: 'source.listing.returned'      (empty listing)
E       KeyError: 'source.listing.returned'      (listing our own filter empties)
E       KeyError: 'source.listing.returned'      (healthy listing)
E       KeyError: 'filter.indexed.received'      (filters drop everything)
E       KeyError: 'filter.indexed.received'      (filters drop some)
E       TypeError: PipelineStep.__call__() got an unexpected keyword argument 'stage'
```

After the fix, same script, same three cases -- the two zeros now differ:

```
case A (listing returned nothing):
  WARNING nothing to index at s3://acme-bucket/inbox: the listing returned no entries
  span: source.listing.returned=0  source.listing.retained=0

case B (listing returned objects, our filter dropped them all):
  WARNING nothing to index at s3://acme-bucket/inbox: the listing returned 2 entries but
          none of them are files with content (kept only type=file and size>0)
  span: source.listing.returned=2  source.listing.retained=0

case C (healthy listing, control):
  no warning; kept 1 file(s)
  span: source.listing.returned=2  source.listing.retained=1
```

Case C is the row that did not change: it stays silent and returns the same file, which is the point -- the reporting only speaks up on a zero. The new test module passes green, and the full unit suite passes with no regressions.

**The third commit's crash, reproduced.** `TestUncompressExtractingNothingDoesNotCrash::test_zero_extracted_files_exits_instead_of_raising_typeerror` drives `Pipeline._run()` with a filterer configured, a downloader returning one record, and an uncompress step returning `[[]]`, so `clean_results` hands back `None`. Reverting only that commit's eight-line guard (`git checkout 7164a74e -- unstructured_ingest/pipeline/pipeline.py`, which leaves the rest of the branch in place) and re-running the module:

```
uv run --locked --no-sync pytest -q test/unit/test_zero_file_diagnostics.py

guard reverted:  1 failed, 6 passed in 0.99s
this head:       7 passed in 1.41s
```

The failure is the crash itself, not an assertion:

```
unstructured_ingest/pipeline/pipeline.py:263: in _run
    downloaded_data = self.apply_filter(records=downloaded_data, stage="uncompressed")

records = None, stage = 'uncompressed'

    def apply_filter(self, records: list[dict], stage: str = "indexed") -> list[dict]:
        if not self.filter_step:
            return records
>       data_to_filter = [{"file_data_path": i["file_data_path"]} for i in records]
E       TypeError: 'NoneType' object is not iterable

unstructured_ingest/pipeline/pipeline.py:199: TypeError
```

With the guard in place the run exits at the uncompress step instead: the partitioner is never called, `No files to process after uncompressing, exiting` is logged, the filter step is called twice (indexed and downloaded) rather than three times, and no `after filtering uncompressed content` line appears, so the filters are not named for a zero they had no part in.

## Dependencies / merge order

none. Composes with https://github.com/Unstructured-IO/unstructured-ingest/pull/805 (missing credentials as the fourth cause); `git merge-tree` against it is clean on `fsspec.py`, and the only conflict is `CHANGELOG.md`, where both claim 1.11.7 -- whichever lands second re-bumps to 1.11.8.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/806?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tabossert:
Approving. I checked that every new branch is reporting-only: `report_listing_counts` returns before it can warn when anything was retained (`fsspec.py:187-188`), `report_filter_counts` hands `results` back untouched (`filter.py:23-30`), and `record_on_current_span` swallows its own exceptions (`otel.py:85-88`). Three notes, none blocking.

**Pin the warning strings.** The two discriminating sentences at `fsspec.py:190-196` are the proof the PR body leads with, and nothing asserts them. `grep caplog test/unit/test_zero_file_diagnostics.py` comes back empty, and all six tests assert only return values and `recorder.attributes[...]` counts. I checked this by mutation rather than by argument: I swapped the string at `fsspec.py:196` for the other arm's text and ran the full unit suite, 1444 passed, 60 skipped, mutant survives. Three `caplog` assertions close it: "returned no entries", "none of them are files with content", "none of the 2 records survived this stage". `test/unit/connectors/test_sharepoint.py:1293` already uses exactly that idiom, and the logger propagates, so caplog does capture these.

**`apply_filter` can be handed `None` at `pipeline.py:263`.** `clean_results` returns `None`, not `[]`, when everything is falsy (`pipeline.py:169-180`). The download path guards for that at 245-248; the uncompress branch goes straight from `clean_results` into `apply_filter`. With a filterer configured, the list comp at `pipeline.py:199` iterates `None` and you get a bare `TypeError: 'NoneType' object is not iterable`. Without one, 197-198 returns `None` harmlessly and the guard at 264 catches it, so the full trigger is uncompress on, a filterer configured, and every archive extracting to zero regular files (`processes/uncompress.py:32-33`, empty `responses`, no exception). Pre-existing on main, so your call whether it is in scope, but 263 is a line you are editing and the crash sits on the exact "everything got dropped" path this PR exists to name. If you take it, take it as a call-site guard like 245-248 rather than `records or []` inside `apply_filter`, which would log "after filtering uncompressed content" for a zero that filtering had no part in.

**Worth one CHANGELOG sentence: which span to query.** Both attribute families land on the root `ingest process` span, not on `indexer` / `filter step`, for two separate reasons. `FilterStep.__call__` reports after `super().__call__()` returns, and that is `@instrument()`-decorated at `interfaces.py:117`, so the step span is already closed. `IndexStep.run` is a generator function under `@instrument(span_name=STEP_ID)` (`steps/index.py:37-38`), so that span opens and closes before the body ever runs. Both are pre-existing decorator behavior and the `record_on_current_span` docstring is accurate about the ambient-span contract, but whoever writes the dashboard query needs to know it is the run span.

One question, not a finding: your second commit covers callers of `apply_filter`, so subclasses overriding `__call__` are the one case left. Nothing in this repo subclasses `FilterStep`, so it is safe here. Does the platform? An out-of-repo `__call__(self, iterable=None)` would raise TypeError on the unconditional `stage=` at `pipeline.py:200`.

tabossert (approved):

paulkarayan:
Thanks Trevor. Took all three, and the answer to your question is no, the platform doesn't subclass it.

**Warning strings: pinned, and I reproduced your mutation first.** Swapped the fsspec.py:196 arm for the other's text and ran the full unit suite -- 1506 passed, mutant survives, exactly as you found. `grep caplog test/unit/test_zero_file_diagnostics.py` was indeed empty.

Six caplog assertions now, using the `test_sharepoint.py:1293` idiom. Each zero arm asserts its own sentence AND the absence of the other arm's, which is what actually kills a swap; the filter case asserts "none of the 2 records survived this stage"; and the two healthy-run tests assert that nothing warns at all, since reporting-only should be silent when the run worked. Two mutations that used to survive now fail:

```
swap the indexer arms                  1 failed, 6 passed
drop the record count from the filter  1 failed, 6 passed
restored                                       7 passed
```

**`apply_filter(None)`: taken, as a call-site guard.** Reproduced it before writing the fix -- drove `Pipeline._run` with an uncompress step returning `[[]]` and a filterer configured, and got `TypeError: 'NoneType' object is not iterable` at pipeline.py:199, on the line you named. The guard mirrors the post-download one at 245-248 and returns with "No files to process after uncompressing, exiting".

Your reasoning for keeping it out of `apply_filter` is the right call for a second reason as well: `records or []` would also have written `filter.uncompressed.received=0` onto the span, so the dashboard would show a filter stage that never ran. The test asserts `filter_step.call_count == 2` and that no "after filtering uncompressed content" line appears, so the wrong fix fails it too. Pre-existing on main, and I said so in the body rather than folding it into the feature.

**CHANGELOG sentence: added, and I checked the span claim rather than taking it.** Confirmed empirically with an `InMemorySpanExporter` -- an attribute set inside an `@instrument`-decorated generator function lands on the parent, and the step span exports empty:

```
index            {}
ingest process   {'source.listing.returned': 7}
```

The CHANGELOG now says to query the root `ingest process` span, not `indexer` or `filter step`, and names both mechanisms.

**Your question: nothing out of repo subclasses `FilterStep`, because nothing out of repo uses it.** `grep -rn "FilterStep\|Filterer\|unstructured_ingest.pipeline" --include='*.py'` across platform-plugins and platform-etl-orchestration, venvs included, returns zero (the same grep for `unstructured_ingest` hits plenty, so it is not a broken pattern). The platform drives connector objects directly through the plugin shell -- `IndexerJob.run()` in platform-plugins goes from the registry to `indexer.run_async()` with no `Pipeline` anywhere -- so `FilterStep`, `apply_filter` and the whole pipeline layer are CLI-and-tests-only in practice. The `stage: str = "indexed"` default from my second commit is still the right defensive move, it just has no out-of-repo caller to defend.

**On the 1.11.7 collision with https://github.com/Unstructured-IO/unstructured-ingest/pull/805.** Both prepend to CHANGELOG.md line 1, so the second one conflicts textually whatever number it carries -- renumbering now would not save the rebase. Re-ran `git merge-tree` across the two current heads: CHANGELOG.md is the only conflict, `fsspec.py` still auto-merges even though both branches now edit `get_file_info`. My recommendation is 805 first: it is the defect fix with the SND evidence behind it, and this one reads better as diagnostics layered on a source path that already classifies its failures. Then this branch rebases, moves its entry under `## [1.11.8]`, and bumps `__version__` to match. I have not renumbered either branch, so both are still 1.11.7 as you found them.

Full unit suite 1507 passed, ruff check and format clean.

PR #807, opened 2026-09-07
fix(registry): make emits_record_version a tri-state capability marker

micmarty-deepsense:
## Why

`SourceRegistryEntry.emits_record_version` (and its destination twin) default to a real `False`, so a connector that never considered record versions is indistinguishable from one that deliberately opted out. Any downstream consumer that treats an explicit bool as authoritative — the ETL incremental-support check reads this marker ahead of its own allowlists — silently flips behavior for every unannotated connector the moment the installed package carries the field: allowlisted connectors whose entries never opted in (box, github, notion) get their incremental support disabled by a default nobody chose.

The `True` side was never validated either: all 13 `emits_record_version=True` entries trace to one mechanical batch commit (f9230114, #752) that touched no indexer implementations. Auditing each indexer against what it actually emits at index time found three claims that do not hold.

## What

- `emits_record_version` becomes `Optional[bool]` defaulting to `None` on both `SourceRegistryEntry` and `DestinationRegistryEntry`. `None` means "no claim": consumers fall back to their own defaults, the same convention `location_shape` already uses. An explicit `True`/`False` stays authoritative.
- Three unverified `True` claims drop to no-claim:
  - **elasticsearch, opensearch**: the indexers emit batch items with no per-record version (`BatchItem(identifier=...)`, version never set); `_version` is only fetched at download time by the downloader's scan query. There is no index-time record version to claim.
  - **slack**: package records carry a real version (`max(ts, latest_reply, edited.ts)`), but file records reuse the parent message timestamp, which is not a per-file change token. A half-true claim is not a claim.
- Verified `True` entries keep it: s3, azure, gcs (etag), dropbox (content_hash), google_drive (Drive `version`), onedrive, sharepoint (etag), salesforce (SystemModstamp), confluence (page `version.number`), outlook (changeKey; the value read itself is fixed in #802).
- box and sftp stay unannotated on purpose. Their no-version state is a fact about *these* implementations; an explicit `False` here would authoritatively override consumers that vouch for a different implementation registered under the same subtype string.

## Consumer impact (both known consumers verified against their current code)

- ETL incremental check: the marker read guards with `isinstance(marker, bool)`, so `None` falls through to its allowlists — behavior for every unannotated connector is preserved with zero code change there. The three dropped claims also fall back to the allowlists, which matches today's deployed behavior (the deployed pin predates the field entirely).
- platform-api capability derivation: `bool(getattr(entry, "emits_record_version", False))` maps `None` to `False` — degrades to "not capable" rather than crashing or claiming falsely.

## Deliberately not here

- github (`file.sha`) and jira (`fields.updated`) populate real versions but never claimed `True`. Promoting them is a per-connector verification decision, separate from this contract fix.

## Testing

- `test_registry_markers.py`: unannotated entries now assert `None`; a new parametrized test pins the three no-claim entries; the fsspec cohort asserts `True`/`None` explicitly.
- Full unit suite green locally (1456 passed, 60 skipped).
- Downstream proof: the ETL repo's incremental unit tests, run against this branch installed in place of the locked package, go from 6 failed (github/notion/box disabled by the default-False override) to all passing, with no ETL code change.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/807?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.11.11]
+
+### Fixes
+
+- **fix(registry): make `emits_record_version` a tri-state capability marker.** The registry defaulted the marker to a real `False`, so a connector that never considered record versions was indistinguishable from one that deliberately opted out, and a downstream consumer that treats an explicit bool as authoritative overrides its own fallback list with an unreviewed default the moment the field exists in the installed package. The field now defaults to `None`, meaning "no claim": consumers fall back to their own defaults exactly as they already do for the other unannotated capability markers, while an explicit `True`/`False` stays authoritative. Three entries whose mechanically-set `True` from the original capability-marker batch does not hold drop the claim rather than keep a false one: elasticsearch and opensearch only learn a document's `_version` at download time (their indexers emit batch items with no per-record version), and slack's two record kinds disagree: conversation packages emit a real index-time change token, but file records reuse the parent message timestamp, so no single connector-level claim is true. Entries whose indexers verifiably populate a per-record change value (s3, azure, gcs, dropbox, google_drive, onedrive, sharepoint, salesforce, confluence, outlook) keep their explicit `True`; outlook's emitted value itself is made real in a sibling change.
+
 ## [1.11.10]
 
 ### Fixes
--- a/test/unit/connectors/test_registry_markers.py
+++ b/test/unit/connectors/test_registry_markers.py
@@ -174,12 +174,47 @@ def test_unannotated_entry_is_unmarked():
 
 
 @pytest.mark.parametrize("connector_type", ["elasticsearch", "opensearch", "slack"])
-def test_download_time_version_sources_make_no_claim(connector_type):
-    # These indexers do not emit a per-record version at index time (elasticsearch
-    # and opensearch only see _version during download; slack file records reuse
-    # the parent message timestamp), so their entries carry no claim rather than
-    # an authoritative bool either way.
+def test_no_authoritative_version_claim(connector_type):
+    # No single connector-level bool is true for these: elasticsearch and
+    # opensearch only see _version during download (their indexers emit batch
+    # items with no per-record version), and slack's two record kinds disagree
+    # (conversation packages emit a real index-time change token, file records
+    # reuse the parent message timestamp). Their entries carry no claim rather
+    # than an authoritative bool either way.
     assert source_registry[connector_type].emits_record_version is None
+
+
+def test_version_claims_are_exactly_the_verified_set():
+    # Snapshot of every explicit emits_record_version claim in both registries.
+    # The original false claims arrived in one mechanical batch commit that no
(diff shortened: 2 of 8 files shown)


micmarty-deepsense:
Two-axis self-review of this branch at b5389226 (diff vs main 43d3344a), posted for reviewer convenience. The Standards axis checks repo conventions plus a fixed code-smell baseline; the Spec axis checks the diff against this PR's own description. The axes are reported separately on purpose, so one can't mask the other.

## Standards

No documented-standard violations (negative result): the repo's docs cover scaffolding new connectors and say nothing about registry field typing or test-file conventions, so nothing applicable is breached.

Baseline smells, all judgement calls rather than defects:

1. **Possible primitive obsession** (`connector_registry.py:61,89`): `Optional[bool]` encodes a three-way concept (no claim / true / false) as a bare optional bool. It deliberately matches the existing `location_shape` None-as-unset idiom on the same class, so it follows local convention rather than introducing a new pattern.
2. **Possible duplicated code, the one worth a second look** (`connector_registry.py:59-61` and `:87-89`): the `emits_record_version` field plus its identical two-line contract comment repeats verbatim in both entry subclasses. The `RegistryEntry` base already groups the shared capability markers (`location_shape`, `supports_recursion`), and unlike `location_identity` right above (which legitimately differs per subclass), nothing here differs. A one-line hoist onto the base would remove the duplication; happy to apply it if reviewers prefer.
3. The new snapshot test (`test_version_claims_are_exactly_the_verified_set`) deliberately re-encodes per-connector facts asserted elsewhere. That is its stated purpose, forcing any future claim change through a conscious test edit, at the cost that such a change edits two tests instead of one.

## Spec

Clean pass, no findings. Every "What" bullet in the description verified against the diff: `Optional[bool]` defaulting to `None` on both entry classes (`connector_registry.py:59`, `:89`); exactly the three claim drops (elasticsearch, opensearch, slack), with a post-diff grep confirming exactly the ten named keepers still carry `True` and no explicit `False` exists anywhere; box and sftp remain unannotated as stated. The tests the description promises all exist and use identity assertions (`is True` / `is None`); the marker test file passes 31/31 locally. The second commit contains only review-feedback items (contract comments, corrected sftp/slack wording, the claim-set snapshot test), no production-logic change beyond the field semantics themselves.

One caveat, not a finding: the description's cross-repo consumer-impact claims reference code outside this repo and were not re-verified as part of this pass.

**Summary:** Standards, 0 hard violations and 3 judgement calls (worst: the duplicated field+comment across the two entry classes). Spec, 0 findings.

awalker4 (approved):

PR #809, opened 2026-09-10
feat(outlook): add exclude_quoted_history to skip quoted thread history

### LINKED ISSUE #810: RFC: give MIME body replacement one opaque, tested boundary
## Implementation status

Implemented in [PR #809](https://github.com/Unstructured-IO/unstructured-ingest/pull/809), commit [acb34422](https://github.com/Unstructured-IO/unstructured-ingest/commit/acb344224ac57c22ba4f52798a46b85cc0e1cc49). Awaiting merge; keep this issue open until then. The PR description includes `Closes #810`.

- `prepare_body_replacement` and opaque `PreparedReplacement.replace` now own the MIME decisions. Rendering and normalized replacement data are the caller-facing values; the MIME tree and selected part stay private.
- Every accepted replacement copies the private tree and selected part together. Repeated replacement and mutation-then-failure retry tests verify stable state.
- Graph vocabulary and SDK normalization stay in the Outlook adapter. MIME replacement owns missing, empty, and mismatched answer rejection.
- Base64 encoding eliminates boundary regeneration. Multipart fixture tests verify decoded text, unchanged parent boundaries, and preservation of attachments/resources.
- Production and tests use the same preparation/replacement boundary, including conservative protected descendants, flowed-text parameter removal, all six body identity headers, and fallback on parsing/serialization/staging failures.

Validation: 192 Outlook tests passed with SDK 2.6.2 and with SDK 3.0.0 in an isolated overlay. Locked standard unit suite: 1,702 passed; separate unstructured unit target: 2 passed. Repository lint and formatting of all five changed files pass. Repository-wide formatting reports 18 untouched baseline files. The lockfile remains unchanged.

The proposal below records the accepted design and historical problem statement.

## Problem

PR #809 adds optional quoted-history removal to Outlook downloads. MIME selection, rendering validation, preservation rules, and serialization jointly decide whether a downloaded message can be safely rewritten. They need one owner and one production-tested boundary.

The published implementation at c5276248 tests helpers that the downloader does not call, while the downloader independently makes eligibility and rendering decisions. The local refactor inspected during review moves MIME code into a separate module, but its prepared value still exposes mutable message and part objects. Moving code alone does not prevent callers from violating preservation invariants.

Related review: https://github.com/Unstructured-IO/unstructured-ingest/pull/809#issuecomment-5631080812

The design must preserve attachments, related resources, body identity headers, and the original downloaded bytes when replacement is declined or fails. Selection must precede the Graph request because it determines the requested rendering.

## Proposed interface

Use an opaque prepared replacement with a small interface:

```python
def prepare_body_replacement(raw: bytes) -> PreparedReplacement | KeepOriginal: ...

class PreparedReplacement:
    @property
    def rendering(self) -> Rendering: ...

    def replace(self, body: ReplacementBody | None) -> bytes | KeepOriginal: ...
```

`ReplacementBody` carries normalized rendering and text. `KeepOriginal` carries a compact reason for existing logging needs. MIME objects, selected parts, traversal, and serialization remain private. Define and test the prepared object's reuse semantics; callers must not depend on a partially mutated object after failure.

The normal caller sequence is:

```python
try:
    prepared = prepare_body_replacement(raw)
    if isinstance(prepared, KeepOriginal):
        log_skip(prepared)
        return

    body = fetch_unique_body(prepared.rendering)
    result = prepared.replace(body)
    if isinstance(result, KeepOriginal):
        log_skip(result)
        return

    write_atomically(download_path, result)
except Exception as error:
    log_optional_reduction_failure(error)
```

The existing completed download remains the fallback. Keep parsing, external lookup, validation, and rewriting within that optional-operation failure boundary. File replacement must remain atomic, and staging cleanup errors must not fail the record.

## Dependency strategy

The MIME module has in-process dependencies: standard-library MIME handling and text-content inspection. It owns eligibility, body selection, rendering agreement, empty-content rejection, related-root handling, preservation rules, and serialization.

Graph remains a true external dependency in the Outlook adapter. The adapter owns authentication, request hooks, SDK response normalization, and translation between MIME `plain` and Graph `text`. Filesystem operations and logging stay with the downloader. No Graph callback, SDK type, file path, or logger is required by the MIME interface.

## Testing strategy

Exercise `prepare -> replace -> parse returned bytes`, the same boundary production uses. Retain fixtures and verify:

- Named inline attachments are excluded from selection and remain intact.
- Signed/encrypted and bodyless messages are declined before a Graph request.
- Related roots honor `start`; non-root resources and their identities survive.
- Superseded body renderings are removed without entering attached messages or containers.
- All preserved body identity headers survive, including Content-Base and Content-Description.
- Delimiter-like replacement text survives without hiding attachments or adding parts.
- Empty, stylesheet-only, missing, and mismatched replacement bodies are declined.
- Parsing or serialization failures leave the completed download unchanged through the real downloader entry point.

Replace tests of production-unused wrappers with these boundary tests. Keep focused adapter tests for response normalization and real SDK event registration/removal. Keep a small downloader suite for request count/rendering, setting-off behavior, atomic staging failures, and cleanup failures. Avoid using production body-classification predicates as the sole test oracle for attachment preservation.

## Implementation recommendations

Make the prepared object opaque rather than exposing a frozen dataclass containing mutable MIME nodes. Normalize replacement values at the adapter boundary, then enforce acceptance inside the replacement operation so callers cannot bypass it.

Use standard base64 content-transfer encoding for the replacement leaf to prevent literal MIME delimiter lines. A focused check preserved delimiter-like text, an attachment, and the parent boundary. Expand that verification across the preservation fixtures before replacing boundary-scanning/regeneration logic. This avoids a second traversal and random boundary mutation.

Migrate the downloader onto the accepted boundary and remove duplicate selection/acceptance paths. Split tests by MIME transformation and Outlook orchestration where it improves navigation; do not merely distribute the same coordination burden across files.

Scope this RFC to MIME replacement ownership. Preserve the optional feature's behavior, record identity, and Graph request semantics. Any implementation must first reconcile the concurrent local refactor and latest PR head; the review baseline above is historical, not a claim about current branch state.

micmarty-deepsense:
## What changes

Adds `exclude_quoted_history` to the Outlook downloader, off by default. When enabled, the downloader asks Graph for each message's `uniqueBody` and replaces the downloaded MIME body so earlier conversation text is not extracted repeatedly. Eligible messages cost one additional Graph request; protected or bodyless messages do not.

The downloader now uses the same prepared-replacement interface as the MIME tests. The MIME module owns selection, rendering agreement, empty-content rejection, preservation, and serialization. Its prepared object keeps MIME state private and copies that state for each replacement, so retries cannot reuse a partially mutated message. Graph response normalization and the MIME `plain` to Graph `text` translation stay in the Outlook adapter.

Replacement text is base64 encoded. Delimiter-like text cannot create MIME parts or truncate attachments, and parent boundaries do not need regeneration. Named attachments, attached messages, related resources, and body identity headers are preserved. Old `format=flowed` and `delsp` parameters are intentionally removed because the replacement is emitted as fixed text. A signed or encrypted descendant conservatively prevents the entire rewrite.

The completed download remains unchanged when analysis, Graph lookup, validation, serialization, or staging fails. Disk replacement is atomic, and staging cleanup failures cannot fail the record. Missing, malformed, mismatched, and non-visible responses retain the full message. Originally empty messages remain an informational skip rather than a warning.

Closes #810.

## Validation

At `acb34422`:

- Locked SDK 2.6.2: 192 Outlook tests, 1,702 standard unit tests, and 2 tests in the separate unstructured unit target passed.
- SDK 3.0.0 in an isolated uv overlay: 192 Outlook tests passed. The project environment and lockfile were not changed for this check.
- Repository-wide Ruff lint, formatting checks on all five changed files, and `git diff --check` passed. A repository-wide format check flags 18 untouched files; those baseline differences remain outside this change.

Coverage includes delimiter/terminator text across multipart fixtures, exact decoded replacement text, unchanged parent boundaries, related-root resources, all six preserved body headers, repeatability and failure-then-retry, conservative protected-descendant refusal, fixed-text parameters, parse/serialization failures through the downloader, SDK response shapes, and real SDK Prefer event registration/removal.

## Operational limits

Graph decides what counts as unique. Forwarded messages keep the sender's added comment rather than the forwarded text; replies written inside quoted blocks may lose that text. HTML is preferred when both body renderings exist. Enabling the setting does not reprocess already downloaded messages because their record version has not changed.

The Outlook extra permits `Office365-REST-Python-Client>=2.6.2,<4`; validation covers 2.6.2 and 3.0.0, not every future 3.x release. This pass does not establish live Exchange preservation of sender-authored MIME boundaries or OneDrive/SharePoint SDK compatibility. Workflow UI exposure is outside this PR. Record identity and the default download behavior are unchanged.


unstructured_ingest/processes/connectors/outlook.py
```diff
@@ -76,6 +113,177 @@ def _prefer_immutable_ids(request: "RequestOptions") -> None:
...
+    None means the message should be kept as downloaded, which happens when
+    the message has no body part to replace, when it is signed or encrypted, or
+    when `unique_content` carries no text. An empty body legitimately yields an
+    empty uniqueBody, so declining is the correct outcome rather than an error.
+    """
+    if _is_protected(raw):
+        return None
+
+    message = email.message_from_bytes(raw, policy=_MIME_POLICY)
+    body = message.get_body(preferencelist=BODY_PART_PREFERENCE)
```
micmarty-deepsense: [P2] Exclude named attachments when selecting the body

get_body() can select a named inline HTML attachment over the actual plain-text body. I reproduced this with multipart/mixed containing a text/plain body followed by text/html with Content-Disposition: inline; filename="report.html". The reducer replaces the report contents, removes its filename/disposition through set_content(), and then removes the original plain-text body during the sweep.

Please apply the same filename/attachment exclusions during body selection that the sweep uses, consistently here and in _primary_body_part(). A regression test with this structure would cover the attachment-preservation contract.

unstructured_ingest/processes/connectors/outlook.py
```diff
@@ -76,6 +113,177 @@ def _prefer_immutable_ids(request: "RequestOptions") -> None:
...
+    if body is None:
+        return None
+
+    subtype = body.get_content_subtype()
+    # The body part decides whether angle brackets in the value are markup or
+    # just characters, so the emptiness test waits until the part is known.
+    if not _carries_text(unique_content, is_markup=subtype == "html"):
+        return None
+
+    body.set_content(unique_content, subtype=subtype, charset="utf-8")
```
micmarty-deepsense: [P2] Preserve Content-ID when replacing a related body

set_content() removes the selected part's Content-ID. For multipart/related with start="<root>", an image as the first child, and the HTML body carrying Content-ID: <root> as the second child, the parent still references <root> after replacement but that ID no longer exists. I reproduced get_body() returning the body before reduction and None afterward.

Please preserve the body's Content-ID during replacement and add a regression test with an explicit start parameter and a non-first HTML root.

micmarty-deepsense (commented):
I found two MIME edge cases that should be addressed before merging; details are inline. Both were reproduced against the current helper functions.

A few maintainability suggestions, separate from those correctness fixes:

- Parse each message once and pass the parsed message and selected body through the helpers. The successful path currently parses the same MIME four times and repeats some checks.
- Shorten the repeated explanations of uniqueBody, rendering preferences, and fallback behavior. Brief comments about header refolding, Graph's text/plain terminology difference, and attachment traversal boundaries are useful and worth keeping.
- Consolidate the repeated child-preservation branches in _strip_superseded_renderings while retaining the attachment safeguards.
- Keep the behavioral test cases, but trim commentary that repeats their names. The test asserting that the SDK has no typed accessor can be removed; adding an accessor should not break a working raw-property lookup.

CI unit tests and lint are passing on Python 3.11–3.13. I did not run the full test suite locally; validation of the two findings used the unchanged helper functions with the standard email library.

unstructured_ingest/processes/connectors/outlook.py
micmarty-deepsense: Fixed in 69b0d75f. Body selection now uses one shared MIME-aware selector that excludes attachment-disposition and named parts while preserving multipart/related root selection. Added regressions for both selection and replacement of a named inline HTML attachment.

unstructured_ingest/processes/connectors/outlook.py
micmarty-deepsense: Fixed in 69b0d75f. The reducer now saves and restores the selected body part Content-ID around set_content(). Added a multipart/related regression with an explicit start parameter and a non-first HTML root, including a post-reduction get_body() assertion.

unstructured_ingest/processes/connectors/outlook.py
```diff
@@ -355,12 +563,129 @@ def _download_message(self, file_data: FileData, download_path: Path) -> None:
...
+        try:
+            staged.write_bytes(_replace_body(message, body, content))
+            staged.replace(download_path)
+        except Exception as e:
+            logger.warning(
+                f"Could not write the reduced body of message {message_id}, "
+                f"keeping the full body including any quoted history: {safe_error_summary(e)}"
+            )
+        finally:
+            staged.unlink(missing_ok=True)
```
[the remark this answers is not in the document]
micmarty-deepsense: Fixed in c5276248. Cleanup now suppresses OSError from unlink(), so it cannot escape after a staging failure. Added a regression where write_bytes() and unlink() both fail and confirmed the original download remains usable.

awalker4 (approved):

micmarty-deepsense:
Solid feature, and "leave the file exactly as downloaded if anything goes wrong" is the right design goal. One issue is worth fixing before merge; the rest are follow-ups.

## A reply containing the message's own MIME boundary can silently break the message

`outlook.py:244`, `_replace_body`

`set_content(content, subtype=..., charset="utf-8")` writes the text unescaped under a 7bit/8bit encoding, so a line in the returned body that matches the message's own delimiter is re-read as a real one. Two outcomes, both silent (no exception, no log, no parser defect):

- a line `--<boundary>` injects a spurious `text/plain` part and truncates the kept body
- a line `--<boundary>--` truncates the message to a single part; an attachment after that point is still in the file but unreachable to any parser

Reproduced on `example-docs/eml/email-with-image.eml`: 3 leaf parts before, 1 after the terminator case, with the 215KB PNG orphaned.

This needs a crafted or unusual message, so it is low-likelihood, but it is silent data loss and the guard is cheap. Suggest checking the new content for `--` plus any ancestor boundary and declining on a match, or forcing a content-transfer-encoding that cannot emit a literal boundary line.

## Two malformed inputs raise instead of degrading

`outlook.py:577-633`

Only the `_fetch_unique_body` call is inside the try/except. A deeply nested message raises `RecursionError` in the parse at line 587, before the guard; and a non-string `content` in the response survives the `or ""` at line 688 and raises in `_has_visible_text` at line 613. In both cases the complete file is already on disk, so widening the guard to cover the whole analysis phase would match the behaviour a failed request already gets.

## Follow-ups, not blocking

- `_reduce_message_body` and `_primary_body_part` have no caller on the download path, which now re-derives the same decision independently. Worth routing the real path through the tested function, or moving the tests onto the real entry point, before the two drift further.
- `_drop_other_body_renderings` does not carve out `multipart/related` the way `get_body()` does, so a non-root text part can be swept and leave a `cid:` reference dangling.
- The Prefer-header deregistration is never exercised against a real request event; every test uses a mock client. Harmless today since each download gets a fresh client.
- `_KEPT_BODY_HEADERS` lists six headers; the test class covers four. `content-base` and `content-description` could be dropped silently.

unstructured_ingest/processes/connectors/outlook_mime.py
```diff
@@ -0,0 +1,279 @@
...
+    MIME-Version a sub-part should not carry, so the headers describing this
+    part's identity and place in the message are put back afterwards.
+    """
+    part = plan.part
+    preserved = [
+        (name, value) for name, value in part.items() if name.lower() in _KEPT_BODY_HEADERS
+    ]
+    had_mime_version = "MIME-Version" in part
+
+    part.set_content(content, subtype=plan.rendering.mime_subtype, charset="utf-8")
```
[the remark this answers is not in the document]
micmarty-deepsense: Keeping the existing behavior deliberately. The replacement is newly supplied text emitted as fixed text, not a format=flowed encoding of the old body. Carrying over format=flowed/delsp would tell readers to join or delete characters around line breaks that the replacement writer did not encode that way. See [RFC 3676 section 4](https://www.rfc-editor.org/rfc/rfc3676.html#section-4).

Pinned in [acb34422](https://github.com/Unstructured-IO/unstructured-ingest/commit/acb344224ac57c22ba4f52798a46b85cc0e1cc49): [the flowed-text regression](https://github.com/Unstructured-IO/unstructured-ingest/blob/acb344224ac57c22ba4f52798a46b85cc0e1cc49/test/unit/connectors/test_outlook_mime.py#L744) checks that both parameters are absent and the exact decoded replacement lines survive, including a line ending in a space. Resolving as an intentional encoding choice, not applying the suggested parameter copy.

unstructured_ingest/processes/connectors/outlook_mime.py
```diff
@@ -0,0 +1,279 @@
...
+
+def plan_body_replacement(raw: bytes) -> Union[BodyReplacement, KeepFullBody]:
+    """What replacing the body of `raw` would rewrite, or why it must not be tried.
+
+    This is the only place that decides which part a partitioner will read and
+    which rendering has to be asked for, so the caller never has to derive
+    either for itself.
+    """
+    message = parse(raw)
+    if any(part.get_content_type() in _PROTECTED_CONTENT_TYPES for part in message.walk()):
```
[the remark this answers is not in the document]
micmarty-deepsense: Keeping the whole-tree refusal deliberately. The serializer processes the complete MIME tree, including protected descendants, even when the replacement selects only the outer body. We have not established byte preservation of signed descendants through that serialization, so narrowing the guard would weaken the conservative fallback. Keeping quoted history is preferable to risking the integrity of a protected attachment.

Pinned in [acb34422](https://github.com/Unstructured-IO/unstructured-ingest/commit/acb344224ac57c22ba4f52798a46b85cc0e1cc49): [the downloader regression](https://github.com/Unstructured-IO/unstructured-ingest/blob/acb344224ac57c22ba4f52798a46b85cc0e1cc49/test/unit/connectors/test_outlook_quoted_history.py#L607) covers both signed and encrypted attached descendants and asserts byte-identical original output with no uniqueBody lookup. Resolving as intentional conservative behavior.

micmarty-deepsense:
Follow-up to [the MIME review](https://github.com/Unstructured-IO/unstructured-ingest/pull/809#issuecomment-5631080812): the fixes and RFC #810 implementation are now in `acb34422`.

- Delimiter collisions are prevented by base64 encoding the replacement leaf. Fixture tests assert complete decoded text, preserved attachments/resources, and unchanged parent boundaries.
- Parsing failures and invalid Graph response content retain the completed download. Rendering acceptance and empty/style/script-only rejection now belong to the prepared replacement used by production and tests.
- Related resources are preserved; all six body identity headers and real SDK Prefer-hook cleanup are covered.
- The two conservative behaviors discussed in the remaining Cubic threads are retained and regression-tested: discard obsolete flowed-text parameters, and decline messages with protected descendants.

Validation: 192 Outlook tests passed on both SDK 2.6.2 and an isolated SDK 3.0.0 overlay; 1,702 standard unit tests and 2 separate unstructured unit tests passed on the locked environment. Repository lint and changed-file formatting pass. Repository-wide formatting still flags 18 untouched files.

The PR description links #810 for closure on merge. The existing approval was preserved; no new review was requested.

PR #812, opened 2026-09-20
fix(teradata): classify row-rejection codes as the customer's, not a connect failure

paulkarayan:
## Why

A row the database read and refused came back to the user as a failure to reach the database.

`upload_dataframe` hands every `teradatasql` failure to `_raise_classified_teradata_error`, which raises `UserError` only for codes in `_USER_FAULT_TERADATA_CODES` and routes everything else into `DestinationConnectionError(_summarize_error(...))`. `_summarize_error` has branches for a timeout, a refusal and a credential rejection, and a catch-all that returns the bare `Failed to connect to server {host}`.

The row-rejection family was not in the map, so it landed on that catch-all. A user whose destination table rejects a value gets told we could not reach their server, and goes looking at DNS, firewalls and egress for a problem that is in their data or their column definitions. The message is also indistinguishable from a real transport failure, so neither they nor we can tell the two apart from the error alone.

These are as user-fixable as the privilege and syntax errors already in the map: fix the value, or fix the column it is going into. Retrying them is pointless, which is the other thing the map decides.

## What

Five codes join `_USER_FAULT_TERADATA_CODES`, so they raise `UserError` (422) naming the code instead of `DestinationConnectionError` (400) with the connect-failure summary:

| code | descriptor |
|---|---|
| 2621 | bad character in format or data |
| 2665 | invalid date |
| 2666 | invalid date |
| 5407 | invalid operation for datetime or interval |
| 6706 | the string contains an untranslatable character |

The descriptors mirror Teradata's own message wording with the interpolated identifiers removed, and that is now written down as the rule for anything added to this map. `Invalid date supplied for {table}.{column}` becomes `invalid date`. Two reasons: a user can match what we print against the same error in their own DBS log or DBQL, and we are not inventing a second vocabulary for errors Teradata already names.

The map is consulted at every `_raise_classified_teradata_error` call site, not just the upload one, so `TeradataIndexer.precheck` now raises `UserError` (422) for these codes too, instead of `SourceConnectionError` (400) with the historical `table 'X' not found or not accessible` context. A source-side caller catching `SourceConnectionError` no longer catches those cases. On a `SELECT TOP 1 *` probe only 6706 and 2621 are realistic.

What still must not cross is the part the driver interpolates, since the raw message names the table, the column and the offending value, and the Go driver wraps it in text that carries host, user and password. Only the numeric code and the fixed descriptor reach the response, and the chain stays suppressed so traceback logging cannot resurface it either.

## Deliberately not here

- **A code we have not classified keeps today's behaviour.** Widening the map is not a claim that every server-side error is the user's fault, so an unlisted code still takes the connection path rather than having an audience guessed for it. That boundary has its own test.
- **2801 (a duplicate value for a unique prime index) is deliberately unlisted, and that is a change from the first version of this PR.** `upload_dataframe` commits each batch separately, so when `can_delete()` is False a retry after a partial upload re-inserts rows an earlier batch already committed. The duplicate is then ours, not the customer's. platform-plugins already pins 2801 as a provider-side error on the byte-identical wire string, and classifying it here would override that decision from upstream, because a `UserError` raised in ingest never reaches the plugin's own classifier.
- **2667 is not added.** It is the same family as 2665/2666 but no verified message text was found for it, and a descriptor is not worth guessing.
- **3811 (a NOT NULL column given no value) and 2616 (numeric overflow) are the same family and are not added.** They are plausible INSERT-time rejections, but neither was among the codes this change set out to cover, and each deserves the same check against real message text before it earns a descriptor.

## Testing

- `test_upload_dataframe_row_rejection_raises_user_error` drives the real INSERT path for each of the five codes and asserts the code, the descriptor, the 422 and the absence of any connect story.
- `test_row_rejection_never_surfaces_the_driver_text` appends a credential to the driver message and asserts it does not survive, along with the table and column the driver named, and that the chain stays suppressed (`__suppress_context__` is True, which is what `raise ... from None` actually sets; `__context__` still holds the driver exception).
- `test_source_table_probe_row_rejection_raises_user_error` drives `TeradataIndexer.precheck` and pins the source-side path described above: same 422, same code, same descriptor.
- `test_an_unlisted_code_still_takes_the_connection_path` pins the boundary above, and 2801 is now one of the codes it covers.
- `test_the_descriptor_is_a_short_fixed_phrase` pins each descriptor, so a correction to the wording is a one-line edit and cannot drift silently.
- Reverting `teradata.py` to its pre-change state fails 20 of the 21 new parametrized assertions; the boundary test is the one that must pass on both sides. With the change, `test/unit/connectors/sql/test_teradata.py` is 127 passed, and the full `test/unit` target is green apart from `test/unit/interfaces/test_downloader.py::test_downloaded_file_atime_matches_remote_created_time`, which fails the same way on `main` because macOS does not preserve the atime it sets.

Note on how the descriptors were checked: Teradata's own message pages would not render to a fetch, so each string was corroborated against verbatim quotations in Teradata support and knowledge-base material. 6706 and 5407 are quoted word for word there; 2621, 2665 and 2666 are quoted with their interpolated identifiers attached. Anyone with the Messages manual to hand is welcome to tighten the wording, and the test above is where to do it.

## Dependencies / merge order

none

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/812?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tabossert (approved):
Approving. The change is well-scoped, the version bump is right, and the tests genuinely bind it: reverting `teradata.py` fails 15 of the 16 new assertions, with only the boundary test passing on both sides, exactly as the PR body claims. Everything below is non-blocking.

Method note, since it changes how much weight to give these: I ran eight candidate findings through an adversarial refutation pass and six were killed, including two that had come back marked blocking. What survived is below, plus what I re-verified by hand. I have listed the killed ones at the end so nobody re-derives them.

---

## 1. The column-qualified invalid-date rejection is 2666, not 2665

`teradata.py:102` adds `2665: "invalid date"`. That entry is correct and should stay: 2665 is a genuine user-fault code and "invalid date" is exactly its message text.

But 2665 is the bare `Invalid date.` It comes from INTERVAL year/month arithmetic, TPT, NOS row building and bulk loaders. The column-qualified form, `Invalid date supplied for {table}.{column}`, is **2666**, and 2666 is what a driver batch insert reports (turbodbc #157: `native error code: -2666 ... Invalid date supplied for LOAD_TEST.PAID_THROUGH_DATE`). Teradata publishes a KB titled "Informatica query failing with Teradata Database error: 2665 and 2666 Invalid date", so they are two distinct errors.

Against the classifier on this branch:

```
2666 in map? False
2666 -> DestinationConnectionError(status=400): Failed to connect to server db.example.com
2665 -> UserError(status=422): Teradata error 2665 (invalid date) for 'test_table'.
```

The pairing is recorded in three places:

- `test/unit/connectors/sql/test_teradata.py:1859-1860` pairs code 2665 with 2666's message text: `"[Teradata Database] [Error 2665] Invalid date supplied for test_table.last_modified."`
- `teradata.py:84-85` promotes it to the rule for the whole map: `"Invalid date supplied for {table}.{column}" becomes "invalid date"`.
- The PR body says "2621 and 2665 are quoted with their interpolated identifiers attached." 2621 genuinely is; 2665 is not.

The test is not vacuous, since the classifier keys only on the `[Error NNNN]` tag and never reads the message body, so it does correctly prove 2665 -> `UserError`. The cost is that the documented rule teaches the next person adding a code something untrue.

Suggestion: add 2666, fix the fixture message and the comment, and consider 2667 (the invalid-time sibling) while you are in there.

## 2. CHANGELOG does not mention the source-side behaviour change

The entry attributes the change to `upload_dataframe`, but the map is consulted at all five `_raise_classified_teradata_error` call sites, including `TeradataIndexer.precheck` (`teradata.py:363`, `direction="source"`), where a listed code now raises `UserError` instead of `SourceConnectionError` with the historical `table 'X' not found or not accessible` context.

The PR that created this map (d11f4ee7) called it out explicitly:

> **Source-side behavior change:** The classification applies to both directions. `TeradataIndexer.precheck` previously always raised `SourceConnectionError` ... Source-side callers that catch `SourceConnectionError` will no longer catch those cases.

Practical impact is low. No caller in this repo catches `SourceConnectionError` on a Teradata path, and realistically only 6706 and 2621 can fire on a `SELECT TOP 1 *` probe. Raising it because the precedent set the convention.

## 3. `assert excinfo.value.__cause__ is None` does not pin chain suppression

`test_teradata.py:1951`. The PR body cites this as the test backing "the chain stays suppressed so traceback logging cannot resurface it either." It does not pin that. Run in this branch's venv:

```
WITH from None       __cause__=None  __suppress_context__=True   __context__=ValueError('driver text password=hunter2')
WITHOUT from None    __cause__=None  __suppress_context__=False  __context__=ValueError('driver text password=hunter2')
```

`raise X` inside an `except` block never sets `__cause__`, so deleting `from None` at `teradata.py:181` leaves this assertion green. `assert excinfo.value.__suppress_context__ is True` is the one that pins it.

Worth noting the third column separately: the raw driver exception, password included, stays reachable as `UserError.__context__` in both cases. `from None` only asks default traceback formatting to hide it, so any logger or error-tracking SDK that walks `__context__` itself still sees the driver text. The guarantee is narrower than "cannot resurface it."

The same assertion already exists at `test_teradata.py:1323`, so this is propagated rather than introduced by this PR.

## 4. The 2801 descriptor is not Teradata's wording, and is narrower than the error

`teradata.py:103` has `2801: "duplicate value for a unique primary index"`. The fixture quotes the server text at `test_teradata.py:1865` as `Duplicate unique prime key error in test_table.` Applying the rule stated at `teradata.py:82-85` would give "duplicate unique prime key error".

It is also narrower than the error: 2801 fires for Unique Secondary Indexes and for `UNIQUE` / `PRIMARY KEY` constraints, which Teradata implements internally as USIs, not only for the UPI. Since the descriptor is the only free text that crosses, a customer whose USI rejected the row is pointed at a primary index that is not the one that rejected it.

## 5. Sibling PR collision, for coordination rather than for this PR

- #811 rewrites the exact line this PR extends: `_USER_FAULT_TERADATA_CODES: Mapping[int, str] = MappingProxyType({` becomes `MappingProxyType(`. Both PRs also append to `test/unit/connectors/sql/test_teradata.py`. Guaranteed conflict whichever merges second.
- #811 is bumped to 1.11.13, which `main` already carries, so it needs a re-bump regardless of ordering.
- #813 declares itself "Blocked by: #812". This PR's "Dependencies / merge order: none" is accurate for itself, just noting #813 sits downstream.

---

## Two questions, neither gating

1. 2666, 3811 and 2616 are all the same INSERT-rejection family. The PR defers 3811 and 2616 pending message-text verification, which is a reasonable line. Now that 2666 has been checked, is there a reason to land it separately rather than with these five?
2. 5407 and 2621 look like statement-evaluation errors rather than row rejections: 5407 from datetime/interval arithmetic and FORMAT mismatches, 2621 from CHAR-to-numeric conversion. Every statement this connector sends is parameter-bound with no datetime expression, so the realistic route to either is a customer-precreated column or a session date-form mismatch. Does the CHANGELOG's "a row the server refuses" framing hold for those two?

---

## On the cubic comments

- **Redaction test has no offending row value**: fair, though minor. The `CONNECTION=` / `hunter2` / `db.example.com` assertions in that test are load-bearing for all five cases, so it is not toothless. A regression leaking only the value would pass.
- **No source-path test for the new codes**: fair, low impact. See point 2. Realistically only 6706 is reachable through `SELECT TOP 1 *`.
- **The `for '<table>'` suffix violates the redaction contract**: I do not think this one holds, and I would push back on it. `table` comes from `upload_config.table_name` / `index_config.table_name`, the connector's own configuration, never from `str(exc)`. The module already prints host and table on the `SourceConnectionError` fallback, with a test that deliberately uses `password_reset_log` to pin exactly that. What is true is narrower and is a docs fix: the docstring at `teradata.py:141-144` says "only the numeric code and the fixed descriptor", which the `target` suffix contradicts.

---

## Checked and dropped

Recording these so they do not get re-raised later:

- **"No log line carries the Teradata code."** Inverted. For the five codes this PR maps, the code now appears in the raised message and the pipeline traceback where it appeared nowhere before. The real residue is that the unmapped fallback branch discards `_extract_teradata_error_code(exc)`, which is pre-existing and a separate follow-up.
- **`matches[-1]` in `_extract_teradata_error_code` versus multi-code batch errors.** The driver does build a set of all `[Error N]` tags at `teradatasql/__init__.py:1082` (20.0.0.66), but that path runs with an empty ignore-set, so it does not establish that a real `executemany` error string ever carries more than one tag. Not provable as stated.
- **"The new test assertions are vacuous."** `assert "test_table" in message` at `:1918` is in fact the only assertion that kills the dropped-`target` mutant, failing 5 of 5 params without it.
- **"2801 blames the customer for a connector-inflicted duplicate."** The customer-facing message carries no remedy at all, by design and pinned by a test. The remedy wording lives only in a source comment.
- **"6706 may be a connector-side charset defect."** This rested on 8555f1a4, which `CHANGELOG.md:615` records as fixing Error **6705** (non-BMP characters), not 6706. Unicode Pass Through was never the mechanism that stops a LATIN column rejecting a non-Latin string, and Teradata documents 6706 firing with UPT on.

Local verification: `pytest test/unit/connectors/sql/test_teradata.py` gives 122 passed on this head. The full `test/unit` run cannot complete on a laptop (21 collection errors from missing optional deps: `faker`, `pyiceberg`, `htmlBuilder`, `networkx`), which is pre-existing and unrelated, and CI's `test_ingest_unit` is green on 3.11, 3.12 and 3.13.

paulkarayan:
# PR 812 review response (draft, not posted)

Two commits on `pk/teradata-widen-user-fault-codes`, local only:

- `a329d66e` fix(teradata): map 2666, not 2801, in the user-fault code set
- `be67d345` test(teradata): pin what the classified error really redacts, and the source path

The classified set is now 2621, 2665, 2666, 5407, 6706. 2801 is out, 2666 is in, so this is a
different change from the one you approved.

---

## cubic: the redaction test never includes an offending row value

FIXED in `be67d345`. The driver message each case feeds in now carries a distinctive value
(`test_teradata.py:1855`, appended at `:1940`) and the test asserts it does not cross
(`:1959`). A mutant that appends only that value to the `UserError` message, leaving everything
else redacted, fails all five cases now and passed before.

## cubic: no source-path test for the new codes

FIXED in `be67d345`. `test_source_table_probe_row_rejection_raises_user_error`
(`test_teradata.py:1988`) drives `TeradataIndexer.precheck` for every code in the set and asserts
the 422, the code, the descriptor, and that the historical `table 'X' not found or not
accessible` context is gone. Realistically only 6706 and 2621 fire on a `SELECT TOP 1 *`; the
rest are pinned so the two directions cannot drift apart.

## cubic: the `for '<table>'` suffix violates the redaction contract

NOT CHANGED in the code; the docstring it contradicted is fixed. `table` is
`upload_config.table_name` / `index_config.table_name`, the connector's own configuration, and is
never read out of `str(exc)`. The module already prints the table on the connection-error
fallback with a test pinning it. What was wrong was the docstring's "only the numeric code and
the fixed descriptor", now corrected to name the caller-supplied table as the third thing that
crosses (`teradata.py:153-160`, `:190-194`).

## 1. The column-qualified invalid-date rejection is 2666

FIXED in `a329d66e`. 2666 is in the map at `teradata.py:115`, the fixture message for 2665 is now
the bare `Invalid date.` and 2666 carries the column-qualified text (`test_teradata.py:1865-1875`),
and the rule above the map says which error is which (`teradata.py:84-86`). Both descriptors are
`invalid date`, the way 3706 and 3707 both map to `SQL syntax error`: the code is what
disambiguates in their DBS log, and removing the interpolated identifiers from 2666's text leaves
nothing else to say without inventing wording.

Not adding 2667. There is no verified message text for it in hand, and that is the bar the map's
own rule sets.

## 2. CHANGELOG does not mention the source-side behaviour change

FIXED in `be67d345`. Second bullet under 1.11.14, in the same terms d11f4ee7 used, including that
source-side callers catching `SourceConnectionError` no longer catch those cases and that only
6706 and 2621 are realistic on the probe.

## 3. `assert __cause__ is None` does not pin chain suppression

FIXED in `be67d345`. The assertion is now `__suppress_context__ is True`
(`test_teradata.py:1967`); deleting the `from None` fails all five cases, and passed before. The
docstring claimed more than the language gives, so it now says `from None` keeps the driver
exception out of default traceback formatting and nothing more, with `__context__` still holding
it (`teradata.py:167-172`).

The identical assertion at `test_teradata.py:1323` is untouched. It predates this PR and fixing
it is not in this change.

## 4. The 2801 descriptor is not Teradata's wording

2801 is removed instead, in `a329d66e`. Your blocking comment on platform-plugins 2404 is the
reason, and it holds here for a second reason that is in this file: `upload_dataframe` commits
each batch independently, and when `can_delete()` is False a retry re-inserts rows an earlier
batch already committed, which is documented in the comment right above the loop. So the
duplicate-key rejection can be this connector's doing. `teradata_vector_v2` pins the same verdict
on the byte-identical wire string (`test_duplicate_registration_is_not_charged_to_the_user`,
503 not 400), and mapping 2801 here would have overridden that from upstream no matter what
allowlist 2404 lands with, since ingest raising `UserError` means the plugin never reaches its own
classifier. The map's stated boundary covers it: an audience we cannot determine from the code
stays unlisted. Rationale is in the map comment at `teradata.py:88-93`.

## 5. Sibling PR collision

NOT CHANGED, acknowledged. 811 and 812 conflict whichever merges second, in
`_USER_FAULT_TERADATA_CODES` and in the test file. I have left the hugging-parens formatting
alone here so the conflict stays mechanical. 811 needs the re-bump off 1.11.13 either way, and
813 sits downstream of this one.

## Question 1: 2666, 3811, 2616

2666 landed here rather than separately: it is the code a driver batch insert actually reports for
an invalid date, so leaving it out meant the INSERT-path case this PR exists for still took the
connection path. 3811 and 2616 stay deferred on the same rule as before, real message text first.

## Question 2: does "a row the server refuses" hold for 5407 and 2621

No, and the CHANGELOG and the map comment now say "value-rejection": the server refused the value
being written against the customer's own column definition or their session's date form. That
still puts the fix with the customer, which is the part the classification turns on.

---

## Testing

`uv run --locked --no-sync pytest test/unit/connectors/sql/test_teradata.py`: 127 passed, up from
122. `ruff check .` clean.

Full `test/unit` (the documented target, minus `test/unit/unstructured`): 1 failed, 1728 passed.
The failure is `test/unit/interfaces/test_downloader.py::test_downloaded_file_atime_matches_remote_created_time`,
in a file this branch does not touch, from macOS not preserving the atime the test sets.

Binding still holds: reverting `teradata.py` to its pre-change state fails 20 of the 21 new
parametrized assertions, and the boundary test passes on both sides.

PR #813, opened 2026-09-20
test(teradata): add live integration coverage for the SQL destination

paulkarayan:
## What & why

**Problem:** The Teradata destination has a full suite of tests and has never once talked to a Teradata. Everything it asserts about what the server returns is asserted against an exception we construct ourselves, so the connector we ship to an OEM partner who runs a quality tracker against us is guarded only by fixtures we wrote. The SQL suite here covers Postgres, SingleStore and SQLite on containers, Snowflake on an emulator and Databricks on a real workspace; Teradata has none, because it has neither a container image nor an emulator. The gap is not hypothetical: the defect this follows was an error-classification bug on the destination write path, and nothing in the repo could have caught it.

**Change:** Adds the first integration coverage for that path: an upload test, and two tests that provoke a real server-side rejection and assert how we classify it. Gated on `TERADATA_*` and skipped without them, following the Databricks precedent in the same directory.

**Blast radius:** 1/5 -- test files, fixtures and a version bump; no application code; reverts by deleting two files.

## Linked ticket

none

## Impact

Local tooling and test coverage only. No product code is touched, nothing ships to any customer environment, and no CI job changes: `.github/workflows/e2e.yml` is deliberately not edited, so no job gains a dependency and no secret is added. The effect is on us: for the first time someone with credentials can verify the Teradata destination end to end, including that a rejected row is reported as the user's to fix rather than as an unreachable host.

## Risk / rollback

Delete the two new files. Nothing depends on them.

## Why it asserts failures, not just a row count

Every other SQL destination test here uploads and checks `SELECT count(*)`. That pattern would have stayed green straight through the defect this follows, because the write was never the broken part: a row the server inspected and refused was classified rather than written, and the classification was wrong. So the file provokes two rejections, each chosen because the connector cannot quietly sanitise it away:

| provocation | code | why it survives the connector |
|---|---|---|
| a repeated `id` against a `UNIQUE PRIMARY INDEX` | 2801 | both rows are individually valid and only collide at the server |
| a character with no `LATIN` representation | 6706 | the column's character set, not the value's shape, is what refuses it |

Both assert `UserError` at 422 naming the code, and that the host, the credentials the driver embeds in its connection text, and the server's own interpolated message are all absent. Against a real server that last part doubles as a leak check on a real driver message rather than a fabricated one, which is the thing no unit test can do.

One DDL file, not two: `schema_strict.sql` exists to refuse and says so, while the upload path no longer pre-creates its table at all. `create_destination()` builds it from the connector's own `assets/teradata_elements_schema.sql`, so those lines run for the first time and there is no second copy of the schema to drift.

## How it was verified

The three live tests have NOT been run against a Teradata server; there is no reachable one, which is the reason this file is worth adding and also means it is unrun code. What was checked: the module collects and is picked up by `--tags sql`; `scripts/check_untagged_tests.sh` reports zero untagged for this module (run with the project venv on PATH it reports 63 repo-wide, none of them teradata); with the four variables unset the three live tests skip with the intended reason rather than erroring; the stager test passes against a fixture generated with `OVERWRITE_FIXTURES=true` and then re-validated without it; `ruff check` and `ruff format --check` are clean.

One trap worth recording for whoever runs that guard next: `check_untagged_tests.sh` shells out to a bare `pytest`, so it silently falls back to the system Python, dies on a missing import, prints `0` and exits `0`. It has to be run with the project venv on `PATH` to mean anything.

## Proof

Proof: n/a (exempt -- tests, fixtures, a generated stager fixture and a version bump; no product code in the diff, so there is no behaviour to reproduce or prove). The `.proof.toml` records what was checked anyway, so "exempt" does not read as "unchecked".

## Dependencies / merge order

Blocked by: https://github.com/Unstructured-IO/unstructured-ingest/pull/812 -- that PR adds 2801 and 6706 to `_USER_FAULT_TERADATA_CODES`. Until it merges, both rejection tests assert a classification the code does not yet make. Nothing goes red in the meantime, because these tests skip without credentials and CI never runs them, but merging this first would commit a test that cannot pass. The module docstring says the same thing, so a runner who supplies credentials early is not left hunting.

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-ingest/pull/813?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tabossert (approved):
Approving. No product code, no CI job gains a dependency, and it reverts by deleting files, so none of the below blocks the merge. But the file is unrun by design, and a few things in it don't yet do what the docstring says they do. Everything here is non-blocking; take or leave as you see fit.

I ran the suite locally to check the claims in "How it was verified": module collects (4 items), the three gated tests skip with the intended reason, `test_teradata_stager` passes, and collection works without `teradatasql` installed because the driver import is lazy. All of that holds up.

---

### 1. The upload test pairs the flattened stager with the JSON-blob schema

`test_teradata.py:193` builds `TeradataUploadStager()`, which defaults to `metadata_as_json=False`. `destination_table` creates the table from `schema.sql`, which is the `metadata_as_json=True` schema. The connector says so twice:

- `teradata.py:488-489` - "the stager must have `metadata_as_json=True` to match"
- `teradata.py:497-498` - "Requires the stager to have `metadata_as_json=True`"

Staged fixture has 12 keys; the table has 7 columns. `_fit_to_schema` (`teradata.py:618` -> `sql.py:404-419`) drops `filename, filetype, languages, page_number, orig_elements, is_continuation` and null-fills `metadata`. The only assertion is `count_rows == expected` (`:206`), which is unaffected by column loss, so this stays green even though `orig_elements` never lands.

Worth noting the sibling schemas are all wide, matching their flat stagers: databricks 43 columns, singlestore 40, snowflake 40, sqlite 37, postgres 35. Teradata's is 7.

Suggest `TeradataUploadStager(upload_stager_config=TeradataUploadStagerConfig(metadata_as_json=True))` plus one content assertion, the way `test_postgres.py:169-186` round-trips `expected_text` and `test_embedding`.

### 2. The comment at `:64-65` contradicts the docstring 57 lines above it

```python
# Codes the server raises for the two rejections provoked below. Both are in
# `_USER_FAULT_TERADATA_CODES`, so both must surface as UserError naming the code.
```

They aren't. `teradata.py:73-82` holds only 3807, 3523, 3706, 3707, 3753, 3754, 5612, 5315, and this PR doesn't touch that file. The module docstring at `:7-11` says exactly the opposite and is the correct one.

To be precise about the blast radius, since the docstring already is: nothing is red today. Both tests are `@requires_env(*REQUIRED_ENV)`, which is `pytest.mark.skipif`, and no `TERADATA_*` value appears anywhere under `.github/`, so they are collected under the `sql` tag and skipped. They fail only for a runner who supplies credentials before #812 lands.

One thing that may be worth a deliberate decision rather than an assumption: #813 is not git-stacked on #812. `cdb6220e`'s only parent is `23c52278`, which is the current tip of `origin/main`, and #812's head has that same parent, so they are siblings. Merging this first is safe for CI but leaves `main` carrying two tests that fail for anyone with credentials, plus a CHANGELOG gap at 1.11.14, the slot #812 takes.

### 3. When classification falls through, the failure reads as a connect error

Unclassified codes go to `_summarize_error`, which matches none of its branches for a 2801 message and returns `teradata.py:194`: `f"Failed to connect to server {host}"`. `DestinationConnectionError` is a sibling of `UserError`, not a subclass, so `pytest.raises(UserError)` at `:230`/`:260` won't catch it and `assert_safe_user_error` never runs.

So the failure mode of the test is the exact misdiagnosis the test exists to prevent, and the docstring's explanation isn't in the output. Gating on the map itself would make it self-clearing:

```python
needs_widened_code_map = pytest.mark.skipif(
    not {DUPLICATE_KEY_CODE, UNTRANSLATABLE_CHARACTER_CODE} <= set(_USER_FAULT_TERADATA_CODES),
    reason="needs 2801/6706 in _USER_FAULT_TERADATA_CODES (#812)",
)
```

### 4. The leak assertions can't fail

This is the one I'd most want a second opinion on, because the docstring sells it as the thing only a real server can provide (`:24-26`, "against a real server they double as a leak check").

`_raise_classified_teradata_error` builds the message from an int code, a fixed map descriptor and the table name (`teradata.py:155-157`), and `UnstructuredIngestError` has no `__str__` override - `error_string` is only applied inside `wrap()` (`error.py:116-117`). Constructing it the way the connector does:

```
UserError str() -> "Teradata error 2801 (duplicate value for a unique primary index) for 'elements_strict_ab12cd34'."
status_code     -> 422
```

So the five assertions at `:176-182` (no host, no user, no password, no `CONNECTION=`, no "Failed to connect") cannot fail on the path that reaches them. That's a guarantee that reads stronger than it is.

Related: `CONNECTION=` appears exactly once in the whole repo, at `:182`. There's no real driver message anywhere to check it against - #812's unit fixtures are hand-written strings. If that isn't the token `teradatasql` actually emits, the assertion is asserting nothing. Do you have one real redacted connect failure and one real 2801 to pin it to?

### 5. The fixtures call the driver unguarded, which is what the docstring forbids

`:24-25` says never print the raw teradatasql message because it embeds host, user and password. Four sites do exactly that, with nothing between them and pytest:

| line | call |
|---|---|
| `:110` | `get_connection()` -> `teradatasql.connect(...)` |
| `:128` | `cursor.execute(read_schema(...))` CREATE |
| `:134` | `cursor.execute('DROP TABLE ...')` |
| `:166` | `cursor.execute('SELECT COUNT(*) ...')` |

A wrong `TERADATA_PASSWORD` raises inside `connect()` during fixture setup, which is the most likely first-run event, and pytest prints it verbatim. There's no root `conftest.py` and no `pytest_exception_interact` / `pytest_runtest_makereport` hook anywhere under `test/` (checked), so nothing redacts it. Wrapping the fixture's `get_cursor` body and re-raising through `safe_error_summary` would make the docstring true of the whole file.

---

### Smaller things

**CHANGELOG heading.** Filed under `### Enhancements` with a `feat(` prefix. Every prior test-only entry is under `### Fixes` or `### Chores`, including a previous `test(teradata)` entry. The PR title is already `test(teradata):`.

**The Databricks analogy is doing less work than it looks.** "Not wired into CI" is right, but the mechanisms differ. Databricks creds *are* plumbed into `sql_connectors_int_test` (`e2e.yml:167-170`); they're inert only because `test_databricks_delta_tables.py:103` carries an unconditional `@pytest.mark.skip` on top of its `@requires_env`. These tests have only `@requires_env`. Practical difference: they start running the first time anyone adds `TERADATA_*` to the job env, whereas the databricks one stays skipped forever.

**`schema.sql` is byte-identical to `assets/teradata_elements_schema.sql`** (verified with `diff`) with no drift guard, and no other connector test duplicates a product asset. Also, because the fixture pre-creates the table, `create_destination()` returns early at the `DBC.TablesV` probe (`teradata.py:511-512`), so the lines that read the asset and run the CREATE (`:515-523`) never execute. Letting `create_destination()` build the table would remove the copy and add coverage.

**`metadata_as_json=True` has no integration fixture.** It's the branch `create_destination()` requires. `test_astradb.py:337-352` is the precedent for a second config via `expected_folder=`.

**`read_schema` fragility.** `startswith("--")` is applied to the raw line and the join has no separator (`:103-104`), so an *indented* `--` comment would survive the filter and comment out the rest of the joined line. Correct for both fixtures today, but `schema_strict.sql` ships a comment block that invites editing. `line.strip().startswith("--")` and `" ".join(...)` would harden it. An `assert new_table_name in body` is also cheap insurance, since a missed substitution would issue `CREATE TABLE "elements"` against a shared database.

**Neither rejection test checks the table after the raise.** `count_rows` is called once, at `:206`. For the single-row 6706 case, `assert count_rows(strict_table) == 0` is a cheap post-condition. For the duplicate-key case I'd confirm against a real server first, since `teradata.py:637-642` documents that batches commit independently, so whether zero or one row survives a 2801 depends on teradatasql array-insert semantics.

**`delete_by_record_id`'s effect is untested.** To be fair, the override does execute on all three live paths, so its quoting, `_get_db_column_name` case resolution and error classification are exercised. What isn't covered is the row-removal semantics: the `rowcount > 0` branch and the replace-on-reupload guarantee. `test_postgres.py:167/181` gets that by running the upload twice and re-asserting the count.

**Source direction is out of scope here, worth stating as a boundary.** All four tests are `DESTINATION_TAG`, so `TeradataIndexer.precheck` gets no live exercise. It isn't untested - `test/unit/connectors/sql/test_teradata.py:1339` and `:1365` cover the fallback and the forwarded message - but a real 3807 on the source-table probe is unverified. Reasonable to defer.

**ndjson parametrization**, purely for symmetry with the five siblings that have it. Being honest about the value: `TeradataUploadStager` overrides only `conform_dict` and `conform_dataframe`, both of which take already-parsed input, so ndjson would feed the same objects through. It also needs a new fixture. Only worth it if you want the file to read like its neighbours.

**Teardown drop is unguarded.** `:134` runs a bare `DROP TABLE` inside the `finally`. If the table is gone or the connection is unhealthy, the teardown exception is reported alongside the real failure. Teradata has no `IF EXISTS`, so the equivalent is swallowing 3807, which the connector already recognises at `teradata.py:74`.

**Docstring run command** omits `--locked`, unlike every Makefile target. Minor, but without it a drifted `pyproject.toml` causes uv to re-resolve and rewrite `uv.lock` as a side effect of running a test.

---

### Two pre-existing defects I hit while reviewing (not this PR)

Flagging because they affect how much the green checks here mean. Both are on `main`; happy to file them separately.

**`check_untagged_tests.sh` is vacuous, in two independent ways.** You found the first in the PR body. The second: lines 9 and 14 use `grep "<Function|<Coroutine" -c` with no `-E`, so in BRE the `|` is a literal character and the pattern never matches. Measured on a file containing both tokens, as-written returns `0` and `grep -E` returns `2`. So even fixing the bare-`pytest` problem leaves the guard green.

**`--tags` swallows the path argument.** `pytest_tagging` declares `--tags` as `nargs="*"`, so `Makefile:95`'s `pytest --tags sql test/integration/connectors` consumes the path as a tag value and pytest falls back to collecting from rootdir. `--tags sql -- <path>` would fix it.

---

Two open questions where a single live run would settle more than any amount of reading:

1. Does `SET SESSION CHARACTER SET UNICODE PASS THROUGH ON` (`teradata.py:285`, run on every connection) change or suppress 6706 for `中文` into a `LATIN` column? If it does, the failure looks identical to the "#812 hasn't landed" failure.
2. Is `VECTOR32` available on the ClearScape trial the docstring points runners at, and will teradatasql bind the stager's comma-separated float string into it through a `?` parameter? If not, the first test dies on the first statement.

Also: `_extract_teradata_error_code` takes the *last* `[Error NNNN]` match. If teradatasql appends its own batch-level tag after the server's 2801, the classifier would read the wrapper code even after #812 lands. Worth checking against one real batch failure.

paulkarayan:
Thanks for running it locally, that is the part I could not do. Most of this was right and is now fixed in two commits on the branch: `e2dd53fd` (the test file, plus deleting the duplicated schema fixture) and `ebc313b0` (the CHANGELOG heading). Line numbers below are post-change.

Taking the file being unrun seriously means an assertion that cannot fail is the only kind of defect it can have, so items 1, 3 and 4 got real changes rather than comment fixes.

**cubic P3, `NON_LATIN_TEXT`.** Fixed. It is the `\u4e2d\u6587` escape now and the file is pure ASCII (`test_teradata.py:85`, checked byte by byte).

**1. Flat stager against the JSON-blob schema.** Fixed, and it took the schema fixture with it. The test now stages with `TeradataUploadStagerConfig(metadata_as_json=True)` and stops pre-creating the table, so `create_destination()` builds it from `assets/teradata_elements_schema.sql` and those lines actually run for the first time. That deletes `env_setup/sql/teradata/destination/schema.sql`, which was your byte-identical copy with no drift guard. Content is checked now, not just the count: the staged record's text and its metadata blob are read back (`test_teradata.py:348-353`), and the file uploads twice and re-asserts the count (`:355-356`), which is the `delete_by_record_id` row-removal coverage you asked for, same shape as `test_postgres.py`.

**2. The comment at `:64-65`.** Fixed. It says neither code is in the map yet and points at the skip that gates on it (`test_teradata.py:76-77`).

**3. Fall-through reads as a connect error.** Fixed with your suggestion nearly verbatim: `needs_widened_code_map` at `test_teradata.py:90-97` skips both rejection tests unless 2801 and 6706 are both in `_USER_FAULT_TERADATA_CODES`, and clears itself when https://github.com/Unstructured-IO/unstructured-ingest/pull/812 lands. With no credentials the run now reports `needs 2801/6706 in _USER_FAULT_TERADATA_CODES (unstructured-ingest#812)` rather than the env-var reason.

**4. The leak assertions can't fail.** You were right, and no, I did not have a real driver message to pin `CONNECTION=` to, so that assertion was asserting nothing. It is gone. Each rejection is now also provoked through a bare cursor (`provoke_raw_driver_error`, `test_teradata.py:247`), and the checks run against the exception the driver really produced: the secret has to BE in the driver's text and NOT in ours (`:304-310`), the driver's text must not appear in ours (`:311-312`), and `_extract_teradata_error_code` has to read the right code off it (`:295`). That last one is also the answer to your closing note about the last-match rule: if teradatasql appends a batch-level tag after the server's 2801, this fails on a real server and names the problem. Secrets are compared as booleans so a failure never prints them.

**5. Fixtures call the driver unguarded.** Fixed. Every driver call in the file goes through `get_cursor` (`test_teradata.py:154-163`), which re-raises a driver failure as a redacted `TeradataTestError` carrying the numeric code. A wrong `TERADATA_PASSWORD` during fixture setup now prints a type name and a code, not the connection text. Anything the test body raises passes through untouched, and the raw provocation catches its own exception inside the cursor block so the real message is still available where it is wanted.

Smaller things:

- **CHANGELOG heading.** Fixed, `ebc313b0`: `### Fixes`, prefix `test(teradata):`, and the text updated to describe what the tests now do.
- **The Databricks analogy.** Fixed in the docstring (`test_teradata.py:13-20`). It now says the mechanisms differ, that the Databricks credentials are in the job environment and that test is inert only because of the unconditional skip, and that these would start running the first time `TERADATA_*` appeared there.
- **`schema.sql` duplicating the product asset.** Fixed by deleting it, see item 1.
- **`metadata_as_json=True` has no integration fixture.** Fixed, it is the only config the upload test uses now. I did not add a second flat-schema fixture: the flat pairing has no connector-side DDL to create it, so it would need the copy back.
- **`read_schema` fragility.** Fixed. Strips before matching `--`, joins with a space, and asserts the substitution happened (`test_teradata.py:144-149`).
- **Post-raise row check.** Fixed for 6706: `assert count_rows(strict_table) == 0` at `test_teradata.py:425`. Not done for 2801, with your reasoning written into the code as the reason to revisit it.
- **Teardown drop.** Fixed. `drop_table` (`test_teradata.py:166-179`) swallows 3807 and re-raises anything else.
- **`--locked`.** Fixed, the docstring command matches the Makefile targets.
- **Source direction.** Not changed. Agreed as a boundary: all four tests are `DESTINATION_TAG`, and a live 3807 on the indexer probe is unverified.
- **ndjson parametrization.** Not changed, on your own argument: `TeradataUploadStager` overrides only `conform_dict` and `conform_dataframe`, both of which take parsed input, so the same objects go through either way and it costs a new fixture for no coverage.

Your two open questions:

1. **`UNICODE PASS THROUGH ON` and 6706.** Unverified, and it stays unverified until someone runs it. What changed is that the two failures are now distinguishable: if the server accepts the character, the raw provocation fails with "the provocation was ACCEPTED by <table>", not with a classification failure, and the 6706 test docstring says so (`test_teradata.py:401-419`).
2. **VECTOR32 on the ClearScape trial.** Unverified. Worth noting the test now hits it the way a user does, through `create_destination()` and the connector's own asset, so if the trial cannot take a VECTOR32 column or teradatasql will not bind the stager's comma-separated float string into one, that is a product finding and not a fixture artifact.

**Your two pre-existing defects.** Both confirmed here, neither touched: I ran the corrected guard (`grep -E`, project venv on PATH) and it reports 63 untagged functions repo-wide, zero of them teradata. `--tags sql -- <path>` collects all four of mine. Out of scope for this PR, so they are yours to file, and the PR body's claim that the guard reports zero should be read as "zero for this module" until the script is fixed.

**Still unrun.** No live Vantage listener is reachable from any of our environments, so nothing in this file has executed against a server. What ran: `ruff format --check` and `ruff check` clean on the file, `pytest --collect-only` collects 4 tests, a credential-free run is 1 passed and 3 skipped with the intended reasons, `--tags sql` picks all four up, and the 106 teradata unit tests still pass. Running it for real needs a Vantage instance reachable on 1025 with a database the user can CREATE TABLE in, which is a ClearScape trial or a customer environment, plus https://github.com/Unstructured-IO/unstructured-ingest/pull/812 merged.

paulkarayan:
Follow-up, because #812 merging changed what this file should assert.

#812 landed 6706 in `_USER_FAULT_TERADATA_CODES` and deliberately left 2801 out: `upload_dataframe`
commits each batch separately, so a retry after a partial upload can re-insert rows this connector
already wrote when `can_delete()` is False, which makes a duplicate as likely to be ours as the
customer's, and platform-plugins pins the same code as a provider-side failure.

That broke the gate you reviewed in two ways. `needs_widened_code_map` required BOTH codes, so it
would have skipped forever, and the duplicate-key test asserted `UserError` for a code the repo had
just decided is not the user's fault. Both are fixed in `c05884a5`:

- The untranslatable-character test gates on 6706 alone, which is now satisfied.
- The duplicate-key test is renamed and inverted. It pins the `DestinationConnectionError` path as
  the deliberate decision, asserts the error is NOT a `UserError`, and still proves through the raw
  cursor that the server really answered 2801, so it cannot pass on a genuinely unreachable host.
  Adding 2801 to the map later skips the test with a reason telling you to update it along with the
  decision, rather than letting it silently become vacuous.
- The module docstring says why the two rejections are classified differently instead of claiming
  both are waiting on a widened map.

Also in this push: a merge of `origin/main` carrying #812. One conflict, the CHANGELOG and version
pair, resolved to 1.11.15 with both entries kept.

Still unrun against a server, as before.

